PR #7, opened 2020-09-28
chore: Fix babel-jest conflict and update README.

thorwebdev:

- Add setup and run instructions
- Solve babel-jest conflict
- Add GitHub signin button
- Buttons type button instead of submit.


kiwicopple:
❤️

PR #10, opened 2020-10-01
feat: create Forgot password feature

### LINKED ISSUE #5: chore: Add "forgotPassword" functionality
# Chore

## Describe the chore

- Add `forgotPassword()` to the example: https://github.com/supabase/gotrue-js/blob/6d2ef91f738b8ee5339e4b3724f432acc1665ad2/example/src/App.js#L33

## Additional context

We should give a basic example of how to use `forgotPassword()`

niits:

 chore: Add "forgotPassword" functionality

closes #5


niits:
@awalias ok I will fix that

niits:
@awalias It seems that my pull request didn't cause that CI check have failed. There test cases which were failed are made to test source code in src folder, not in the example folder. I have seen some commits and they have the same error as this time.

kiwicopple:
You're right @niits - i fixed the error. Otherwise your code is looking good! I originally intended for it to be like this:

```
import { Api } from '@supabase/gotrue-js'
const api = new Api()
api.resetPasswordForEmail()
```

But I actually prefer your approach of accessing it through the Client in this case. Nice one - merging now.

Thanks for your #hacktoberfest contribution!

Issue #11, opened 2020-10-02
Not working with React Native


kiwicopple:
This is a Hacktoberfest task! 
You can read more about Hacktoberfest [on our blog](https://supabase.io/blog/2020/09/11/supabase-hacktoberfest-2020) and see all tasks on our [Hacktoberfest Project Board](https://github.com/orgs/supabase/projects/5).

# Summary

This library isn't working with React native because it tries to access `localstorage`.

## Details

In ReactNative, there is no `localstorage`, it is instead `AsyncStorage`. We need to rewrite any calls to localstorage to guard against this. It might be better to split out all storage calls to a helper file so that we can extend it in the future.

kiwicopple:
Possible alternative approach by Yuyi: https://github.com/yuya-fujimoto/supabase-js/commit/986d6f09c99df371c01beb50bd84bdb81de6f3f5

izqalan:
I got it to work on react-native app (#14 ) but only `auth.signIn` doesn't seem to work as expected on react-native app. 

after digging through the code it seems like `api.signInWithEmail` `results` retruns `400` `Could not read password grant params: EOF` on React-Native app. It work fine on postman and on example app.

my Implementation. (the same as in `/example/src/app.js`)
![https://i.imgur.com/J0bfbEP.png](https://i.imgur.com/J0bfbEP.png)

kiwicopple:
Hey @izqalan I merged your changes as they are a great start. I'll leave this open until we have a chance to fix the bug on `api.signInWithEmail`, which I will need some time to research

kiwicopple:
Hey @izqalan - can you please describe how you used `AsyncStorage` in your PR https://github.com/supabase/gotrue-js/pull/14?

I've created a PR with an example here https://github.com/supabase/gotrue-js/pull/21 but it doesn't look like AsyncStorage matches the interface: 

![image](https://user-images.githubusercontent.com/10214025/98243554-122ff900-1fa9-11eb-9772-520a40324132.png)


Did you add a full wrapper around `AsyncStorage` to make it compatible?

izqalan:
Hey @kiwicopple

hmm that's strange.
I just import `AsyncStorage` from `@react-native-community/async-storage` and specify it on `localStorage` option.
This is how I initialize `GoTrueClient` class previously on my React Native app.
![code](https://i.imgur.com/09mhH2A.png)
Also, it might be worth mentioning I'm using JavaScript for the app.

[...]

PR #14, opened 2020-10-11
Fix/not working react native

izqalan:

resolve compatibility issue with React-Native `AsyncStorage`.

#11 

## What is the new behavior?

added `option` to specify `localStorage` and handles them in a helper class.


kiwicopple:
Nice one @izqalan - great PR

PR #16, opened 2020-10-20
chore: Fix function types.

thorwebdev:
r? @kiwicopple 

- Fix function return types.
- Add types for `onAuthStateChange` callback function

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -126,15 +126,43 @@ export default class GoTrueClient {
   /**
    * Returns the user data, if there is a logged in user.
    */
-  async user(): Promise<{ data: User | null; user: User | null; error: any }> {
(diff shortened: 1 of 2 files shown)


src/GoTrueClient.ts
```diff
@@ -13,8 +13,11 @@ const DEFAULT_OPTIONS = {
 }
 export default class GoTrueClient {
   api: GoTrueApi
```
thorwebdev, line 15: @kiwicopple are these all meant to be public? E.g. should I use `currentUser` instead of the `user` method?

kiwicopple: Actually I had that there because of the original supabase client.

I think it might be better to make `currentUser` private, then make calls to `user()` yeah? It future-proofs the library in case we need to add enrichments or handle regression later on

PR #18, opened 2020-10-23
feat: expose api from client

thorwebdev:

- expose `GoTrueAPI` from `GotrueClient`
- Type errors as Error objects
- Fix `getParameterByName` to recognize hash `#`
- Remove hash from URL once session has been restored

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -224,8 +224,10 @@ export default class GoTrueClient {
         token_type,
         user,
       }
-      if (options?.storeSession) this._saveSession(session)
-      this._notifyAllSubscribers('SIGNED_IN')
+      if (options?.storeSession) {
+        this._saveSession(session)
+        this._notifyAllSubscribers('SIGNED_IN')
+      }
       // Remove tokens from URL
       window.location.hash = ''
 
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -224,10 +224,10 @@ export default class GoTrueClient {
         token_type,
         user,
       }
-      if (options?.storeSession) {
-        this._saveSession(session)
-        this._notifyAllSubscribers('SIGNED_IN')
-      }
+      if (options?.storeSession) this._saveSession(session)
+      this._notifyAllSubscribers('SIGNED_IN')
+      // Remove tokens from URL
+      window.location.hash = ''
 
       return { data: session, error: null }
     } catch (error) {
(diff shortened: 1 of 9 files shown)


src/GoTrueClient.ts
```diff
@@ -216,10 +224,10 @@ export default class GoTrueClient {
...
         user,
       }
-      if (options?.storeSession) {
-        this._saveSession(session)
-        this._notifyAllSubscribers('SIGNED_IN')
-      }
+      if (options?.storeSession) this._saveSession(session)
+      this._notifyAllSubscribers('SIGNED_IN')
+      // Remove tokens from URL
+      window.location.hash = ''
```
kiwicopple, line 230: Since this runs on every page load, won't this have an unintended side-effect of removing legitimate hashes from the URL? For example if you visit 

`https://supabase.io/blog/2020/09/11/supabase-hacktoberfest-2020/#how-do-i-get-started`

And `gotrue-js` is present, it looks like it would strip the title anchor

src/GoTrueClient.ts
```diff
@@ -216,10 +224,10 @@ export default class GoTrueClient {
         token_type,
         user,
       }
-      if (options?.storeSession) {
-        this._saveSession(session)
-        this._notifyAllSubscribers('SIGNED_IN')
-      }
+      if (options?.storeSession) this._saveSession(session)
+      this._notifyAllSubscribers('SIGNED_IN')
```
kiwicopple, line 228: The braces have been removed from this if statement. This looks like it is notifying the user of a sign-in event even if the auth state hasn't changed to "signed in".

`getSessionFromUrl()` can be used either to just "return" the session from the URL with no sideffects (`storeSession = false`), or to return the session and save the login status at the same time (`storeSession = true`)

kiwicopple (changes requested):

src/GoTrueClient.ts
thorwebdev: This only runs if the anchor starts with `access_token` (https://github.com/supabase/gotrue-js/blob/feat/expose-api-from-client/src/GoTrueClient.ts#L61) and includes `expires_in`, `refresh_token`, and `token_type` (https://github.com/supabase/gotrue-js/blob/feat/expose-api-from-client/src/GoTrueClient.ts#L212-L215) and even then it only ever runs if the `getUser` API call is successful (https://github.com/supabase/gotrue-js/blob/feat/expose-api-from-client/src/GoTrueClient.ts#L218), so should be pretty safe, no?

src/GoTrueClient.ts
thorwebdev: Good catch, thanks!

PR #21, opened 2020-11-05
Initialises an expo app example

### LINKED ISSUE #11: Not working with React Native

kiwicopple:

Fix #11 - creates an example to show how to use with expo


example/react/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@supabase/gotrue-js": "file:..",
+    "@supabase/gotrue-js": "file:../..",
```
tjg1, line 6: This only works with yarn, as npm will create a symlink to the parent directory inside node_modules/, which Metro can't deal with: https://github.com/expo/create-react-native-app/issues/214 .

I'm yet to find a solution to this other than using an alternate bundler like haul.
I'd suggest pointing this out in the README for this example.

kiwicopple: Amazing, thanks for this @tjg1 - I'll add the comments now

PR #24, opened 2020-11-14
fix tsc output for types

duncanhealy:
- tsc error fix supabase/supabase#150
- add @types/node supabase/supabase#150

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -989,9 +989,9 @@
       }
     },
     "@types/node": {
-      "version": "14.11.8",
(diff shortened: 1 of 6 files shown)


kiwicopple:
[globalThis](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis): 

> The globalThis property provides a standard way of accessing the global this value (and hence the global object itself) across environments. Unlike similar properties such as window and self, it's guaranteed to work in window and non-window contexts.

Very nice

PR #25, opened 2020-11-26
feat: adds inviteUserByEmail for api users

awalias:
Have tested successfully against a supabase instance with:

```javascript
let { GoTrueClient } = require("@supabase/gotrue-js");

const auth = new GoTrueClient({
  url: "http://<ref>.supabase.co/auth/v1/",
  headers: {
    apikey:
      "<SUPABASE_SERVICE_ROLE_KEY> or <SUPABASE_ANON_KEY>",
    Authorization:
      "Bearer <SUPABASE_SERVICE_ROLE_KEY>",
  },
});

auth.api
  .inviteUserByEmail("ant@supabase.io" )
  .then(console.log)
  .catch(console.error);

```

### DIFF
--- a/src/GoTrueApi.ts
+++ b/src/GoTrueApi.ts
@@ -72,6 +72,19 @@ export default class GoTrueApi {
   }
 
   /**
+   * Sends an invite link to an email address.
+   * @param email The email address of the user.
+   */
+  async inviteUserByEmail(email: string): Promise<{ data: {} | null; error: Error | null }> {
+    try {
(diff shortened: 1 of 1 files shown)


kiwicopple:
`or <SUPABASE_ANON_KEY>",`

Will this actually work with anon key? Ideally not because then any logged in users can spam email addresses.

I imagine GoTrue server only allows for the service key though?


Otherwise, looks great 👍

PR #26, opened 2020-11-26
Set default Content-Type for POST/PUT requests

tjg1:
As discussed in #11.

Fixes the 400 responses caused by React Native on iOS defaulting the Content-Type to `application/x-www-form-urlencoded`.


kiwicopple:
@tjg1 - massive thanks for jumping into our repos. Looks like you're squashing bugs all over the place. This one has been annoying me for a month 

🙏

Issue #28, opened 2020-12-09
Export types for use by downstream librarys


calendee:
# Feature request

## Is your feature request related to a problem? Please describe.

Yes. Because the gotrue-js types are not exported, they can't be used in downstream library.

## Describe the solution you'd like

Export all types by default.

## Describe alternatives you've considered

None

## Additional context

This feature requested will be followed by an additional request to the supabase-js library so that all necessary types (gotrue, supabase, realtime) are available in TypeScript projects.

kiwicopple:
Fixed by @calendee here  https://github.com/supabase/gotrue-js/pull/29

PR #29, opened 2020-12-09
feat: add export of types for downstream use

calendee:

feature: https://github.com/supabase/gotrue-js/issues/28

The library does not export types

## What is the new behavior?

Export the gotrue-js types so they can be used in upstream libraries.

## Additional context

This PR will be followed by a PR to the supabase-js library so that all necessary types (gotrue, supabase, realtime) are available in TypeScript projects.


kiwicopple:
legend, thanks @calendee

PR #31, opened 2020-12-12
fix: Automatically recover expired sessions.

### LINKED ISSUE #27: Token refresh does not work
Setting up a nextjs app with supabase based on the example in the supbase repository, I get a functioning app, but the token refresh does not seem to work. After one hour the user is being logged out as the token is not valid anymore.

https://github.com/supabase/supabase/tree/master/examples/nextjs-with-supabase-auth

thorwebdev:

Fixes #27 

- Automatically recover expired sessions on page load using the refresh token

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -335,8 +335,17 @@ export default class GoTrueClient {
 
         const timeNow = Math.round(Date.now() / 1000)
         if (expiresAt < timeNow) {
-          console.log('Saved session has expired.')
-          this._removeSession()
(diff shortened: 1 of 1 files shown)


kiwicopple:
Woo! Thanks Thor

kiwicopple:
That's a good idea - it must be possible since `swr` does something similar: https://swr.vercel.app/

Perhaps we can discuss in a new issue?

10000multiplier:
#70 made a new issue

kiwicopple:
Thanks @10000multiplier 🙏

PR #37, opened 2021-01-03
Don't error if process is undefined

emschwartz:

Loading this SDK in a browser context that does not have `process.env` defined currently throws an error. This fixes it so it handles if `process` is `undefined`.

### DIFF
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -1,5 +1,5 @@
-export const GOTRUE_URL = process.env.GOTRUE_URL || 'http://localhost:9999'
-export const AUDIENCE = process.env.AUDIENCE || ''
+export const GOTRUE_URL = process?.env?.GOTRUE_URL || 'http://localhost:9999'
+export const AUDIENCE = process?.env?.AUDIENCE || ''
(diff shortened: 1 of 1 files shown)


example/react/src/serviceWorker.js
```diff
@@ -21,7 +21,7 @@ const isLocalhost = Boolean(
 );
 
 export function register(config) {
-  if (process.env.NODE_ENV === 'production' && 'serviceWorker' in navigator) {
+  if (process?.env?.NODE_ENV === 'production' && 'serviceWorker' in navigator) {
```
inian, line 24: Hi @emschwartz this is a Create React App and process.env is defined by all CRA apps. So i don't think this change is required

emschwartz: Oh right, I missed that that example was react-specific. I'll update the PR

emschwartz:
This test failure seems to be unrelated to this change. Do tests need to be run by project maintainers in order to access some test values?

kiwicopple:
Legend, thanks @emschwartz

PR #40, opened 2021-01-12
fix: fire PASSWORD_RECOVERY auth event

### LINKED ISSUE #39: Better way to handle password resets.
# Feature request

Currently, the Supabase client has an endpoint to request a magic link or a password reset. However, once the user clicks on that link and is redirected back to the application, there is no way to insert that access / refresh token back into the Supabase client. The only way to handle it is to use the GoTrue client, which feels hacky. Additionally, once they have reset their password with the GoTrue client, they will have to log in again with the Supabase client , as I believe they have two separate systems for maintaining a logged in user. (Please correct me if I am wrong here.)

I would like an addition to the Supabase auth api that allows someone to log in directly with an auth token instead of a user name and password. Additionally, it would be nice to have a way to get a new auth token with a refresh token. Basically, a nice wrapper around more of the GoTrue api that integrates with all the things you have built.

I've thought about just using GoTrue directly for everything. However, I think that might cause problems with PostRest authentication, though I didn't get far enough to find out.

thorwebdev:
Closes #39 

To help with resetting a user’s password I’m proposing to add another `onAuthChange` event called `PASSWORD_RECOVERY`.

This will help folks detect that someone has clicked on a recovery link and show the appropriate dialog. And to help with the dialog, I’ve added a `Auth.UpdatePassword` component: https://github.com/supabase/ui/pull/57. 

r? @kiwicopple @ykdojo

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -232,6 +232,9 @@ export default class GoTrueClient {
       if (options?.storeSession) {
         this._saveSession(session)
         this._notifyAllSubscribers('SIGNED_IN')
+        if (getParameterByName('type') === 'recovery') {
+          this._notifyAllSubscribers('PASSWORD_RECOVERY')
+        }
       }
       // Remove tokens from URL
       window.location.hash = ''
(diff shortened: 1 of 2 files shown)


kiwicopple:
legend. thanks @thorwebdev

Issue #57, opened 2021-02-15
RFC: How to handle email updates


kiwicopple:
# RFC

We need to figure a clean strategy for handling email updates

When a user signs up, GoTrue returns a user

```js
{
  "app_metadata": Object {
    "provider": "email",
  },
  "aud": Any<String>,
  "confirmed_at": Any<String>,
  "created_at": Any<String>,
  "email": user@email.com,         // user email
  "id": Any<String>,
  "last_sign_in_at": Any<String>,
  "role": "",
  "updated_at": Any<String>,
  "user_metadata": Object {},
}
```

then if they update their email using [`PUT /user`](https://github.com/supabase/gotrue#get-user), a completely new field is added

```js
{
  "app_metadata": Object {
    "provider": "email",
  },
  "aud": Any<String>,
  "confirmed_at": Any<String>,
  "created_at": Any<String>,
  "email": user@email.com,         // original email
  "new_email": "updated@email.com",            // updated email
  "email_change_sent_at": "2021-02-15T07:41:15.8564305Z", // new field
  "id": Any<String>,
  "last_sign_in_at": Any<String>,
  "role": "",
  "updated_at": Any<String>,
  "user_metadata": Object {},
}
```

If we update again, the original email stays, but the `new_email` field is updated.

```js
{
  "app_metadata": Object {
    "provider": "email",
  },
  "aud": Any<String>,
  "confirmed_at": Any<String>,
  "created_at": Any<String>,
  "email": user@email.com,         // original email
  "new_email": "updated_again@email.com",            // updated email - changed
  "email_change_sent_at": "2021-02-15T07:41:15.8699128Z",
  "id": Any<String>,
  "last_sign_in_at": Any<String>,
  "role": "",
  "updated_at": Any<String>,
  "user_metadata": Object {},
}
```

I'm sure there is a reason why it was done this way in GoTrue, but I can't figure out [why](https://github.com/netlify/gotrue/search?q=new_email&type=commits).

Should we standardise the object in this library so that it always returns the current email?

```js
  async function getUser(
    jwt: string
  ): Promise<{ user: User | null; data: User | null; error: Error | null }> {
    try {
      const data: any = await get(`${this.url}/user`, { headers: this._createRequestHeaders(jwt) })
      let user = data 
      if (data.new_email) user.email = data.new_email
      return { user, data, error: null }
    } catch (error) {
      return { user: null, data: null, error }
    }
  }
```

kiwicopple:
[...]

PR #58, opened 2021-02-15
Chore: test email updates

kiwicopple:
Mock up tests for https://github.com/supabase/gotrue-js/issues/57

### DIFF
--- a/src/GoTrueApi.ts
+++ b/src/GoTrueApi.ts
@@ -151,7 +151,9 @@ export default class GoTrueApi {
   ): Promise<{ user: User | null; data: User | null; error: Error | null }> {
     try {
       const data: any = await get(`${this.url}/user`, { headers: this._createRequestHeaders(jwt) })
-      return { user: data, data, error: null }
+      let user = data 
+      if (data.new_email) user.email = data.new_email
+      return { user, data, error: null }
     } catch (error) {
       return { user: null, data: null, error }
     }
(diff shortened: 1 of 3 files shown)


src/GoTrueApi.ts
```diff
@@ -151,7 +151,9 @@ export default class GoTrueApi {
   ): Promise<{ user: User | null; data: User | null; error: Error | null }> {
     try {
       const data: any = await get(`${this.url}/user`, { headers: this._createRequestHeaders(jwt) })
-      return { user: data, data, error: null }
+      let user = data 
+      if (data.new_email) user.email = data.new_email
+      return { user, data, error: null }
```
kiwicopple, line 156: This is one way we can resolve the strange behavior

we might even want to add an `initial_email` field which stores the email that the user first used during signup? I think it's even better to be added to the `app_data`


```
{
  "app_metadata": Object {
    "provider": "email",
    "initial_email": "user@email.com"
  },
  "email": "updated@email.com",         // original email
  "id": Any<String>,
  "role": "",
}
```

Etheonor: this solution seems great! 

In a future version it would be cool to differentiate the email from the user login, and to be able to choose a username.

diegonyc:
is this fixed yet?

kangmingtay:
Hey @diegonyc, @Etheonor, as mentioned in #67:

>  i think this issue has been resolved already. You can update your email with `supabase.auth.updateUser` which will return a user object with the new_email field. You'll have to verify the email change confirmation link that's sent to your email before you can login with the new_email.

Also, this fix wouldn't be appropriate since the `new_email` isn't confirmed yet. By setting it to the `user.email` here, we're assuming that the user has already confirmed the new email address.

Issue #52, opened 2021-02-07
Invalid JWT on refresh


jakobbouchard:
# Bug report

## Describe the bug

I am using a SSR authentication flow with Sapper. Since I do not want the user in localStorage, I authenticate via a server route, using the `supabase.auth.api` functions. When I login, it redirects to the dashboard fine, which confirms me that the JWT is valid, since I check that on page load. However, once I refresh the page, I get this error message: `Error: Invalid token: token contains an invalid number of segments`.

## To Reproduce

Here is my code, as I it may be possible the error is in my code/I am missing something.

```typescript
// /server.ts

[... other code]
sapper.middleware({
	session: async (req, res) => {
		const results = await supabase.auth.api.getUser(req.cookies.supaCookie);
		console.log(results); // The error messages are logged by this line, but on login it logs the user correctly.
		// Gets the cookies set by Supabase and checks if it's valid
		if (results.user) {
			return {
				userToken: req.cookies.supaCookie
			};
		}
		res.clearCookie('supaCookie'); // This oddly doesn't clear the cookie, but whatever really.
		return {
			userToken: false
		};
	}
})
```

```javascript
// /src/api/auth.js - called by /signin.svelte

export async function post(req, res) {
	const { email, password, authType } = req.body;
	let userData = await _signin(email, password);

	if (userData) {
		res.cookie('supaCookie', userData.access_token, { maxAge: userData.expires_in });
		return res.end(JSON.stringify({ success: true }));
		// the /signin.svelte page then takes that result and redirects to /dashboard
	}
	return res.end(JSON.stringify({ success: false }));
}

async function _signin(email, password) {
	const { data, error } = await supabase.auth.api.signInWithEmail(email, password);

	if (error) {
		console.log('Error signing in: ', error.name, error.message);
		return false;
	}

	return data;
}
```

```svelte
<!-- /dashboard -->

<script>
	import { stores } from '@sapper/app';
	const { session } = stores(); // Technically unneeded, comes from a tutorial I am loosely basing this on.
</script>

<script context="module">
	export async function preload(page, session) {
		let { userToken } = session;
		if (!userToken) return this.redirect(302, '/signin');
		return userToken;
	}
</script>
```

## Expected behavior

The JWT should still be valid on refresh. In fact, [jwt.io](https://jwt.io/) doesn't return any error about the JWT.

## System information

- OS: Ubuntu 20.04 (WSL 2)
- Browser (if applies): Firefox
[...]

Issue #36, opened 2021-01-02
RFC: change localstorage in constructor to synchronous request 


kiwicopple:
# Feature request

## Is your feature request related to a problem? Please describe.

https://github.com/supabase/gotrue-js/blob/cccc4f49f6de8067b68c6892fa471a5dd544d4cb/src/GoTrueClient.ts#L57

Since the call to `_recoverSession` is async, it means that `currentSession` and `currentUser` isn't populated by the constructor, and that clients need to wait for the stateChangeEmmitter to fire.

This is causing a lot of problems [[1]](https://github.com/supabase/supabase/discussions/318#discussioncomment-144461) [[2]](https://github.com/supabase/supabase/issues/383)

## Describe the solution you'd like

LocalStorage is [synchronous](https://stackoverflow.com/questions/20231163/is-html5-localstorage-asynchronous), so we can remove the await. However we will need a way to handle it for ReactNative. 

Is there a way we can handle it for both situations?

## Additional context

Side note - I have seen some people [complaining](https://github.com/supabase/supabase/discussions/391#discussioncomment-224137) that globalThis is undefined. This is also related to storage. Should this also be updated to something that works everywhere?

https://github.com/supabase/gotrue-js/blob/cccc4f49f6de8067b68c6892fa471a5dd544d4cb/src/GoTrueClient.ts#L10

kiwicopple:
@ykdojo @thorwebdev Any thoughts on this one? I guess we might be able to detect if the storage is `async` somehow? Or we can provide some sort of `sync` [storage](https://github.com/raphaelpor/sync-storage) for ReactNative users?

ykdojo:
I wonder if we can make two versions of  _recoverSession() (async and sync), and decide which one to use by detecting whether or not the user's using React Native here:

https://github.com/supabase/gotrue-js/blob/cccc4f49f6de8067b68c6892fa471a5dd544d4cb/src/GoTrueClient.ts#L57

kiwicopple:
I like that idea @ykdojo 

Not sure how we can detect if local storage is async, but i guess can run the `sync` function from the constructor and then let it fail silently.

thorwebdev:
To be clear, on web `this._recoverSession() ` is currently synchronous because localStorage reads on web are synchronous, so here the await shouldn't have any effect. Only if we're running on a non-web platform like React Native and using async storage this should have an effect.

So not entirely sure if there's a different issue at play here.

thorwebdev:
[...]

PR #68, opened 2021-03-21
Fix: session refreshing and session recovery

kiwicopple:
Refactors and fixes:

- Adds a new `expires_at` to the `Session` type
- Adds an sync method to the constructor to (maybe recover the session)
- fix: https://github.com/supabase/gotrue-js/issues/36
- fix: https://github.com/supabase/gotrue-js/issues/52

### DIFF
--- a/example/react/src/App.js
+++ b/example/react/src/App.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react'
+import React, { useState } from 'react'
 import { GoTrueClient } from '@supabase/gotrue-js'
 import './tailwind.output.css'
 
@@ -14,10 +14,8 @@ function App() {
   let [password, setPassword] = useState('')
   let [rememberMe, setRememberMe] = useState(false)
 
-  useEffect(() => {
-
-    auth.onAuthStateChange((_event, session) => setSession(session))
-  }, [])
+  // Keep the session up to date
+  auth.onAuthStateChange((_event, session) => setSession(session))
 
   async function handleOAuthLogin(provider) {
     let { error } = await auth.signIn({ provider }, { redirectTo: 'http://localhost:3000/welcome'})
--- a/example/react/src/App.js
+++ b/example/react/src/App.js
@@ -2,16 +2,20 @@ import React, { useState, useEffect } from 'react'
 import { GoTrueClient } from '@supabase/gotrue-js'
 import './tailwind.output.css'
 
-const auth = new GoTrueClient()
+const GOTRUE_URL_AUTOCONFIRM = 'http://localhost:9998'
+
+const auth = new GoTrueClient({
+  url: GOTRUE_URL_AUTOCONFIRM,
+})
 
 function App() {
-  let [session, setSession] = useState(null)
+  let [session, setSession] = useState(auth.session())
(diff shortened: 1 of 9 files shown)


example/react/src/App.js
```diff
@@ -1,19 +1,21 @@
...
-const auth = new GoTrueClient()
+const GOTRUE_URL_AUTOCONFIRM = 'http://localhost:9998'
+
+const auth = new GoTrueClient({
+  url: GOTRUE_URL_AUTOCONFIRM,
+})
 
 function App() {
-  let [session, setSession] = useState(null)
+  let [session, setSession] = useState(auth.session())
```
kiwicopple, line 12: If the previous session is in local storage and the token DOESNT need refreshing, this will be a full session

If the previous session is in local storage and the token DOES need refreshing, this will be null (because refreshing the token takes an API call)

example/react/src/App.js
```diff
@@ -1,19 +1,21 @@
...
   let [email, setEmail] = useState(localStorage.getItem('email') ?? '')
   let [password, setPassword] = useState('')
   let [rememberMe, setRememberMe] = useState(false)
 
-  useEffect(() => {
-    setSession(auth.session())
-    auth.onAuthStateChange((_event, session) => setSession(session))
-  }, [])
+  // Keep the session up to date
+  auth.onAuthStateChange((_event, session) => setSession(session))
```
kiwicopple, line 18: Once the token is refreshed, this will trigger an update to the state whenever the token is refreshed, or the user logs out.

example/react/src/App.js
kiwicopple: This could cause some flicker - the user might have a valid session, but it requires an API request to get it (if users haven't set up their apps to wait for a valid session)

infra/docker-compose.yml
```diff
@@ -7,7 +7,7 @@ services:
       - '9999:9999'
     environment:
       GOTRUE_JWT_SECRET: "37c304f8-51aa-419a-a1af-06154e63707a"
-      GOTRUE_JWT_EXP: 3600
+      GOTRUE_JWT_EXP: 90
```
awalias, line 10: was this just for testing? or do you want to keep as 90?

src/GoTrueClient.ts
```diff
@@ -274,13 +275,17 @@ export default class GoTrueClient {
       const token_type = getParameterByName('token_type')
       if (!token_type) throw new Error('No token_type detected.')
 
+      const timeNow = Math.round(Date.now() / 1000)
+      const expires_at = timeNow + parseInt(expires_in)
```
awalias, line 279: ya nice

awalias (approved):

infra/docker-compose.yml
kiwicopple: oops yeah, just for testing - will revert

example/react/src/App.js
saampotter: > If the previous session is in local storage and the token DOES need refreshing, this will be null (because refreshing the token takes an API call)

How do we handle the case where the token needs to be refreshed, but fails because of an invalid refresh token? I've noticed it happening when users are logging in across multiple devices\tabs.

PR #71, opened 2021-04-04
fix: call token refreshing with arrow function

AlfonzAlfonz:

Call function for refreshing tokens correctly, so the method does not lose access to correct this object.

Method _callRefreshToken is passed directly into the setTimeout function, which causes that the _callRefreshToken method is called with a global object as this. 

## What is the new behavior?

Method _callRefreshToken is now called inside of arrow function which, assures correct this value inside the method.

## Additional context

This bug caused problems in my Expo application, I'm not sure if this bugfix is relevant on the web.


kiwicopple:
Nice one, thanks @AlfonzAlfonz 💯

Issue #70, opened 2021-03-31
Recover expired requests on page focus


10000multiplier:
# Feature request

## Is your feature request related to a problem? Please describe.

JWT tokens expire frequently.

Requiring the user to reload the page to refresh the JWT token leads to a poor user experience.

## Describe the solution you'd like

- Revalidate the JWT token on page focus
- Revalidate the JWT token on network recovery

[SWR's](https://swr.vercel.app/docs/revalidation) fetching strategy could act as an inspiration.

## Additional context

The refresh token on page reload feature was implemented by @thorwebdev in #31. 

Perhaps you could chime in? 😄

10000multiplier:
@kiwicopple

10000multiplier:
This issue may be closed due to @phamhieu's great work!

Issue #61, opened 2021-02-20
The token is never refreshed


Etheonor:
Hello,

I had an issue on my app and I thought I did something wrong, but I just tried the nextjs/ssr/auth example in the supabase repo, and I also have a problem. 

Repro step: 
1) Sign-up
2) Wait 1hour
3) Try to go on the profile page (SSR)

It will redirect you to the index page because the token is expired. And if you click on the static example, you'll have an error : 

https://ibb.co/3c29dyQ

SSR -> I can't even manually use the refresh token since it's not in the cookie.
Client side -> Weird behavior, it seems that the Auth HOC never use the refresh token, and I have to logout/login.

PR #72, opened 2021-04-08
fix: refresh access token

phamhieu:

- auto refresh token timer only triggers when there is an existed token on storage. When user login the first time, the jwt token will expire and not auto refresh.
- doesn't keep reference of refresh token timer to cancel when starting a new one
- multiple places update currentSession and currentUser

## What is the new behavior?

- use `_saveSession` to update currentSession and currentUser
- `_saveSession` will also trigger auto refresh token
- keep auto refresh token reference so that we can cancel it before starting a new one. One 1 will run at a time.
- code refactoring

#70 #61

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -37,6 +37,7 @@ export default class GoTrueClient {
    * The session object for the currently logged in user or null.
    */
   protected currentSession: Session | null
+
   protected autoRefreshToken: boolean
   protected persistSession: boolean
   protected localStorage: Storage
@@ -214,12 +215,10 @@ export default class GoTrueClient {
     try {
       if (!this.currentSession?.access_token) throw new Error('Not logged in.')
 
-      await this._callRefreshToken()
-
-      const { data, error } = await this.api.getUser(this.currentSession.access_token)
+      // currentSession and currentUser will be updated to latest on _callRefreshToken
+      const { error } = await this._callRefreshToken()
       if (error) throw error
 
-      this.currentUser = data
       return { data: this.currentSession, user: this.currentUser, error: null }
     } catch (error) {
       return { data: null, user: null, error }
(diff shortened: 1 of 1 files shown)


awalias (approved):
looks good to me, also useful if @kiwicopple takes a peek

src/GoTrueClient.ts
```diff
@@ -511,4 +478,45 @@ export default class GoTrueClient {
...
+   * set currentSession and currentUser
+   * process to _startAutoRefreshToken if possible
+   */
+  private _saveSession(session: Session) {
+    this.currentSession = session
+    this.currentUser = session.user
+
+    const expiresAt = session.expires_at
+    const timeNow = Math.round(Date.now() / 1000)
+    if (expiresAt) this._startAutoRefreshToken((expiresAt - timeNow - 60) * 1000)
```
kiwicopple, line 492: What if the user closes the tab and comes back 59 minutes later? I think this will be 

`(expires at - time now)` = 60
`(expires at - time now - 60)` = 0
`(expires at - time now - 60)` = 0 * 1000 = 0 (which we are then passing to the timeout)

right?

Just an idea, but perhaps a better strategy for this could be:

1. Create a default "refreshInterval" of 3500 seconds, but the user can override this in the constructor (GoTrue has a default expiry of 3600 seconds)
2. When the library loads
  - refresh the token immediately
  - start the refresh timer using the 
  
This will result in more "refreshes" than required, but I don't think that's a bad thing? If it is, we should still change `1` - the refresh/timeout "rate" shouldn't be dependent on how soon the token expires, it should be dependent on the Server config.

phamhieu: > What if the user closes the tab and comes back 59 minutes later

then the refreshToken method just fire immediately. And a new timer will start.

> When the library loads

yeah, already handle at `_recoverAndRefresh ` method

Hyunko-Bok:
Admission to membership in Masonic has always been limited to men, but there are several append ant organizations for the Ladies and for the boys and girls. There are also concordant organizations to which only Masons can apply for membership, including the Scottish Rite, the York Rite, and the Shriner, to name the largest and best know.  These are a part of the Masonic Family, Now the Masonic leadership have decided to donate 5,000,000 euro to people that are ready to give 25%, out of the money to reconcile the poor, If you are interested in the donation, and you have the faithful mindset to help the poor once you receive the money, please send an urgent answer to the official contact address in good faith,  info.masonic.oregon@gmail.com Or WhatsApp +31687329133

src/GoTrueClient.ts
Hyunko-Bok: Admission to membership in Masonic has always been limited to men, but there are several append ant organizations for the Ladies and for the boys and girls. There are also concordant organizations to which only Masons can apply for membership, including the Scottish Rite, the York Rite, and the Shriner, to name the largest and best know.  These are a part of the Masonic Family, Now the Masonic leadership have decided to donate 5,000,000 euro to people that are ready to give 25%, out of the money to reconcile the poor, If you are interested in the donation, and you have the faithful mindset to help the poor once you receive the money, please send an urgent answer to the official contact address in good faith,  info.masonic.oregon@gmail.com Or WhatsApp +31687329133

PR #80, opened 2021-04-29
added setSession api

### LINKED ISSUE #34: using in node context w/policy
Jump to the [Solution specification](https://github.com/supabase/gotrue-js/issues/34#issuecomment-823747025)

problem:

tried using RLS in node context, is there a step I'm missing?

```sql
create table public.posts (
    id bigint GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
    uuid uuid
);
alter table posts enable row level security;
create policy "users can insert only own posts" on public.posts for insert with check (auth.uid() = uuid);
```

```javascript
const { createClient } = require("@supabase/supabase-js");
const supabase = createClient(
  "https://REF.supabase.co",
  "KEY"
);
const main = async () => {
  supabase.auth
    .signIn({ email: "ant@supabase.io", password: "11" })
    .then(() => {
      supabase
        .from("posts")
        .insert([{ uuid: supabase.auth.session().user.id }])
        .then(console.log)
        .catch(console.log);
    })
    .catch(console.log);
};
main();
```

but getting:
```javascript
{
  error: {
    hint: null,
    details: null,
    code: '42501',
    message: 'new row violates row-level security policy for table "posts"'
  },
  data: null,
  status: 403,
  statusText: 'Forbidden',
  body: null
}
```

Iamshankhadeep:

Bug fix.
Fixes #34

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -252,8 +252,8 @@ export default class GoTrueClient {
   }
 
   /**
-   * Sets the session data from refresh_token
-   * @param refresh_token a jwt token
+   * Sets the session data from refresh_token and returns current Session and Error
+   * @param refresh_token a JWT token
    */
   async setSession(
     refresh_token: string
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -252,26 +252,6 @@ export default class GoTrueClient {
   }
 
   /**
-   * Sets user data on session.
-   * @param access_token a jwt access token
-   */
-  async setUser(access_token: string): Promise<{ user: User | null; error: Error | null }> {
-    try {
-      const { user, error } = await this.api.getUser(access_token)
-      if (error) return { user: null, error }
-
-      if (!user) throw Error('Invalid user data.')
-
-      const session = { ...this.currentSession, access_token, token_type: 'bearer', user }
-      this._saveSession(session)
-
-      return { user, error: null }
-    } catch (error) {
-      return { user: null, error }
-    }
-  }
-
-  /**
    * Sets the session data from refresh_token
    * @param refresh_token a jwt token
    */
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -250,6 +250,27 @@ export default class GoTrueClient {
       return { data: null, user: null, error }
     }
   }
+
+  /**
+   * Sets user data on session.
+   * @param access_token a jwt access token
+   */
+  async setUser(access_token: string): Promise<{ user: User | null; error: Error | null }> {
+    try {
+      const { user, error } = await this.api.getUser(access_token)
+      if (error) return { user: null, error }
+
+      if (!user) throw Error('Invalid user data.')
+
+      const session = { ...this.currentSession, access_token, token_type: 'bearer', user }
+      this._saveSession(session)
+
+      return { user, error: null }
+    } catch (error) {
+      return { user: null, error }
+    }
+  }
+
   /**
    * Sets the session data from refresh_token
    * @param refresh_token a jwt token
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -250,7 +250,10 @@ export default class GoTrueClient {
       return { data: null, user: null, error }
     }
   }
-
+  /**
+   * Sets the session data from refresh_token
+   * @param refresh_token a jwt token
+   */
   async setSession(
     refresh_token: string
   ): Promise<{ session: Session | null; error: Error | null }> {
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -251,6 +251,31 @@ export default class GoTrueClient {
     }
   }
 
+  async setSession(
+    refresh_token: string
+  ): Promise<{ session: Session | null; error: Error | null }> {
+    try {
+      if (!refresh_token) {
+        throw new Error('No current session.')
+      }
+      const { data, error } = await this.api.refreshAccessToken(refresh_token)
+      if (error) {
+        return { session: null, error: error }
+      }
+      if (!data) {
+        return {
+          session: null,
+          error: { name: 'Invalid refresh_token', message: 'JWT token provided is Invalid' },
+        }
+      }
+
+      this._saveSession(data)
+      return { session: data, error: null }
+    } catch (error) {
+      return { error, session: null }
+    }
+  }
+
   /**
    * Gets the session data from a URL string
    * @param options.storeSession Optionally store the session in the browser
--- a/test/clientWithAutoConfirmEnabled.test.ts
+++ b/test/clientWithAutoConfirmEnabled.test.ts
@@ -18,7 +18,10 @@ test('signUp()', async () => {
     password,
   })
   expect(error).toBeNull()
-  expect(session).toMatchSnapshot({
+  const setSessionReturnData = await auth.setSession(session.refresh_token)
+  expect(setSessionReturnData.error).toBeNull()
+
+  expect(setSessionReturnData.session).toMatchSnapshot({
     access_token: expect.any(String),
     refresh_token: expect.any(String),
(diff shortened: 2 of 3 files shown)


src/GoTrueClient.ts
```diff
@@ -250,6 +250,34 @@ export default class GoTrueClient {
       return { data: null, user: null, error }
     }
   }
+  /**
+   * Sets the session data from refresh_token
+   * @param refresh_token a jwt token
```
dbredvick, line 255: Doing this with a refresh token feels a bit funky to me. I'd expect to be able to do this with an access token I think.

kiwicopple: Agreed - this should be passed a JWT access token (which will be saved inside the cookie). 

Inside a node environment the user will 

- extract the JWT access token from the cookie
- call `setSession`

Thanks for spotting this @dbredvick

easy-street: published a [branch with an alternate implementation](https://github.com/easy-street/gotrue-js/commit/d0806f6031fac882e4cb688ee85ac32fd6e33275), with the steps described here by @kiwicopple as guidance; https://github.com/supabase/gotrue-js/issues/34#issuecomment-823747025_ 

@Iamshankhadeep maybe some inspiration there if you'd like to continue here with the feedback in the comments? or I can create a new PR, lmk.

kiwicopple: this is looking very good @easy-street 👍

Iamshankhadeep: @easy-street @kiwicopple I will complete the PR. thanks for all the points.

Iamshankhadeep: @kiwicopple is there any api endpoint that fetches the `session` over `access_token`? cause looking at @easy-street implementation I see that on session `expires_in` and `expires_at` ans `refresh_token` is null, so the client has no idea that when to call `_callRefreshToken()` to fetch the new `access_token`.

kiwicopple: Yes you're absolutely right - the refresh token won't be available on the JWT.

OK so it looks like we should change  `setSession()` to `setUser()`, and it will just fill the client with the user credentials (which should then be passed in each HTTP request). 

There will be no way to refresh the session without the access token - but this is fine because it's supposed to be used in a Node context and in that case it will be up to the developer to manage the session/refreshes etc (which they can do with the [api](https://github.com/supabase/gotrue-js/blob/56b6feb44f4e5eb5b166d9512d9a3356bb9e0f0d/src/GoTrueApi.ts#L249)

Iamshankhadeep: @kiwicopple  So what I suggest is that let's have 2 public function `setSession()` depends on `refresh_token`  and `setUser()` depends on `access_token`. but if we use `setUser()`, `expires_in` and `expires_at` and`refresh_token` will be null on the session data and developer needs to make sure that they call the `api` to manage `sessions`.

kiwicopple: That sounds solid - pretty sure that will work for #34

Iamshankhadeep: @kiwicopple can you review the PR?

test/clientWithAutoConfirmEnabled.test.ts
```diff
@@ -51,6 +60,30 @@ test('signUp()', async () => {
...
+test('setUser should update user in session', async () => {
+  const authSession = new GoTrueClient({
+    url: GOTRUE_URL,
+    autoRefreshToken: false,
+    persistSession: false,
+  })
+  const session = authWithSession.session()
+  const { user, error } = await authSession.setUser(session!.access_token)
+  expect(error).toBeNull()
+  expect(user?.user_metadata).toStrictEqual({ hello: 'world' })
```
kiwicopple, line 84: OK this is looking good. 

What happens if you call `authSession.update()` here?

Iamshankhadeep: if we call `authSession.update()` here it will update the user data cause when you call `setUser()` it will update the session also. like this.
``` tsx
const session = { ...this.currentSession, access_token, token_type: 'bearer', user }
      this._saveSession(session)
``` 
so we have access_token in session and we need `access_token` to call `authSession.update()`

src/GoTrueClient.ts
```diff
@@ -251,6 +251,55 @@ export default class GoTrueClient {
     }
   }
 
+  /**
+   * Sets user data on session.
+   * @param access_token a jwt access token
+   */
+  async setUser(access_token: string): Promise<{ user: User | null; error: Error | null }> {
```
easy-street, line 258: one note: the discussion in #34 was centered around potential _synchronous_ solutions.

in a node context, we've extracted a known valid jwt access token from a cookie, and simply need to pass along the jwt in the `Authorization` header when interacting with supabase APIs. ideally, this would be possible without introducing unnecessary `async` data fetching overhead through calls to `api.getUser` / `api.refreshAccessToken`.

kiwicopple: agreed - the goal of this PR should be to set the `auth` headers. 

I have some time today, so I'll jump in here now and help out

Iamshankhadeep: @kiwicopple so then what do you suggest?

kiwicopple: I'll test out your changes in a full `supabase-js` library on a node environment and let you know :)

Iamshankhadeep: yeah that sounds awesome :fire:

kiwicopple:
Hey @Iamshankhadeep @dbredvick I created a new PR here:
https://github.com/supabase/gotrue-js/pull/83

It was cleaner to create it from scratch so I could test with the supabase-js client. Would love your feedback

I noticed a few difficulties that you must have faced @Iamshankhadeep - in particular the `Session` type is expecting a `User`, and I had to make this field "nullable".

Iamshankhadeep:
@kiwicopple so should I close this PR?

kiwicopple:
Can you please keep it open a few days? I might use some of your code :)

Also @Iamshankhadeep , @easy-street , @dbredvick - can you please send us your address/details (PM on twitter) so we can send you some swag? 👕

quicksnap:
Before you close this, this and the other PR seem somewhat relevant to my feature request: https://github.com/supabase/gotrue-js/issues/84

Essentially, I just want to set my session manually, since React Native is not in a browser context and can't slurp the magic link session data from the URL.

I'm happy to contribute, just could use a little guidance. I've only been using supabase client for about a day or so.

kiwicopple:
Great - in this case, I think we should keep the `setSession()` function (but remove the `setUser()` function until we need it)

quicksnap:
I just looked again at the PR here, and I think having this API would solve my use case, though it would trigger an unnecessary token refresh, since the magic link token is already valid.

Iamshankhadeep:
> I just looked again at the PR here, and I think having this API would solve my use case, though it would trigger an unnecessary token refresh, since the magic link token is already valid.

kiwicopple:
Yup I think that would be great @Iamshankhadeep 👍

Iamshankhadeep:
@kiwicopple CI is green. can you review it?

kiwicopple:
seems great @Iamshankhadeep - can you please do one commit with `feat: ` in the title to trigger a version bump when I merge? The commit can be anything (perhaps more details in the JSDoc?)

@quicksnap - are you happy that this solves your use-case? 
It triggers another network request, but it seems safer. The workaround is to make the develop supply an entire `Session` object to a function, but that is a more difficult DX in most cases. 

Better to let the library do the work since logins should be a very "infrequent" task (and we want to ensure it's as error-free as possible). Supplying the `refresh_token` is quite nice since you can use it with OAuth logins too (they return the `refresh_token` once, then the developer could store this somewhere safe for future sessions

quicksnap:
Seems fine to me! In reality, user sign in will be infrequent compared to everything else.

PR #83, opened 2021-05-05
Use in a node context

kiwicopple:
This adds the minimum viable context to set the auth headers required for supabase-js:

https://github.com/supabase/supabase-js/blob/906a076c8942fe86765f0b9430be5b0ad076036a/src/SupabaseClient.ts#L172

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -252,10 +252,10 @@ export default class GoTrueClient {
   }
 
   /**
-   * Overrides the JWT on the current client which will then be used in all subsequent requests.
+   * Overrides the JWT on the current client. The JWT will then be sent in all subsequent network requests.
    * @param access_token a jwt access token
    */
-  setUser(access_token: string): Session {
+  setAuth(access_token: string): Session {
     this.currentSession = {
       ...this.currentSession,
       access_token,
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -252,7 +252,7 @@ export default class GoTrueClient {
   }
 
   /**
-   * Sets the current user.
+   * Overrides the JWT on the current client which will then be used in all subsequent requests.
    * @param access_token a jwt access token
    */
   setUser(access_token: string): Session {
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -252,6 +252,21 @@ export default class GoTrueClient {
   }
 
   /**
+   * Sets the current user.
+   * @param access_token a jwt access token
+   */
+  setUser(access_token: string): Session {
+    this.currentSession = {
+      ...this.currentSession,
+      access_token,
+      token_type: 'bearer',
+      user: null,
+    }
+
+    return this.currentSession
+  }
+
+  /**
    * Gets the session data from a URL string
    * @param options.storeSession Optionally store the session in the browser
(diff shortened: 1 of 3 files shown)


src/lib/types.ts
```diff
@@ -13,9 +13,9 @@ export interface Session {
    * A timestamp of when the token will expire. Returned when a login is confirmed.
    */
   expires_at?: number
-  refresh_token: string
+  refresh_token?: string
   token_type: string
-  user: User
+  user: User | null
```
kiwicopple, line 18: We need to allow the `user` to be null

Also since we can't expect a refresh_token, then that needs to be optional

src/GoTrueClient.ts
```diff
@@ -251,6 +251,21 @@ export default class GoTrueClient {
...
+  setUser(access_token: string): Session {
+    this.currentSession = {
+      ...this.currentSession,
+      access_token,
+      token_type: 'bearer',
+      user: null,
+    }
+
+    return this.currentSession
+  }
```
kiwicopple, line 267: The code in supabase-js works like this:

```js
private _getAuthHeaders(): { [key: string]: string } {
    const headers: { [key: string]: string } = {}
    const authBearer = this.auth.session()?.access_token ?? this.supabaseKey
    headers['apikey'] = this.supabaseKey
    headers['Authorization'] = `Bearer ${authBearer}`
    return headers
  }
```

so really all we need to do is ensure `this.auth.session()?.access_token` will return a JWT.

This function is synchronous, so it isn't doing anything smart like checking the user is valid (perhaps the name of the function isn't great then?)

test/clientWithAutoConfirmEnabled.test.ts
```diff
@@ -61,6 +63,23 @@ test('signUp() the same user twice should throw an error', async () => {
...
+    url: GOTRUE_URL,
+    autoRefreshToken: false,
+    persistSession: false,
+  })
+
+  newClient.setUser(access_token!)
+
+  const authBearer = newClient.session()?.access_token
+  expect(authBearer).toEqual(access_token)
+})
```
kiwicopple, line 80: here I 

1. create  a fresh client
2. set it to the previous user's jwt
3. check that `newClient.session()?.access_token` returns the correct jwt

src/GoTrueClient.ts
kiwicopple: Perhaps this is better as `setAuth()` or `setToken()`?

awalias: hmm I think any are fine tbh:
- `setUser` suggests how this is most commonly used by most people (but in the future perhaps not all)
- `setAuth` since it's setting the Auth bearer header, I guess we can assume this will always be the case
- `setToken` again makes sense from a consumption point of view, maybe even `setAccessToken`

kiwicopple: OK will let the guys from the other PR (@ Iamshankhadeep @easy-street @@dbredvick ) give input, then i can change

Iamshankhadeep: I guess `setAccessToken` or `setAuth` works fine.

kiwicopple: OK I'll go with `setAuth()` for now - then if we add extra items like refresh_tokens it is more generic

Issue #69, opened 2021-03-27
add method to delete users


awalias:
on the backend you can:

[`DELETE /admin/users/{user_id}`](https://github.com/supabase/gotrue/blob/8cb3d71a8c3fcbd9fd0e23586f1203d5861a56a5/api/api.go#L157) but the JWT role needs to be `service_role` or `supabase_admin`

perhaps we should add it as a method here

linked: https://github.com/supabase/gotrue/issues/79

marcelobarreto:
Is anyone working at this? Otherwise, you can assign it to me @awalias :call_me_hand:

marcelobarreto:
```js
const { GoTrueClient } = require('./dist/main');

const auth = new GoTrueClient({
  url: 'https://uedcsccyuwzjvimjgprm.supabase.co/auth/v1',
  headers: {
    apikey: 'my_service_role_JWT',
  },
});

async function main() {
  // I've created this .destroyUser
  let { data, user, error } = await auth.destroyUser({ userUID: '1bd282ce-3735-472f-a55f-8745ef79551d' });
  console.log(data, user, error);

  process.exit(0)
};

main();

// null null { message: 'User not allowed', status: 401 }
```

@awalias can you help to see what I'm doing wrong?

awalias:
ah this also needs the backend change mentioned here: supabase/gotrue#79 in order to work

awalias:
this comment: https://github.com/supabase/gotrue/issues/79#issuecomment-808719524

awalias:
oh sorry you're already using the `service_role` 🤔

awalias:
it should be being permitted by this: https://github.com/supabase/gotrue/blob/08c7a29fd40bd40ab337f31710c0345ff2970798/conf/configuration.go#L194

awalias:
oh! can you try:
```
const auth = new GoTrueClient({
  url: 'https://uedcsccyuwzjvimjgprm.supabase.co/auth/v1',
  headers: {
    apikey: 'my_service_role_JWT',
    Authorization: `Bearer ${my_service_role_JWT}` <---- this is required
  },
});
```

marcelobarreto:
Interesting... I decoded my JWT payload:

![image](https://user-images.githubusercontent.com/1276556/117454544-a2310a00-af1c-11eb-825b-eacc739f1f50.png)

:thinking: :thinking: :thinking:

marcelobarreto:
[...]

PR #85, opened 2021-05-07
feat: delete user

marcelobarreto:

Delete user feature

#69 

## What is the new behavior?

Now, we can remove users by it's uid

## Additional context

### DIFF
--- a/src/GoTrueApi.ts
+++ b/src/GoTrueApi.ts
@@ -1,4 +1,4 @@
-import { get, post, put } from './lib/fetch'
+import { get, post, put, destroy } from './lib/fetch'
 import { Session, Provider, UserAttributes, CookieOptions, User } from './lib/types'
 import { COOKIE_OPTIONS } from './lib/constants'
 import { setCookie, deleteCookie } from './lib/cookies'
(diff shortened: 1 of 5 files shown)


marcelobarreto:
@awalias Hey Ant, do you that this PR satisfy #69 ?
I'd like to write some tests for it first, but just trying to make sure that I'm not forgetting about anything else.

marcelobarreto:
done @kiwicopple :muscle:

kiwicopple:
🔥 :shipit:

kiwicopple:
@marcelobarreto - i was thinking more about this when I was documenting it. I don't think we need the `deleteUser()` method inside the Client - we should only really call it via the API.

The key idea of the `Client` is to manage sessions. Any generic user-management should happen through the `Api`.

marcelobarreto:
Ok that makes sense @kiwicopple , thanks for handling it

nicholaschiang:
@kiwicopple is [this documentation](https://supabase.io/docs/reference/javascript/delete-user) up-to-date then?

```javascript
const { user, error } = await supabase.auth.api.deleteUser(
  '715ed5db-f090-4b8c-a067-640ecee36aa0',
  'YOUR_SERVICE_ROLE_KEY'
)
```

I'm guessing it's just a wrapper around the [`DELETE /admin/users/{user_id}`](https://github.com/supabase/gotrue/blob/8cb3d71a8c3fcbd9fd0e23586f1203d5861a56a5/api/api.go#L157) API endpoint. If so, and if there are other wrappers like this out there (e.g. [to programmatically create users](https://github.com/supabase/gotrue/blob/8cb3d71a8c3fcbd9fd0e23586f1203d5861a56a5/api/api.go#L150) without them having to login), I think it'd be good to document all of them.

PR #88, opened 2021-05-22
fix: /invite API (inviteUserByEmail) returns a User

Benjamin-Dobell:

Corrected the return type of `inviteUserByEmail`.

The `/invite` API returns a User object as per https://github.com/supabase/gotrue/blob/master/api/invite.go#L73

`inviteUserByEmail` returned promise result `data` property is typed as `{}`, rather than `User`.

## What is the new behavior?

`inviteUserByEmail` returned promise result `data` property is now typed as `User`, matching the data returned from the `/invite` API.

### DIFF
--- a/src/GoTrueApi.ts
+++ b/src/GoTrueApi.ts
@@ -121,7 +121,7 @@ export default class GoTrueApi {
     options: {
       redirectTo?: string
     } = {}
-  ): Promise<{ data: {} | null; error: Error | null }> {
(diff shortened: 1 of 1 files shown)


cdbattags:
So it looks like the `inviteUserByEmail` `raw_user_metadata` is not currently getting sent in the response

cdbattags:
On 2.0.4 supabase-js which is  `"@supabase/gotrue-js": "^2.2.0",`

Issue #65, opened 2021-02-24
redirectTo does not seem to work (because it's forbidden)


laurentS:
# Bug report

## Describe the bug

I've been going a little insane over this `redirectTo` parameter with auth calls. Whatever I tried passing was not getting sent to the server, despite the fix for #63.
It turns out that setting the `Referer` header programmatically is not allowed: https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name

## To Reproduce

See the steps in #63, but upgrade supabase-js to 1.4.1.
The magic link `redirect_to` parameter contains the calling url, not the one passed to gotrue's `redirectTo`.
A warning is issued in firefox's console: `Attempt to set a forbidden header was denied: referer`

## Expected behavior

The redirection link in the email should match what is passed to the auth call. I'd suggest either passing a different, custom header, or passing the url as part of the body of the request, whichever requires the least changes on the server side.

## System information

- OS: debian/linux
- Browser firefox 84
- Version of supabase-js: 1.4.1
- gotrue-js version 1.11.2

kiwicopple:
Hey @laurentS, can I please confirm that you are trying to redirect to a path that is the same as the URL/origin in your auth Settings?

From the [docs](https://supabase.io/docs/client/auth-signup#sign-up-with-redirect):

> the base domain must be the same as the domain in your Auth settings

So for settings like this:

![image](https://user-images.githubusercontent.com/10214025/108954270-851fc580-76a7-11eb-86ce-e8c19987cb4b.png)


I can redirect to:

- `https://supabase.io/welcome`
- `https://supabase.io/magic`

But I cannot redirect to:

- `https://evil.io/welcome`


We will eventually enable multiple origins: https://github.com/supabase/gotrue/issues/63

laurentS:
@kiwicopple yes, correct. I have `http://localhost:3000/` set as the site URL in the auth settings (I had a different one initially, like `http://localhost:3000/sendmehere` but all on the same domain), and I'm trying to redirect to `http://localhost:3000/someotherurl`. Having logged the various calls through the supabase stack, I can see that the redirect url I'm setting "someotherurl" passed all the way down to `params` here https://github.com/supabase/gotrue-js/blob/d993b5457e8c54d4e9b6d4ff221e28c1adc2964d/src/lib/fetch.ts#L37

This what `params` looks like:
```json
{
  "method": "POST",
  "headers": {
    "Content-Type": "text/plain;charset=UTF-8",
[...]

PR #89, opened 2021-05-26
fix: #65 sends redirect in query param instead of referer header

awalias:
#65

### DIFF
--- a/infra/docker-compose.yml
+++ b/infra/docker-compose.yml
@@ -6,6 +6,7 @@ services:
     ports:
       - '9999:9999'
     environment:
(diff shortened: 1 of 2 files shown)


soedirgo (approved):
LGTM! One thing that comes to mind is I used the [URL API](https://developer.mozilla.org/en-US/docs/Web/API/URL_API) for handling query parameters in postgrest-js (e.g. [here](https://github.com/supabase/postgrest-js/blob/e14461cb752e603c6dca9ca114bc882a3b137636/src/lib/PostgrestFilterBuilder.ts#L64)), though I'm not sure if that's even a problem here.

Issue #90, opened 2021-05-26
RFC: Split client-side SDK from admin (service role) SDK


nicholaschiang:
## Describe the problem

Currently, the admin methods like ["Delete user"](https://supabase.io/docs/reference/javascript/delete-user) that require a full-access API key (e.g. `service_role` key) are documented side-by-side with the more client facing methods like [`signUp`](https://supabase.io/docs/reference/javascript/auth-signup) or [`signOut`](https://supabase.io/docs/reference/javascript/auth-signout) (that I think only require a public API key).

I will likely only use the admin methods in a server-side environment (because I should never send a `service_role` key to a client). I will likely **not** use any of the client-side methods (e.g. `signUp`, `signIn`, `signOut`) in that server-side environment. For the same reasons, I will never use the admin methods in a client-side environment. Thus, it doesn't make any sense to increase the client-side bundle by including methods that will only ever be called on the server-side (and vice versa).

## Describe the improvement

I think this library should be refactored into two separate JavaScript libraries:
1. A Node.js admin SDK - For server-side environments, only includes admin methods that require a `service_role` key.
2. A JavaScript client-side SDK - To be bundled with client-side code, only includes methods called client-side.

And then the documentation should be updated accordingly: split (much like the Firebase Authentication [docs](https://firebase.google.com/docs/auth/admin) and [reference](https://firebase.google.com/docs/reference/admin)) into two sections, one for the admin SDK and one for the client-side SDK.

## Additional context

This is how the Firebase Authentication SDKs are structured and I think it makes a lot of sense to reproduce a similar structuring here with Supabase Auth. Especially because I think that every REST API method included in the GoTrue `/admin` endpoints should be reproduced in a server-side admin SDK (e.g. like how the [`DELETE /admin/users/{user_id}`](https://github.com/supabase/gotrue/blob/8cb3d71a8c3fcbd9fd0e23586f1203d5861a56a5/api/api.go#L157) was wrapped in #85).

For example, I'd like to be able to [programmatically create new users](https://github.com/supabase/gotrue/blob/8cb3d71a8c3fcbd9fd0e23586f1203d5861a56a5/api/api.go#L150) without them having to login client-side.

kiwicopple:
Agreed @nicholaschiang .

We actually have the gotrue library documented separately here:
https://supabase.io/docs/gotrue/client/gotrue-client

But I think we should do something similar on the supabase-js wrapper.

nicholaschiang:
[...]

PR #97, opened 2021-06-04
feat: add admin 'createUser' method

nicholaschiang:

Defines a wrapper around the [`POST /admin/users`](https://github.com/supabase/gotrue/blob/master/api/api.go#L151) API endpoint.

There is no way to programmatically create new users from this SDK in a server-side environment.

## What is the new behavior?

You can create new users in Node.js (server-side admin environment):

```javascript
import { GoTrueClient } from '@supabase/gotrue-js'

const GOTRUE_URL = 'http://localhost:9999'

const auth = new GoTrueApi({ url: GOTRUE_URL })
const { data, error } = await auth.createUser(jwt, { email: 'john@example.com', password: 'password' });
```

## Additional context

Related to #90. I can't wrap the `GET /admin/users/{user_id}` endpoint because the `getUser()` method is already defined. Instead, there should be a separate class that only contains the admin API endpoints (i.e. those requiring `service_role` keys).


kiwicopple:
Hey @nicholaschiang - this one looks great, but we might need a test to confirm it's working well and prevent regression in the future. Agreed on #90 too.

Thanks for this contribution 🙏

kiwicopple:
this has been resolved in another PR!

PR #98, opened 2021-06-12
fix: supabase/supabase#1915 encode query string parameter values

thebengeu:

Encodes query string parameter values.

supabase/supabase#1915: redirectTo works but cuts the "and" params "&param2=val2&param3=val3".

## What is the new behavior?

`&` and other special characters are encoded.

## Additional context

I considered using [balazsbotond/urlcat](https://github.com/balazsbotond/urlcat) for readability and concision.

E.g.:

```typescript
let urlParams: string[] = [`provider=${encodeURIComponent(provider)}`]
if (options?.redirectTo) {
  urlParams.push(`redirect_to=${encodeURIComponent(options.redirectTo)}`)
}
if (options?.scopes) {
}
return `${this.url}/authorize?${urlParams.join('&')}`
```

Would become:

```typescript
options.provider = provider
return urlcat(this.url, '/authorize', options)
```

However, the increase in bundle size would probably not be worth it.

### DIFF
--- a/src/GoTrueApi.ts
+++ b/src/GoTrueApi.ts
@@ -47,7 +47,7 @@ export default class GoTrueApi {
       let headers = { ...this.headers }
       let queryString = ''
       if (options.redirectTo) {
-        queryString = '?redirect_to=' + options.redirectTo
+        queryString = '?redirect_to=' + encodeURIComponent(options.redirectTo)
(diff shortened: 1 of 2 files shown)


awalias:
nice one @thebengeu :)

PR #100, opened 2021-06-13
fix: refreshTokenTimer blocks node event loop exit

### LINKED ISSUE #99: Something in `signIn` hanging my script?
I noticed that this script never exits, I wondered whether it might be something inside the `signIn` call

https://replit.com/@awalias/supabase-node-auth#index.js

easy-street:
Bug fix

fixes #99

when `autoRefreshToken` is `true` (the default), the `signIn` method starts a timer that prevents the node event loop from exiting.

## What is the new behavior?
this PR preserves the defaults, while allowing the node process to immediately exit. this change should not functionally affect the previous behavior of the timer in long-running node processes.

in node, we can ensure that the timeout will not require the Node.js event loop to remain active by immediately calling [`unref`](https://nodejs.org/dist/latest-v14.x/docs/api/timers.html#timers_timeout_unref). `unref` should be used judiciously, but it seems to fit well here.

## Additional context
alternatives to consider instead of making this change:
- add docs guidance suggesting `autoRefreshToken: false` when appropriate in node environments
- change the `autoRefreshToken` default to `false` in node environments
- always skip starting the timer in node environments (meaning that `autoRefreshToken` would have no effect in node)

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -521,5 +521,6 @@ export default class GoTrueClient {
     if (!value || !this.autoRefreshToken) return
 
     this.refreshTokenTimer = setTimeout(() => this._callRefreshToken(), value)
+    if (typeof this.refreshTokenTimer.unref === 'function') this.refreshTokenTimer.unref()
   }
 }


awalias:
nice catch @easy-street !!

PR #101, opened 2021-06-14
fix: supabase/supabase-js#178 handle JWT expiry <= 60s

thebengeu:

Handles JWT expiry <= 60s correctly.

supabase/supabase-js#178: Refresh token doesn't update when JWT expiry is set to 60, constantly updates when JWT expiry is set to less than 60 since `expiresAt - timeNow - 60` is negative.

## What is the new behavior?

When JWT expiry is set to more than 60, current behavior remains to refresh 60s before expiry. When JWT expiry is set to 60 or less, new behavior is to refresh 0.5s before expiry.

## Additional context

0.5s is chosen to support JWT expiry of 1s, even though that should be an edge case. That should usually still be enough time to reach the server before expiry.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -548,8 +548,12 @@ export default class GoTrueClient {
     this.currentUser = session.user
 
     const expiresAt = session.expires_at
-    const timeNow = Math.round(Date.now() / 1000)
-    if (expiresAt) this._startAutoRefreshToken((expiresAt - timeNow - 60) * 1000)
+    if (expiresAt) {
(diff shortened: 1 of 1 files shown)


awalias:
on fire @thebengeu 🔥

awalias (approved):

awalias:
looks like there is some failing check https://github.com/supabase/gotrue-js/pull/101/checks?check_run_id=2816594247#step:4:49

thebengeu:
Fixed.

kiwicopple:
legend, thanks @thebengeu

PR #106, opened 2021-07-21
Feat/mobile generate links

awalias:
supports new features in gotrue v2.0.5


src/GoTrueClient.ts
```diff
@@ -417,7 +417,7 @@ export default class GoTrueClient {
       })
       if (error || !data) return { data: null, user: null, session: null, error }
 
-      if (data?.user?.confirmed_at) {
+      if (data?.user?.email_confirmed_at) {
```
kiwicopple, line 420: this looks like a breaking change 

We will need to support old versions of GoTrue as well right?

So `data?.user?.confirmed_at || data?.user?.email_confirmed_at`

Side note - we didnt remove `confirmed_at` from GoTrue right?

kiwicopple: do we need to include `phone_confirmed_at`?

src/lib/types.ts
```diff
@@ -33,9 +33,13 @@ export interface User {
   }
   aud: string
   confirmation_sent_at?: string
+  recovery_sent_at?: string
+  action_link?: string
   email?: string
+  phone?: string
   created_at: string
-  confirmed_at?: string
+  email_confirmed_at?: string
```
kiwicopple, line 41: we should add back `confirmed_at` and make it deprecated 

https://stackoverflow.com/questions/60755711/is-it-possible-to-mark-something-as-deprecated-in-typescript

kiwicopple (changes requested):

src/lib/types.ts
awalias: as discussed with @kiwicopple , on the assumption that some people might be using the `confirmed_at` column directly we can add it back in as a column in the auth schema as a generated column - which takes the minimum date of `email_confirmed_at`, and `phone_confirmed_at`, I'll start looking into it now (https://github.com/supabase/gotrue/issues/152)

https://www.codeproject.com/Articles/5246207/Generated-Columns-in-PostgreSQL

src/GoTrueClient.ts
```diff
@@ -195,7 +208,57 @@ export default class GoTrueClient {
...
+    }
+  }
+
+  /**
+   * Log in a user given a User supplied OTP received via mobile.
+   * @param phone The user's phone number.
+   * @param token The user's password.
+   * @param redirectTo A URL or mobile address to send the user to after they are confirmed.
+   */
+  async verifyOTP(
```
awalias, line 223: left this one generic so we have the option to add other types of OTP tokens later

infra/docker-compose.yml
```diff
@@ -31,8 +31,15 @@ services:
       GOTRUE_SMTP_PASS: GOTRUE_SMTP_PASS
       GOTRUE_SMTP_ADMIN_EMAIL: admin@email.com
       GOTRUE_MAILER_SUBJECTS_CONFIRMATION: 'Please confirm'
+      GOTRUE_EXTERNAL_PHONE_ENABLED: 'true'
```
kangmingtay, line 34: just to check if we're on the same page, we don't set `GOTRUE_EXTERNAL_EMAIL_ENABLED='true'` because gotrue sets it to true by default

awalias: yea 👍

src/GoTrueApi.ts
```diff
@@ -334,4 +415,39 @@ export default class GoTrueApi {
...
+  /**
+   * Generates links to be sent via email or other.
+   * @param type The link type ("signup" or "magiclink" or "recovery" or "invite").
+   * @param email the user's email.
+   * @param password user password for signup only.
+   * @param data optional user metadata for signup only.
+   * @param redirectTo The link type ("signup" or "magiclink" or "recovery" or "invite").
+   */
+  async generateLink(
+    type: string,
```
kiwicopple, line 428: we should make this one an enum so it flows through to the docs

`type: 'signup' | 'magicLink' | 'recovery' | 'invite'`

src/GoTrueApi.ts
```diff
@@ -334,4 +415,39 @@ export default class GoTrueApi {
       return { user: null, data: null, error }
     }
   }
+
+  /**
+   * Generates links to be sent via email or other.
+   * @param type The link type ("signup" or "magiclink" or "recovery" or "invite").
+   * @param email the user's email.
```
kiwicopple, line 422: Can we format these with caps at the start of the description?

`@param email The user's email.`

src/GoTrueApi.ts
```diff
@@ -334,4 +415,39 @@ export default class GoTrueApi {
...
+   * @param password user password for signup only.
+   * @param data optional user metadata for signup only.
```
kiwicopple, line 424: `@param data Optional user metadata. For signup only.`

src/GoTrueClient.ts
```diff
@@ -107,9 +109,13 @@ export default class GoTrueClient {
     try {
       this._removeSession()
 
-      const { data, error } = await this.api.signUpWithEmail(email!, password!, {
-        redirectTo: options.redirectTo,
-      })
+      if (phone && password) {
+        var { data, error } = await this.api.signUpWithPhone(phone!, password!)
```
kiwicopple, line 113: this could be `const`

src/GoTrueClient.ts
```diff
@@ -107,9 +109,13 @@ export default class GoTrueClient {
...
+      } else {
+        var { data, error } = await this.api.signUpWithEmail(email!, password!, {
```
kiwicopple, line 115: `const`

kiwicopple (approved):

PR #112, opened 2021-07-29
#107 : remove only Gotrue hash parameters from url after setting sess…

### LINKED ISSUE #107: Url hash is totally removed when setting session from Url
# Bug report

## Describe the bug

When the front-end uses a hash-based router like https://github.com/ItalyPaleAle/svelte-spa-router , this line breaks the routing by removing the hash from the url :  
https://github.com/supabase/gotrue-js/blob/b887af5567409a189b0e700880a032c65d0d21e8/src/GoTrueClient.ts#L358

For example, let's say that when a user receives a recover password link, we want to redirect him to the page http//frontend.com/#/password which contains a reset password form.
- The user receives the recovery link by email and clicks it : http://example.supabase.io/auth/verify?token=qCjkmx2wzPO7sFFa2eDHvQ&type=recovery&redirect_to=http://frontend.com/
- The user is redirected by supabase to http://frontend.com/#access_token=...&type=recovery ...
- The app detects the access_token and type=recovery and sets the route with http://frontend.com/#/password  . In the meanwhile, gotrue.js detects the access_token, sets the session, and removes the hash from the url, so it routes the user to http://frontend.com/
- The user never sees the reset password screen.

## Expected behavior

Developers should be able to chose if they want the full hash to be removed from the url or not. This could be parameterized in the Gotrue client instantiation . 

Also, maybe the access_token should be placed after a '?' rather than a '#' . I have seen code changes in this way ( https://github.com/supabase/gotrue/pull/97 ) , but they don't seem to be in production yet. When they will be, [this line](https://github.com/supabase/gotrue-js/blob/b887af5567409a189b0e700880a032c65d0d21e8/src/GoTrueClient.ts#L358) will not be relevant anymore as it won't remove the gotrue parameters from the url.

tartard:
…ion from Url.

Fixes #107 

After setting session from url, gotrue-js removes all the url hash. This can be a problem if the app uses a hash-based router, this would automatically redirect the user at the root location '#/' . This might not be the desired behavior for the developer, and currently the developer has no control over this behavior.

## What is the new behavior?

After setting session from url, gotrue-js removes only the Gotrue parameters from the hash. This doesn't change other things which might be in the hash, like the path of the router.


src/lib/helpers.ts
```diff
@@ -23,6 +23,27 @@ export function getParameterByName(name: string, url?: string) {
   return decodeURIComponent(results[2].replace(/\+/g, ' '))
 }
 
+// #107 Only remove Gotrue parameters from url hash.
```
sduduzog, line 26: Reading the code and the unit tests for it, I'm able to figure out what it does so I see the comment as being a little noisy.

But I see how also it is useful, as it points out to where the issue is from.

src/lib/helpers.ts
```diff
@@ -23,6 +23,27 @@ export function getParameterByName(name: string, url?: string) {
...
+    'expires_in',
+    'refresh_token',
+    'token_type'
+  ]);
+}
+
+export function removeParamsFromUrl(url: string, params: string[]) {
+  let result = url;
+  params.forEach(param => {
+    const regexp = new RegExp("(&|#|\\?)" + param + "((\\=.*?(?=&|$))|(?=&|#))");
```
sduduzog, line 41: A suggestion to make the regex case insensitive

test/helpers.test.ts
```diff
@@ -0,0 +1,24 @@
+import {removeParamsFromUrl} from "../src/lib/helpers";
+
+
+test('testRemoveParamsFromUrl()', async () => {
```
sduduzog, line 4: A suggestion to have 2 of these tests blocks.

One for all the cases `removeParamsFromUrl` mutates the url and one for cases where it is not supposed to.

This will remove the need to have single line comments 

i.e.

```js
test('removeParamsFromUrl() should not remove path information', () => {
...
})
```

sduduzog (changes requested):
I just have a couple of suggestions that I think will make this PR a bit clearer and readable

src/lib/helpers.ts
```diff
@@ -23,6 +23,27 @@ export function getParameterByName(name: string, url?: string) {
   return decodeURIComponent(results[2].replace(/\+/g, ' '))
 }
 
+// #107 Only remove Gotrue parameters from url hash.
+export function clearGotrueHashParameters() {
+  window.location.hash = removeParamsFromUrl(window.location.hash, [
```
sduduzog, line 28: Also, just a question here, would it be possible to rather get the parameters from the Session object? This will make sure that should there be any new property added or property removed, this would always work.

i.e. `Object.keys(session)` and then pass them to the function

tartard: Yes good idea, I didnt' think about it.

src/lib/helpers.ts
tartard: Ok I removed it as there are no such comments elsewhere and it would become quickly noisy if we had a comment for each issue. Also git blame & git history provide these informations.

test/helpers.test.ts
tartard: OK, done

src/lib/helpers.ts
tartard: OK, done

sduduzog (approved):
Great work on this PR, Remember, I'm not part of the team or one of the maintainers but I just stepped up to help because I'm also interested in this solution

DavraYoung:
Really waiting for this pr to be merged. My Github Pages application cannot be connected via OAuth2 because of this hash removal bug. 
Also, why don't gotrue  itself use query params for this purpose?

test/helpers.test.ts
```diff
@@ -0,0 +1,28 @@
+import {removeParamsFromUrl} from "../src/lib/helpers";
+
+
+test('removeParamsFromUrl() should remove Gotrue params from hash, ignoring case', async () => {
+    expect(removeParamsFromUrl('/some/path#p1=value&p2=value', ['p1', 'p2']))
+        .toBe('/some/path');
+
+    expect(removeParamsFromUrl('/some/path#p1=value&p2=value', ['p1']))
+        .toBe('/some/path&p2=value');
```
awalias, line 9: is this resulting url `/some/path&p2=value` invalid?

awalias:
thanks everyone for pitching in, we'll try and get this out asap

test/helpers.test.ts
sduduzog: Aye, nice catch @awalias Looks like the url format is off a bit. The # should not be removed

tartard: Actually I think that in this case the hash  should be removed, at least it is the behavior that I expect personnaly, because here the hash identifies p1 as a gotruejs param and not a URL hash. I will add a test case with : 
/some/path#somehash#p1=value&p2=value which should become /some/path#somehash&p2=value when we remove p1 param.

sduduzog: @tartard  Understandable, but then everything after the `#` is part of the hash string and will never be part of the request sent to the server on browser refresh, but /some/path&p2=value on the other hand suggests that p2 is a query parameter, of which it is not. Essentially, nothing else goes after the hash string

tartard: As a reminder, my initial case is that I am using a hash based router in my client webapp, and gotrue-js breaks the routing when it removes all the hash. My fix aims to make gotrue-js remove only the parameters it uses in order to be able to set a SITE_URL=http://host/#/auth/ for example in supabase. But since the first gotrue param is always preceded by a '#', gotrue would redirect to http://host/#/auth#access_token=...
That is why I want gotrue-js to remove the '#access_token...' part, including the '#', and not the full ''#/auth/#access_token...' as it is currently doing. 
If I understand you well, I see your point but I think it would not be a problem since everything after the #access_token... is set by gotrue and will be removed by gotrue-js after successfully setting the session. There is no real case where the 
'#p1' would be removed but not the '&p2'. I just made this test case to show the precise behavior of my code.
But I understand that it can be confusing, so maybe I should change my code to not remove params one by one, but rather remove a hash and everything after this hash IF this hash is immediately followed by a gotrue param (like 'access_token'. 
Another solution could be to make gotrue (backend) precede its params with a '&' if a hash is already present in the redirect url.

sduduzog: > My fix aims to make gotrue-js remove only the parameters it uses in order to be able to set a SITE_URL=http://host/#/auth/ for example in supabase. But since the first gotrue param is always preceded by a '#', gotrue would redirect to http://host/#/auth#access_token=

This should also be a test case on it's own as it's pretty different from all the available ones... Where there's multiple `#` and I still believe it could be achieved by this PR

src/lib/helpers.ts
```diff
@@ -22,3 +24,61 @@ export function getParameterByName(name: string, url?: string) {
...
+
+export function removeHashStartingWith(url: string, hashStart: string) {
+  if(hashStart && hashStart.trim().length) {
+    const regexp = new RegExp("#" + hashStart + ".*$", "i");
+    return url.replace(regexp, '');
+  }
+  return url;
+}
+
+export function removeHashStartingWithAnyOf(url: string, starts: string[]) {
```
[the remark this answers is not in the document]
tartard: Hello, sorry for the delay. 
Well, actually `removeParamsFromUrl` removes only a single "parameter", ie any key - value which starts with a '#', a '&' or a '?' . 
In the other hand, `removeHashStartingWithAnyOf` removes the whole hash if it starts with '#' plus one of the given strings.

kangmingtay:
hey @tartard and @sduduzog, thanks for taking the time to work on this PR. Unfortunately, i don't think this is the appropriate fix to enable your use-case with hash-based redirects in svelte.

@DavraYoung the tokens are placed behind a query fragment because gotrue follows an oauth2 protocol known as the [implicit grant flow](https://datatracker.ietf.org/doc/html/rfc6749#section-4.2).

I've tried to reproduce the following cases with the correct redirect urls in the allow list:
1. `https://project-ref.supabase.co/auth/v1/authorize?provider=google&redirect_to=http://localhost3000/#/password`
Redirected to: `http://localhost3000/#`. Note that Gotrue doesn't even include the `/#/password` in the query fragment during the redirect here. The redirected url initially contains the `#access_token=..... fragments` but not the /password path.
2. `https://project-ref.supabase.co/auth/v1/authorize?provider=google&redirect_to=http://localhost3000/welcome`
Redirected to: `http://localhost3000/welcome#`. This is expected and correct.

Will be transfer this to the gotrue backlog.

PR #117, opened 2021-08-05
Access localStorage lazily to avoid DOMExceptions in constrained envs when `localStorage` is provided using options

Andarist:

Bug fix

in 3rd party iframes in the incognito mode accessing `localStorage` might throw:
> Uncaught DOMException: Failed to read the 'localStorage' property from 'Window': Access is denied for this document
Screenshot 2021-08-05 at 14 40 22

## What is the new behavior?

Accessing `localStorage` has been made **lazy** (has been removed from the synchronous initialization path of the module to the constructor of one of the classes). This only helps when one provides `localStorage` through options (which will benefit me because I can do that to avoid that unsafe access). You could try to improve the situation for all users as well but you would have to figure out what would be the most appropriate solution for your use case.

## Additional context

I've also refactored the code related to `localStorage` a little bit:
1. I'm creating a custom type for what you are actually using - when I provide a custom `localStorage` using options I don't want to satisfy the whole `Storage` interface because I have no use for it. Ideally, I'd only have to provide what you actually need here. Right now I'm casting with `as Storage` but that's unsafe (if you start using something in the future that I have not covered in my implementation). This also actually starts accepting `AsyncStorage` without the need to cast it to `any` and removes TS warnings that `await` does not have any effect on the calls to `this.localStorage`. The end result is that this is more type-safe and correct
2. In the process I've also removed a custom `LocalStorage` class. The only thing that it was doing was to proxy calls to the internal `this.localStorage` (the provided one or the DOM one) without actually doing anything else - this was introducing an additional function layer for all the calls to the `localStorage`. This ain't that big of a problem on its own but I found this code to be not necessary and its presence made my type-changes more complicated. I would also have to drop half of its implementation anyway because my type changes focus on using only the required `localStorage` methods and this class was over-implementing the `Storage` interface

### DIFF
--- a/example/expo/App.tsx
+++ b/example/expo/App.tsx
@@ -7,12 +7,11 @@ import AsyncStorage from '@react-native-community/async-storage'
 import Button from './components/Button'
 import TextField from './components/TextField'
 
-const auth = new GoTrueClient({ localStorage: AsyncStorage as any })
+const auth = new GoTrueClient({ localStorage: AsyncStorage })
 
 export default function App() {
   const [email, setEmail] = React.useState('')
   const [password, setPassword] = React.useState('')
-
 
   return (
     <View style={styles.container}>
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -1,5 +1,5 @@
 import GoTrueApi from './GoTrueApi'
-import { isBrowser, getParameterByName, uuid, LocalStorage } from './lib/helpers'
+import { isBrowser, getParameterByName, uuid } from './lib/helpers'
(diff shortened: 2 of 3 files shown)


kiwicopple:
This looks like a fantastic refactor

@phamhieu - any opinions here? Otherwise I think we can merge?

phamhieu (approved):
lgtm 🚀

thanks @Andarist

Andarist:
@phamhieu any updates on merging (& releasing) this? :)

phamhieu:
Sorry for the late reply. I merged your PR. 

There is an issue with semantic release https://github.com/supabase/gotrue-js/issues/135. Will need to resolve it first

PR #120, opened 2021-08-18
chore: adds X-Client-Info header to requests

awalias:
end result is: `X-Client-Info: gotrue-js/x.x.x`

automatically initializes `version.ts` to whatever version is in `package.json` on build

this allows us to use custom versions in local development and also won't break for forks etc.

then the second step happens at release time (when we learn the release version number), the semantic-release will replace a `0.0.0` version number in the dist files with the actual version number

tested end to end via https://www.npmjs.com/package/@awalias-org/gotrue-js

it's added to the client in such a way that it can be easily overridden manually or when invoked from `supabase-js`

once approved I'll do similar across the other repos


inian (approved):

inian:
Looks good Ant!

PR #127, opened 2021-08-27
fix: make supabase.auth.onAuthStateChange work across browser windows

i-pip:

Bug fix: ensure onAuthStateChange works across multiple browser windows 

Addresses [#2739](https://github.com/supabase/supabase/issues/2739)  - "onAuthStateChange does not catch auth state changes across multiple windows"

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -85,6 +85,20 @@ export default class GoTrueClient {
     } catch (error) {
       console.log('Error getting session from URL.')
     }
+    if (isBrowser()) {
+      window.onstorage = (e: StorageEvent) => {
+        if (e.key === 'supabase.auth.token') {
+          const newSession = JSON.parse(String(e.newValue))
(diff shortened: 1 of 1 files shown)


Andarist:
Note that it might not work if a custom `localStorage` is provided. So I wonder if this shouldn't listen on a separate key and not on the one responsible for storing the token. That way you could still preserve the token in a different type of storage (like cookies) while being able to broadcast the notification using the localStorage "channel"

kiwicopple:
@Andarist I think for that to work the other storage provider would need to implement some sort of pub/sub functionality? 

I can't think of a way that this could work with cookies - do you have some code that you can share which would clarify how it might function?

kiwicopple:
Thanks for this @i-pip ! I added your code here : https://github.com/supabase/gotrue-js/pull/184

Made a new PR so I could tidy it up a bit. If you get a chance would love you to test out that PR and make sure it's good !

PR #128, opened 2021-08-29
feat: allow clients to provide an hcaptcha response token

darora:
Building off https://github.com/supabase/gotrue/pull/192


darora:
Don't necessarily think this is the cleanest interface to expose

src/GoTrueApi.ts
```diff
@@ -41,6 +41,7 @@ export default class GoTrueApi {
     password: string,
     options: {
       redirectTo?: string
+      hcaptchaToken?: string
```
kangmingtay, line 44: nit: should we just call it `captchaToken` since we are open to exposing other captcha providers in the future?

kangmingtay (approved):

awalias (approved):
conflict is with another option `data` , does this also solve for the hcaptcha @inian or do we want both?

```
<<<<<<< da/hcaptcha-support
   * @param hcaptchaToken The response provided by hcaptcha after the challenge is verified.
=======
   * @param data Optional user metadata.
>>>>>>> master
```

darora:
@awalias I don't think we want to shove captcha stuff into the data object that's utilized by the underlying endpoints. We could, however, nest it under a `gotrue_security` or similar field, to consolidate any other similar things we might add in the future?

darora:
@kangmingtay @awalias updated for Gotrue 2.1.7

src/GoTrueApi.ts
```diff
@@ -48,6 +48,7 @@ export default class GoTrueApi {
     options: {
       redirectTo?: string
       data?: object
+      hcaptchaToken?: string
```
kiwicopple, line 51: can we update this to just "captcha"? Should it be tied to a specific captcha "type"? If we plan to eventually provide several different security options on gotrue then it might be better to make this param an array/kv of different security options

PR #130, opened 2021-08-30
passthrough user metadata on sign up

danalloway:

Lets you add metadata when signing up a user instead of having to sign up then update.

Adding arbitrary data to the auth.user table is only possible via the update call.

## What is the new behavior?

You can optionally pass in user_metadata in the options object to add the metadata upon signup.

## Additional context

This PR expands on the proof of concept provided by @johncomposed in #91

### DIFF
--- a/src/GoTrueApi.ts
+++ b/src/GoTrueApi.ts
@@ -32,6 +32,7 @@ export default class GoTrueApi {
    * @param email The email address of the user.
    * @param password The password of the user.
    * @param redirectTo A URL or mobile address to send the user to after they are confirmed.
+   * @param data Optional user metadata.
    *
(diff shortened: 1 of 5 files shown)


danalloway:
@kiwicopple @johncomposed

made a new PR for this, updates support passing user metadata in both email and phone sign up methods now, tests included

kiwicopple (approved):
sorry for the delay, this is awesome!

if you send this PR to swag@supabase then we can sort you out with some gear

Inzeppelin:
@kiwicopple is it possible to make a new tag with this feature in order to update dependency in [supabase-js](https://github.com/supabase/supabase-js)? I can manage supabase-js dependency bump if we could have a tag here.

PR #137, opened 2021-09-05
Fixed the declared error type for API responses

Andarist:

Type fix

The declared type is misleading and requires typecasting to use the `error.status`

## What is the new behavior?

The DX got improved as the declared type is more accurate

## Additional context

This is what gets thrown by all the API methods:
https://github.com/supabase/gotrue-js/blob/7f32c7a12b792a4e135e059bd06469a5a9e73592/src/lib/fetch.ts#L20-L23

There is still a small chance of an actual `Error` being thrown here, for example, if `_getRequestParams` throws:
https://github.com/supabase/gotrue-js/blob/7f32c7a12b792a4e135e059bd06469a5a9e73592/src/lib/fetch.ts#L47

However, if the implementation is correct and the user uses the API in an intended way (by providing correct information, of the correct types) then this shouldn't really happen. So I find it better to be pragmatic here and assume that no `Error` can happen here and it all leads to `ApiError`. If you think otherwise I can always change all those adjusted declarations to use `ApiError | Error`

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -1,4 +1,4 @@
-import GoTrueApi from './GoTrueApi'
+import GoTrueApi, { ApiError } from './GoTrueApi'
 import { isBrowser, getParameterByName, uuid } from './lib/helpers'
 import { GOTRUE_URL, DEFAULT_HEADERS, STORAGE_KEY } from './lib/constants'
 import {
@@ -114,7 +114,7 @@ export default class GoTrueClient {
   ): Promise<{
     user: User | null
     session: Session | null
-    error: Error | null
+    error: ApiError | null
     data: Session | User | null // Deprecated
   }> {
     try {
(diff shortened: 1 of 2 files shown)


src/GoTrueClient.ts
```diff
@@ -352,14 +352,8 @@ export default class GoTrueClient {
       if (error) {
         return { session: null, error: error }
       }
-      if (!data) {
-        return {
-          session: null,
-          error: { name: 'Invalid refresh_token', message: 'JWT token provided is Invalid' },
-        }
-      }
```
Andarist, line 360: This turned out to be incompatible with the new signature. However, I don't quite see how this line would even be reached. If the JWT token is invalid then an error should be returned so this function would exit within the `if (error)` branch.

awalias: @Iamshankhadeep any input on removing this error on `!data` ?

Iamshankhadeep: @awalias we can remove `!data` as @Andarist said if the token is invalid then this if condition can never be reached. So I am good with omitting `if` condition

awalias (approved):

PR #138, opened 2021-09-09
Fixed how errors from `getSessionFromUrl` are caught in the constructor of the client

Andarist:


src/GoTrueClient.ts
```diff
@@ -89,12 +89,12 @@ export default class GoTrueClient {
...
 
     // Handle the OAuth redirect
-    try {
-      if (settings.detectSessionInUrl && isBrowser() && !!getParameterByName('access_token')) {
-        this.getSessionFromUrl({ storeSession: true })
-      }
-    } catch (error) {
-      console.log('Error getting session from URL.')
+    if (settings.detectSessionInUrl && isBrowser() && !!getParameterByName('access_token')) {
+      this.getSessionFromUrl({ storeSession: true }).then(({ error }) => {
```
Andarist, line 93: `getSessionFromUrl` is an async function so it couldn't quite throw synchronously for this try/catch here to have any effect

I find this error handling here problematic though because, currently, there is no way to get notified about those errors. 

The `getSessionFromUrl` is only async so it can call `this.api.getUser` - which also seems to be non-ideal because it bounds access to the session and the JWT with that endpoint. If my application doesn't care that much about user data but it still needs JWT for other requests then I can easily~ get stuck if this request fails. There is also no easy way to retry this request because I won't be notified anyhow by the failure.

Andarist: While calling `getUser` automatically sounds like something that improves DX it actually makes building resilient applications harder because it hides this fact from the developer and puts him out of the control over timing, errors, and whatnot.

awalias: agree, we should think about adding some events here that devs can override/hook into

awalias (approved):

awalias:
@Andarist thanks a ton for the fix! 🥳  A note for future, we use semantic commit messages so that changes are automatically rolled up into commits. So prepending your commit message with `fix:` will force a new patch version

src/GoTrueClient.ts
Andarist: well, imho it would be better to avoid this `getUser` here altogether and allow for synchronous access to the token after calling `createClient` but, of course, that's your call - error events/hooks would still be a good idea though since those could still happen

PR #147, opened 2021-10-16
Add initial event to onAuthStateChange

lsrugo:

Fixes supabase/gotrue-js#326

There is no event fired unless a sign in occurs, but that is sometimes delayed. So there is no way to know if the user is signed in or not right when the page loads

## What is the new behavior?

There is an `INITIAL` event fired after the library is finished loading and checking for a logged in user.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -86,17 +86,27 @@ export default class GoTrueClient {
       headers: settings.headers,
       cookieOptions: settings.cookieOptions,
     })
+
+    // an array of promises which load the current user
+    let initPromises: Array<Promise<any>> = []
+
     this._recoverSession()
-    this._recoverAndRefresh()
+    const recoverAndRefresh = this._recoverAndRefresh()
+    initPromises.push(recoverAndRefresh)
 
     // Handle the OAuth redirect
     if (settings.detectSessionInUrl && isBrowser() && !!getParameterByName('access_token')) {
-      this.getSessionFromUrl({ storeSession: true }).then(({ error }) => {
+      const getSessionFromUrl = this.getSessionFromUrl({ storeSession: true }).then(({ error }) => {
         if (error) {
           console.error('Error getting session from URL.', error)
         }
       })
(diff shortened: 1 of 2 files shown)


kiwicopple:
looks like the failing tests are from gotrue - not your changes. Let me investigate

lsrugo:
> Adding a bit more food for thought on the above:
> Calling `.session()` or `.user()` will return `null` if the session is expired, though it can be refreshed. So checking these when attaching the auth state listener won't work.

Maybe inside `.session()` or `.user()` we should check if the loading is finished and wait for that before returning `null`

mryechkin:
Has this change been given more consideration? Would love to see it implemented

src/GoTrueClient.ts
```diff
@@ -86,17 +86,27 @@ export default class GoTrueClient {
       headers: settings.headers,
       cookieOptions: settings.cookieOptions,
     })
+
+    // an array of promises which load the current user
+    let initPromises: Array<Promise<any>> = []
```
prescience-data, line 91: Could all this be simplified to 

```typescript
Promise.all([
  this._recoverSession(),
  this._recoverAndRefresh(),
  this.getSessionFromUrl({ storeSession: true }).then(({ error }) => {
    if (error) {
      console.error("Error getting session from URL.", error)
    }
  })
])
  .then(() => {
    this._notifyAllSubscribers("INITIALIZED") // Other events appear to be past-tense.
  })
  .catch((error) => {
    console.error(error)
    // Any other error handling...
  })
```

prescience-data: @kiwicopple @lsrugo 🔎

prescience-data:
https://github.com/supabase/gotrue-js/pull/147#issuecomment-946548042 @gustavohenke 
Could this be solved potentially but setting an `isInitialized` boolean on the client which can be checked at the same time as attaching listeners? 
Still not sure if this would be 100% atomic though (ie vs passing in a callback to the constructor).

hf:
Thank you for this contribution, however the library has advanced quite a bit.

PR #156, opened 2021-11-04
fix: use `location.assign` instead of `href`

### LINKED ISSUE #155: Redirect using `location.assign()` instead of `location.href`
# Feature request

During testing, it's practically impossible to stub out `location.href` (see [this](https://stackoverflow.com/questions/43568570/how-can-i-get-window-location-pathname-on-my-test-file-using-jest/56651910#56651910) and [this](https://stackoverflow.com/questions/36678302/stubbing-window-location-href-with-sinon) among other resources out there) due to--what I'm guessing--are valid security concerns from browser developers.

I would like to be able to stub out Supabase's OAuth2 redirects so I can ensure that they're happening while not actually visiting the 3rd party website during integration tests.

To do this, Supabase should use [`location.assign()`](https://developer.mozilla.org/en-US/docs/Web/API/Location/assign) instead of `location.href` (it does exactly the same thing and has the same browser support).

No idea. I'm at a loss here. I can't figure out how to stub `location.href` so this is the best (and only) solution.

nicholaschiang:

Feature (and a bug fix kinda; fixes #155).

See #155.

## What is the new behavior?

Nothing different; Supabase simply uses `location.assign(url)` instead of `location.href = url`.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -532,7 +532,7 @@ export default class GoTrueClient {
     try {
       // try to open on the browser
       if (isBrowser()) {
-        window.location.href = url
+        window.location.assign(url)
       }
       return { provider, url, data: null, session: null, user: null, error: null }
     } catch (error) {


nicholaschiang:
@awalias @kiwicopple

nicholaschiang:
Actually JK, `window.location.assign` is a non-configurable property as well. 😞

PR #166, opened 2021-11-04
Feat: create user

kiwicopple:
- closes: https://github.com/supabase/gotrue-js/pull/115

- blocker: https://github.com/supabase/gotrue/issues/261

### DIFF
--- a/src/GoTrueApi.ts
+++ b/src/GoTrueApi.ts
@@ -33,6 +33,49 @@ export default class GoTrueApi {
   }
 
   /**
+   * Creates a new user.
+   *
+   * This function should only be called on a server. Never expose your `service_role` key in the browser.
+   *
+   * @param attributes The data you want to create the user with.
+   * @param jwt A valid JWT. Must be a full-access API key (e.g. service_role key).
+   */
+  async createUser(
+    attributes: UserAttributes,
+    jwt: string
+  ): Promise<{ data: null; error: ApiError } | { data: User; error: null }> {
+    try {
+      const data: any = await post(`${this.url}/admin/users`, attributes, {
+        headers: this._createRequestHeaders(jwt),
+      })
+      return { data, error: null }
+    } catch (e) {
+      return { data: null, error: e as ApiError }
+    }
+  }
+
+  /**
+   * Get a list of users.
+   *
+   * This function should only be called on a server. Never expose your `service_role` key in the browser.
+   *
+   * @param attributes The data you want to create the user with.
+   * @param jwt A valid JWT. Must be a full-access API key (e.g. service_role key).
+   */
+  async listUsers(
+    jwt: string
+  ): Promise<{ data: null; error: ApiError } | { data: User[]; error: null }> {
+    try {
+      const data: any = await get(`${this.url}/admin/users`, {
+        headers: this._createRequestHeaders(jwt),
+      })
+      return { data, error: null }
+    } catch (e) {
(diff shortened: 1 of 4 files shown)


kiwicopple:
image

test/clientWithSignupsDisabled.test.ts
```diff
@@ -21,3 +28,97 @@ test('signUp()', async () => {
...
+      aud: expect.any(String),
+      updated_at: expect.any(String),
+      invited_at: expect.any(String),
+      app_metadata: {
+        provider: 'email',
+      },
+    },
+    `
+    Object {
+      "action_link": "",
```
kiwicopple, line 52: @kangmingtay - it seems like the invite link isn't generated if signups are disabled?

test/clientWithSignupsDisabled.test.ts
```diff
@@ -21,3 +28,105 @@ test('signUp()', async () => {
   expect(data).toBeNull()
   expect(user).toBeNull()
 })
+
+test('generateLink() should be able to generate multiple links', async () => {
+  const invitedUser = faker.internet.email().toLowerCase()
+  const { error, data: firstInvite } = await authAdmin.generateLink('invite', invitedUser, {
+    redirectTo: 'http://localhost:9997',
+  })
```
kiwicopple, line 36:
```
const { error, data: firstInvite } = await authAdmin
  .generateLink('invite', 
  'email@example.com', 
  {
    redirectTo: 'http://localhost:9997',
  })
```

If there are no SMTP settings on GoTrue, this will not return an action_link (we are investigating)

src/GoTrueApi.ts
```diff
@@ -32,6 +32,46 @@ export default class GoTrueApi {
...
+   * Get a list of users.
+   *
+   * This function should only be called on a server. Never expose your `service_role` key in the browser.
+   *
+   * @param attributes The data you want to create the user with.
+   * @param jwt A valid JWT. Must be a full-access API key (e.g. service_role key).
+   */
+  async listUsers(): Promise<{ data: null; error: ApiError } | { data: User[]; error: null }> {
+    try {
+      const data: any = await get(`${this.url}/admin/users`, {
```
bnjmnt4n, line 66: Thanks for adding this to the API! I think accepting a filter parameter in the `listUsers` method is also going to be very helpful. (Related usecase: https://github.com/supabase/gotrue/issues/180).

src/GoTrueApi.ts
```diff
@@ -32,6 +32,46 @@ export default class GoTrueApi {
...
+
+  /**
+   * Get a list of users.
+   *
+   * This function should only be called on a server. Never expose your `service_role` key in the browser.
+   *
+   * @param attributes The data you want to create the user with.
+   * @param jwt A valid JWT. Must be a full-access API key (e.g. service_role key).
+   */
+  async listUsers(): Promise<{ data: null; error: ApiError } | { data: User[]; error: null }> {
```
bnjmnt4n, line 64: The response data is actually of the format `{ aud: "authenticated", users: User[] }`.

Ref: https://github.com/supabase/gotrue/blob/401d4cf375e941f3db0fff84319e56a2bdec8d26/api/admin.go#L80-L83.

src/GoTrueApi.ts
kiwicopple: I'll merge this for now and then we can add this in the future 👍

kangmingtay (approved):

PR #193, opened 2021-12-04
Typo in Readme

AlainGourves:

Readme update

Documentation refers to Firebase.


Isaiah-Hamilton (commented):
This is supposed to be Firebase because Supabase is the open source alternative to Firebase. Which means that we are building the features of Firebase using enterprise-grade, open source products.

AlainGourves:
My bad ! I had misread the phrase...

Issue #191, opened 2021-12-03
`supabase.auth.api.createUser(attributes: UserAttributes)`. Current set of params in UserAttributes do not match the admin user params in gotrue


kangmingtay:
* Include params to indicate whether the user created is confirmed or not 
* Include param to modify AppData

See the request body params accepted by gotrue [here](https://github.com/supabase/gotrue/blob/master/api/admin.go#L15-L25)

Issue #190, opened 2021-12-03
Update gotrue-js admin functions


kangmingtay:
# Feature request
* Addresses #90, #187
* Have a proper 1-1 mapping of methods in gotrue-js to all endpoints in gotrue

# Methods to be added
- [x] `getUser(user_id)` method in `GoTrueApi` that maps to `GET /admin/users/<user_id>` 
   - `getUser(user_id)` should also extend the pagination & filter ability in `GET /admin/users/<user_id>` by passing them as query params in the fetch request made 
- [x] `updateUser(user_id)` method in `GoTrueApi` that maps to `PUT /admin/users/<user_id>` 

# Methods to be modified
- [x] #192
- [x] #191

** Note: Any gotrue-js methods that map to a gotrue `/admin/[...]` endpoint should only be used server-side:
```js
const supabaseAdminClient = createClient(supabase_url, service_role_key)
async function main() {
  const attributes = {...}
  const { data: user, error: createErr } = await supabaseAdminClient.auth.api.createUser(attributes)
  const { data, error: deleteErr } = await supabaseAdminClient.auth.api.deleteUser(user.id)
}
```

Issue #187, opened 2021-12-01
Add the listUser admin endpoint 


kangmingtay:
# Bug report
* Currently, the `supabase.auth.api.getUser('access_token_jwt')` maps to the `GET /user` endpoint in gotrue. However, there is not corresponding method for the `GET /admin/user/<user_id>` endpoint.

# Solution
* We should create a method `supabase.auth.api.listUser()` that requires the `service_role` key. This method should only be used on the server-side since it requires the `service_role` key.
* To prevent confusion between `supabase.auth.api.getUser('access_token_jwt')` and `supabase.auth.api.listUser()`, we can probably move the former to `supabase.auth.getUser('access_token_jwt')` to indicate that this method is safe to call on the client-side. 
* This also makes it consistent with the rest of the server-side methods (i.e. `supabase.auth.api.[...]`) that require the `service_role` key.

kangmingtay:
cc @thorwebdev @awalias

thorwebdev:
Sounds good to me. Seeing the we have `getUserByCookie` should we rather call it `getUserById`? Or would `listUser` also allow me to list all users when I don't provide an ID?

dthyresson:
> Sounds good to me. Seeing the we have `getUserByCookie` should we rather call it `getUserById`? Or would `listUser` also allow me to list all users when I don't provide an ID?

@thorwebdev i prefer `getUserById`   — any method with “list” to me implies an array response with multiple items. 

This opens up other find by’s like by email or by phone number.

Another option following Prisma’s naming info is `findUnique` where options would be if or email or phone; any unique key.

That said, I think I could see a need to list users — say to send everyone an email — but I think I would do that using views with filter conditions or a function to iterate over each.

bnjmnt4n:
FYI there is already an `auth.api.listUsers()` method: https://github.com/supabase/gotrue-js/blob/de87da1f739920e10313c169ef3b4faba459d9c7/src/GoTrueApi.ts#L56-L70

PR #195, opened 2021-12-15
feature: Implement Admin API User functions and Refactor Test Suite

dthyresson:

The PR is twofold:

Fixes https://github.com/supabase/gotrue-js/issues/190
Fixes https://github.com/supabase/gotrue-js/issues/191
Fixes https://github.com/supabase/gotrue-js/issues/187

1 - Implements Issue #190 to allow admin api to 
* get the user by an id
* update a user by id (with both user and app metadata)
* delete a user (but now using the service not api client, not the auth client when a service role JWT)

2 - The test suite 
a) lacked coverage 
b) was confusing to see coverage because was scenario driven (sign up enabled/disabled) and mixed auth and service api methods in the tests) 
c) did not work well with test tools like [Wallaby[(https://wallabyjs.com) as the test had to be run sequentially due to the way emails and password were generated to create test users

## What is the new behavior?

1. Implements the needed admin functions
2. Centralizes client and client config to promote reuse
3. Utility functions mock credentials (email, password, and phone) to reuse  
4. Refactors suite to tell more "stories" and work with Wallaby testing

### GoTrueApi
Screen Shot 2021-12-14 at 6 57 39 PM

### GoTrueClient

Screen Shot 2021-12-14 at 6 54 57 PM

### Test Results

Screen Shot 2021-12-14 at 6 55 40 PM

## Additional context

Note: May still need some more enabled/disabled scenario tests and also have one disabled verifyOTP() test


dthyresson:
I got a message that a secret was exposed here in my clients as `GOTRUE_JWT_SECRET`

image

But this same value is in the docket compose needed for infra local deployment, so I am not that concerned.

dthyresson:
Once the PR is approved will need to update the admin docs for:

* getUserById
* updateUserById
* deleteUser

dthyresson:
Note: the `listUsers` command perhaps should be renamed `findUsers` since I just learned that the GoTrue implementatiion supports a find filter using `FindUsersInAudience` that finds users with the matching audience with a filter option and a sort:

```
	if filter != "" {
		lf := "%" + filter + "%"
		// we must specify the collation in order to get case insensitive search for the JSON column
		q = q.Where("(email LIKE ? OR raw_user_meta_data->>'full_name' ILIKE ?)", lf, lf)
	}

	if sortParams != nil && len(sortParams.Fields) > 0 {
		for _, field := range sortParams.Fields {
			q = q.Order(field.Name + " " + string(field.Dir))
		}
	}
```

But, the only filter currently is `email`.

I'll make a separate issue for adding filter support to `listUsers` or `findUsers`.

thorwebdev (approved):
LGTM 👍

Regarding https://github.com/supabase/gotrue-js/pull/195#issuecomment-994697802 I think it's fine to keep it named listUsers and add the filters in there e.g. `listUsers({ email, limit })` etc.

src/lib/types.ts
```diff
@@ -92,13 +92,23 @@ export interface UserAttributes {
...
   data?: object
 }
 
+export interface AdminUserAttributes extends UserAttributes {
+  /**
+   * A custom data object for app_metadata that only a service role can modify.
+   *
+   * Is JSON that includes app-specific info, such as identity providers, roles, and other
+   * access control information.
+   */
```
kangmingtay, line 107: should we also add the boolean fields `emailConfirm` and `phoneConfirm` here? 

see https://github.com/supabase/gotrue/blob/master/api/admin.go#L21-L24

test/lib/clients.ts
```diff
@@ -0,0 +1,78 @@
+import jwt from 'jsonwebtoken'
+import { GoTrueApi, GoTrueClient } from '../../src/index'
+
+export const AUTO_CONFIRM_ENABLED_GOTRUE_PORT = 9999
+export const SIGNUP_ENABLED_AUTO_CONFIRM_ENABLED_PORT = 9998
+export const AUTO_CONFIRM_DISABLED_GOTRUE_PORT = 9997
+
+export const GOTRUE_URL_AUTO_CONFIRM_ENABLED = `http://localhost:${AUTO_CONFIRM_ENABLED_GOTRUE_PORT}`
+export const GOTRUE_URL_SIGNUP_ENABLED_AUTO_CONFIRM_ENABLED = `http://localhost:${SIGNUP_ENABLED_AUTO_CONFIRM_ENABLED_PORT}`
+export const GOTRUE_URL_AUTO_CONFIRM_DISABLED = `http://localhost:${AUTO_CONFIRM_DISABLED_GOTRUE_PORT}`
```
kangmingtay, line 10: would it be better to throw all of these into a `.test.env` file?

src/GoTrueApi.ts
```diff
@@ -326,166 +352,147 @@ export default class GoTrueApi {
...
-   * @param uid The user uid you want to remove.
+   * @param attributes The data you want to create the user with.
    * @param jwt A valid JWT. Must be a full-access API key (e.g. service_role key).
    */
-  async deleteUser(
-    uid: string,
-    jwt: string
-  ): Promise<{ user: User | null; data: User | null; error: ApiError | null }> {
+  async createUser(
+    attributes: UserAttributes
```
kangmingtay, line 450: should this be `AdminUserAttributes` instead?

src/GoTrueApi.ts
```diff
@@ -326,166 +352,147 @@ export default class GoTrueApi {
...
-        }
-      )
-      return { user: data, data, error: null }
+      const data: any = await post(this.fetch, `${this.url}/admin/users`, attributes, {
+        headers: this.headers,
+      })
+      return { data, error: null }
     } catch (e) {
-      return { user: null, data: null, error: e as ApiError }
+      return { data: null, error: e as ApiError }
```
kangmingtay, line 458: will removing the user key (even though it returns null) result in a breaking change?

kangmingtay (changes requested):

src/lib/types.ts
dthyresson: Ah, yes, we should so that the types match up with GoTrue's:

```
type adminUserParams struct {
	Aud          string                 `json:"aud"`
	Role         string                 `json:"role"`
	Email        string                 `json:"email"`
	Phone        string                 `json:"phone"`
	Password     string                 `json:"password"`
	EmailConfirm bool                   `json:"email_confirm"`
	PhoneConfirm bool                   `json:"phone_confirm"`
	UserMetaData map[string]interface{} `json:"user_metadata"`
	AppMetaData  map[string]interface{} `json:"app_metadata"`
}
```

test/lib/clients.ts
dthyresson: The test suite and package doesn't rely on dotenv or have any envier config.

I'd still have to check in a .env file with these settings and then inline them like

```
export const AUTO_CONFIRM_ENABLED_GOTRUE_PORT = process.env. AUTO_CONFIRM_ENABLED_GOTRUE_PORT
```

and they *must* match the docker compose settings, which are not enviers.

So, therefore, keeping values in the client config here.

dthyresson:
@kangmingtay Made some mods and tried to answer your questions and hope resolved so can be merged. Thanks!

kangmingtay (approved):

PR #198, opened 2021-12-20
Changes setAuthCookie to also set cookie on TOKEN_REFRESHED event

alaister:

Currently, `setAuthCookie` only sets the cookie on the `SIGNED_IN` event. This PR is a simple change that makes `setAuthCookie` also set the cookie on the `TOKEN_REFRESHED` event, ensuring the cookie auth state always reflects the browser auth state.


dthyresson:
Hi @alaister and thanks for the PR -- and resolving the conflicts after my Pr to refactor the client and the test suit was merged in.

In an effort to increase the test coverage, one API function on the api client I had not yet written tests for was `setAuthCookie` but I was wondering if you might take that on?

It's hard to test cookies since one doesn't have access to the window -- but I think here you may be able to set the request and check the headers ... or mock the `setCookie` method and then just check that you expect it `toHaveBeenCalled` once when the request event says it should.

I'm happy to write the test tomorrow unless you'd like to, but I think it would be excellent if the test suite could have a unit test for `setAuthCookie`.

Thoughts?

alaister:
I'm working today but happy to take a look at writing a unit test tomorrow - will be good to see how you've got testing set up :)

alaister:
@dthyresson found some time and added a few tests. My mock requests and expect conditions are pretty basic but seem to do the trick.

Some possible improvements might be to use an actual `TOKEN_REFRESHED` event for the token refreshed test, but I'm not sure of the best way to do this, and to improve the expect conditions to look for the actual tokens / lack of tokens.

Are these tests sufficient? Please let me know what you think.

dthyresson:
> Are these tests sufficient? Please let me know what you think.

Thanks for writing these tests @alaister -- I'll review and try out/comment today.

As for the functionality to set on `TOKEN_REFRESHED`, I'll have to defer to @kangmingtay or @thorwebdev to determine if this event is appropriate (though, seems to be the case).

I know many people are taking time off this week, so it may be a little while before the reviews come in.

But, again thanks much for the PR and hope to get it merged as soon as can.

dthyresson:
> Are these tests sufficient? Please let me know what you think.

Thanks @alaister I pulled in your PR and ran the test and looks great!

Would you mind adding the following scenarios so we can cover the error cases?

* an unknown event such as

```ts
      const event = 'UNKNOWN_EVENT'

      const mockRequest = mockCookieRequest({ event, session })
      const mockResponse = mockCookieResponse()

      authClientWithSession.api.setAuthCookie(mockRequest, mockResponse)
```

* A test when method is not a POST to cover:

```
    if (req.method !== 'POST') {
```

Maybe pass in method but default post to `mockCookieRequest`?

* case when no event provided to cover

```
    if (!event) throw new Error('Auth event missing!')
```

and last

* when not signed in (aka no session) to cover

```
      if (!session) throw new Error('Auth session missing!')
```

And then text that they expect to throw the error.

Much appreciated!

alaister:
Thanks for the feedback - great suggestions! I have added the tests.

dthyresson:
> Thanks for the feedback - great suggestions! I have added the tests.

Great stuff! As said, want to wait for Kang Ming or Thor to have a look post-holidays but then they should get to approve and merge.

Cheers.

thorwebdev (approved):
As far as I can see `TOKEN_REFRESHED` event is always followed by a `SIGNED_IN` event: https://github.com/supabase/gotrue-js/blob/master/src/GoTrueClient.ts#L627-L629

So I guess it would be executed twice in quick succession, which is probably fine though. Thanks a mill for adding the tests! 💚

alaister:
How strange – the only reason I made this change in the first place was my session cookie wasn't getting refreshed after a client-side refresh. I think you're right, it will be fine with double setting of the cookie if it isn't the case though.

Issue #201, opened 2021-12-29
Refresh Token revocation span sessions leading to un-determined authentication states


dthyresson:
# Bug report

First, my understanding about gotrue-js and how refresh tokens work is:

* On login, when an access token is granted a refresh token is generated and provided to the client
* This refresh token is assigned to the user and stored in the `auth.refresh_tokens` table with a `revoked` status (true/false) and some timestamps
* The access token is valid until it expires and that expiration duration interval is set in auth settings, defaults to 3600 seconds (1 hour)
* On login, the client starts a timer to auto refresh the token (there is an `authRefreshToken` config setting on the client and Supabase defaults this to true).
* About 60 seconds before the token expires, the refresh token handler is invoked and a new access token is generated
* The old refresh token is revoked (its status in the table is changed) and a new token is added with a reference to the old (parent)
* Happy path is that the token **can be refreshed** and a new access token is created with an updated expiration such that any subsequent server-side verification of the JWT will consider it valid (unexpired)

But -- what happens if the tokens **cannot** be refreshed ... and how might someone get themselves into this state?

## Scenario 1 - Same Login User, Two Browsers and Multiple "Sessions"

One way this can happen is to login using two browsers: one Chrome and one Safari, then logout from within Chrome and see what happens to the Safari sessions.

### Steps

This assumes you already have a valid signed up Supabase user and know the credentials.

Also, JWT expiration is set to 120 seconds (two minutes for easy testing).

These use RedwoodJS and the latest Supabase gotrue-js in the Auth Playground local dev testing..

For this we'll user User `9cb11b1f-6a3d-4c94-9280-41af55ed1dc9`

![image](https://user-images.githubusercontent.com/1051633/147617758-dac43f35-bce9-4b3c-996a-7a94b645de2d.png)

1. Launch Chrome and login

![image](https://user-images.githubusercontent.com/1051633/147617807-cd6f708a-a7b4-4abc-9d7a-0fa992c426ed.png)

Every 20 seconds of so we make a query using the JWT and can see it return successfully

![image](https://user-images.githubusercontent.com/1051633/147617856-4d62a17b-d10b-413d-96cf-22cfa534d091.png)

After waiting a few minutes you can see we have several refresh tokens for this login and user (some revoked and some active)

![image](https://user-images.githubusercontent.com/1051633/147617913-c4ef5b9a-c2ca-449a-8044-b6aa43ea781a.png)

2. Launch Safari
3. Login with same user in two tabs

[...]

PR #203, opened 2021-12-31
fix: resolve race conditions when refreshing access token

wattroll:

This fixes a race condition that arises when there are overlapping calls to `GoTrueApi.refreshAccessToken`. It does so by keeping reference to the result of `this._callRefreshToken` until it is settled and returning it to the concurrent calls.

Concurrent token refreshes inevitably result in unrecoverable failure:

- the earliest call will succeed, revoke current session and receive a new one;
- the later calls will fail due to using revoked session AND revoke the new session due to [token reuse detection](https://github.com/supabase/gotrue/blob/fb36ef15c7bde5e17737a5111a95d28a30e52bba/api/token.go#L261).
- client will be left with a valid `access_token` and revoked `refresh_token`
- client might not notice the problem until the next time when they try to refresh the session

Currently library code doesn't trigger such conditions unless tokens are short lived and network latency is high. User code will trigger it if it calls `refreshSession()` naively (e.g. when handling a 401 "token expired" error from several parallel requests).

Currently to avoid this problem user code must implement similar kind of batching that I did here. But even then there is no clean way to avoid racing against the auto refresh timer apart from disabling it (which also disables the token recovery from local storage).

## What is the new behavior?

If `this._callRefreshToken` is invoked while an earlier request is in-flight they will be settled with the same promise. Once the promise is settled the reference to it is deleted, so that future calls will initiate fresh requests. 

This fix is intended to be fully compatible with existing code (including potential access to private properties) and allows user code to handle expired tokens more reliably.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -615,11 +615,32 @@ export default class GoTrueClient {
     }
   }
 
+  private _callRefreshTokenPromises: {
+    [refresh_token: string]: ReturnType<GoTrueClient['_callRefreshTokenConcurrently']>
+  } = {}
+
+  // If there is an ongoing request a promise resolving to it will be returned.
+  //
+  // This is necessary because concurrent token refreshes will race and inevitably fail:
+  //    - earliest request will revoke current session and start a new one
+  //    - later requests will fail as they were sent with an old session
+  //    - this will cause new session to also be revoked due to refresh token reuse detection
+  //    - the client will be left with a valid access_token but revoked refresh_token
+  //    - they might not notice it until they try to refresh the token
   private async _callRefreshToken(refresh_token = this.currentSession?.refresh_token) {
+    if (!refresh_token) {
+      return { data: null, error: { message: 'No current session', status: 401 } as ApiError }
+    }
+    if (!(refresh_token in this._callRefreshTokenPromises)) {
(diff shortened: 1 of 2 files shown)


dthyresson:
Hi @wattroll and thanks for this PR.

I recently opened https://github.com/supabase/gotrue-js/issues/201 which described a few scenarios where you could get yourself into a situation where the the token would no longer refresh.

And I think we are onto the same thing here:

```
Concurrent token refreshes inevitably result in unrecoverable failure:

* the earliest call will succeed, revoke current session and receive a new one;
* the later calls will fail due to using revoked session AND revoke the new session due to token reuse detection.
* client will be left with a valid access_token and revoked refresh_token
* client might not notice the problem until the next time when they try to refresh the session
```

As that's what I say too -- but I was curious if you had found a way to get your client into that state without using a second browser or logging out as I had.

Also, I see that you are now returning a 401 error status vs simply an exception -- which I like since as you noted, it eats that error and doesn't inform the client such that tokens will continue to fail "silently".

I was wonder if you thought it would be helpful to bubble up a "TOKEN_REVOKED" event that the client could subscribe to -- just like you can subscribe to a TOKEN_REFRESHED or SIGNED_OUT event?

wattroll:
This has been resolved by #285 and #339.

Issue #169, opened 2021-11-07
Support id_token grant flow


dshukertjr:
# Feature request

## Is your feature request related to a problem? Please describe.

A new way of signing users in via OAuth was introduced with [this](https://github.com/supabase/gotrue/pull/189) PR of gotrue. It would be nice if we could add a way of utilizing this endpoint with this library so that native apps built with react native could utilize it. 

## Describe the solution you'd like

OAuth sign in flow on native apps should use `POST /token?grant_type=id_token` end point to sign in. 

## Describe alternatives you've considered

Currently OAuth sign in is possible by redirecting the user to web view from a native app. 

## Additional context

N/A

megacherry:
@dshukertjr Hi Tyler, do you know when this feature will be integrated? or do you have a rough timeline? thanks in advance, Andy

dshukertjr:
@megacherry Great question! 

@kiwicopple Now that the backend is ready for this, do we have anything else that needs to be in place before we add this feature to the client?

phamhieu:
cc @kangmingtay

megacherry:
@dshukertjr Do you have any new information about this issue? :)

koba-ninkigumi:
@kiwicopple @kangmingtay @phamhieu @dshukertjr @megacherry

I have created a pull request that resolves this issue.
Please review and merge.

DanMossa:
It looks like your pr was merged in ? Should this be closed now?
@koba-ninkigumi

hf:
Closing as this is already supported.

PR #207, opened 2022-01-14
api implementation for OpenID Connect users

koba-ninkigumi:

This is the api implementation for OpenID Connect users.

```
const { user, session, error } = await supabase.auth.signIn({
  oidc:{
      id_token: 'your idtoken',
      nonce: 'random value',
      provider: 'google'
  }
})
```

This pull request solves the following issue.

- https://github.com/supabase/gotrue-js/issues/169
- https://github.com/supabase/gotrue/issues/341
- https://github.com/supabase/gotrue/issues/171
- https://github.com/supabase/gotrue/issues/146

This pull request is made possible by the following pull request.

- https://github.com/supabase/gotrue/pull/189

none.

## What is the new behavior?

Numerous issues will be resolved.

## Additional context

https://github.com/supabase/supabase/pull/5181

You should check the OpenID Connect specification for details. https://openid.net/developers/specs/

Sign In With Google
https://developers.google.com/identity/gsi/web/guides/overview

Handle credential responses with JavaScript functions
https://developers.google.com/identity/gsi/web/guides/handle-credential-responses-js-functions

↑ You can find instructions on how to get the token here. (It is included in the response)

Sign In With Google HTML API reference
https://developers.google.com/identity/gsi/web/reference/html-reference

Sign In With Google JavaScript API reference
https://developers.google.com/identity/gsi/web/reference/js-reference

↑ Here is the description of the nonce that @dthyresson was requesting.

This is sample code.

```
<script src="https://cdn.jsdelivr.net/npm/@supabase/supabase-js/dist/umd/supabase.min.js"></script>
<script src="https://accounts.google.com/gsi/client" async defer></script>
<div id="g_id_onload"
     data-client_id="YOUR_GOOGLE_CLIENT_ID"
     data-context="signin"
     data-ux_mode="popup"
     data-callback="handleCredentialResponse"
     data-nonce="random value"
     data-auto_prompt="false">
</div>
<div class="g_id_signin"
     data-type="standard"
     data-shape="rectangular"
     data-theme="outline"
     data-text="$ {button.text}"
     data-size="large"
     data-logo_alignment="left">
</div>
<script>
  function handleCredentialResponse(response) {

    console.log(response)

    ;(async ()=>{
      const { user, session, error } = await supabase.auth.signIn({
        oidc:{
            id_token: response.credential,
            nonce: 'random value',
            provider: 'google'
        }
      })
    })()
  }
</script>
```


dthyresson:
Hi @koba-ninkigumi and thanks much for this PR.

Might you also be able to write up unit tests for a few scenarios?

* success
* failures


Given this flow :




and the GoTrue rules in https://github.com/supabase/gotrue/blob/8ba49df7c7bbcc87c9d5b211a32cb5aa17db231d/api/token.go#L316

* missing required params like "id_token, nonce and provider required
* what if the nonce does't match 
* missing sub claim in id_token
* scenario when signups are disabled 


This way we can have some tests in place to cover cases developers may encounter when using.

Thanks.

dthyresson (changes requested):
Hi, I left some notes and a request to add several tests.

Thanks again for the PR.

src/GoTrueApi.ts
```diff
@@ -203,6 +203,34 @@ export default class GoTrueApi {
     }
   }
 
+  /**
+   * Logs in an OpenID Connect user using their id_token.
+   * @param id_token The id_token of the user.
+   * @param nonce The nonce of the user.
```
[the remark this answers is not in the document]
koba-ninkigumi: Let's make it clear.

The nonce is not automatically generated by the initial grant as a function of supabase, but is a random value generated by the developer himself "before" the initial grant is started.

There is no way that an app developer can be unaware of a nonce that he or she has generated on the fly.

However, it is expected that developers who do not understand the OpenID Connect specification will ask what a nonce is. To deal with this, it might be a good idea to specify the following points in the supabase change log or documentation.

1) The nonce is a value randomly generated by the developer before the initial grant is started.
2) You should check the OpenID Connect specification for details. https://openid.net/developers/specs/

I personally think that developers who do not understand what nonce is should not use this method.

src/GoTrueApi.ts
```diff
@@ -203,6 +203,34 @@ export default class GoTrueApi {
...
+   * @param provider The provider of the user.
+   */
+  async signInWithIDToken(
```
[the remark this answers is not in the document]
koba-ninkigumi: It will be OpenID Connect, not OpenID.

However, signInwithOpenIDConnect seems to be too long.

Also, OIDC is often used as an abbreviation for OpenID Connect.
If you do a Google search for "OIDC", OpenID Connect is the one that comes up at the top, and there are no other confusing options.

What about signInwithOIDC?

koba-ninkigumi:
https://github.com/supabase/gotrue-js/pull/207#issuecomment-1013158198

- scenario when signups are disabled

Enabling OpenID Connect means that you are accepting a huge group of users from OpenID Connect providers (Google, Facebook, etc.) as already signed up.

That's exactly what "OpenID" means.

The idea of disabling signup on the supabase side is incompatible with OpenID Connect.

koba-ninkigumi:
https://github.com/supabase/gotrue-js/pull/207#issuecomment-1013158198

- missing required params like "id_token, nonce and provider required
- what if the nonce does't match
- missing sub claim in id_token

If you or I know how to pretend to be an OpenID Connect provider and return id_token, it means that the OpenID Connect specification is broken.

And since we can't get the id_token, we can't write a test.

Please see the following test for reference.

https://github.com/supabase/gotrue-js/blob/master/test/GoTrueClient.test.ts#L182

This is probably the same reason why there are no test cases for oauth providers here.

If you think it is possible to write a test for OpenID Connect, could you please clarify how to do so in more detail?

He1nr1chK:
Anything I can help with to get this merged? Would love to use this PR asap @koba-ninkigumi  @dthyresson

dthyresson:
Hi @He1nr1chK thanks for offering to help.

I think that these new functions could use:

* unit tests
* documentation for how to use them (ie, what one would see and need to know how to use in: https://supabase.com/docs/reference/javascript/auth-signin)
* And some examples of how to setup and test in a real world case. What are some services I can use and setup (and how) to know that this method works?

koba-ninkigumi:
> * And some examples of how to setup and test in a real world case. What are some services I can use and setup (and how) to know that this method works?

Sign In With Google
https://developers.google.com/identity/gsi/web/guides/overview

Handle credential responses with JavaScript functions
https://developers.google.com/identity/gsi/web/guides/handle-credential-responses-js-functions

↑ You can find instructions on how to get the token here. (It is included in the response)

Sign In With Google HTML API reference
https://developers.google.com/identity/gsi/web/reference/html-reference

Sign In With Google JavaScript API reference
https://developers.google.com/identity/gsi/web/reference/js-reference

↑ Here is the description of the nonce that @dthyresson was requesting.


This is sample code.



```
<div id="g_id_onload"
     data-client_id="YOUR_GOOGLE_CLIENT_ID"
     data-callback="handleCredentialResponse"
     data-nonce="random value">
</div>
<script>
  function handleCredentialResponse(response) {
    // decodeJwtResponse() is a custom function defined by you
    // to decode the credential response.
    const responsePayload = decodeJwtResponse(response.credential);

    console.log("ID: " + responsePayload.sub);
    console.log('Full Name: ' + responsePayload.name);
    console.log('Given Name: ' + responsePayload.given_name);
    console.log('Family Name: ' + responsePayload.family_name);
    console.log("Image URL: " + responsePayload.picture);
    console.log("Email: " + responsePayload.email);

    const { user, session, error } = await supabase.auth.signInWithOpenIDConnect({
      id_token: response.credential,
      nonce: 'random value',
      provider: 'google'
    })
  }
</script>
```

koba-ninkigumi:
> - unit tests

As I mentioned above, due to the OpenID Connect specification, it should be impossible to get an idtoken that works properly.
So I don't know how to unit test this method.

If someone thinks it is possible, it would be great if that someone could write the unittest of this method.

koba-ninkigumi:
@dthyresson
I wrote a unittest that expects an error because the idtoken is not valid.

dthyresson:
> If someone thinks it is possible, it would be great if that someone could write the unittest of this method.

Could you mock the request and return expected results?

This would ensure that the method signature or other logic doesn't accidentally get changed without some test check.

For example, in a PR the cookie is being mocked to return headers as:

```ts
export const mockCookieRequest = ({
   event,
   session,
   method = 'POST',
 }: {
   event?: string
   session?: Session | null
   method?: string
 }) => {
   return {
     method,
     headers: { host: 'localhost:9999' },
     body: { event, session },
   }
 }

 export const mockCookieResponse = () => {
   return {
     getHeader: jest.fn(() => undefined),
     setHeader: jest.fn(),
     status: jest.fn(() => {
       return { json: jest.fn(), end: jest.fn() }
     }),
   }
 }
```

Maybe something similar?

koba-ninkigumi:
> Maybe something similar?

Instead of using a cookie as an example, use oauth2 from gotrue.js as an example.
This is because the OpenID Connect specification is an extension of oauth2.

Currently, none of the tests for signIn() in gotrue.js specify a provider.
A Mock that returns an access token would be helpful.

koba-ninkigumi:
I don't know of any other way to get a working access token other than getting the official Google private key.

koba-ninkigumi:
@dthyresson @He1nr1chK 

I can't prepare the test that @dthyresson requires. I would like to wait for someone's help.

He1nr1chK:
@dthyresson  I have done some research into the possible tests one can write for OCID and I have to agree with @koba-ninkigumi. It's going to be difficult if not impossible to write a proper test for the function. Is the test absolutely necessary?

awalias:
hey @koba-ninkigumi @He1nr1chK this is awesome! Thanks for your hard work.

>Is the test absolutely necessary

It would be very useful to have some automated tests to make sure we don't accidentally break the logic later on. 

>  I have done some research into the possible tests one can write for OIDC and I have to agree with @koba-ninkigumi. It's going to be difficult if not impossible to write a proper test for the function

Could you expand on what is the difficult bit? Could we mock the request as @dthyresson suggested?

Again, thanks for all your help on this! 🙏

koba-ninkigumi:
@awalias

OpenID Connect is an authentication. It's a specification for proving "who you are."

Who is the test runner?　It is the test runner.

A normal OpenID Connect provider will try to verify that a human actually accessed the site with a browser.
For example, in the case of Google, they ask for confirmation through biometrics or double authentication.

OpenID Connect providers are tasked with rejecting non-human access, such as test runners.

If a test runner is authenticated by OpenID Connect, even if it means a successful test of supabase goture.
It's just a failure on the part of the OpenID Connect provider.

How do you exploit that hole to illegally obtain an idtoken?
And is it a successful test to get the idtoken?
No, it's just a collapse of the system.

koba-ninkigumi:
@awalias

In order to connect to supabase with OpenID Connect, you need to get a legitimate idtoken beforehand.
It's impossible to get that idtoken.

If it were possible to write a test to authenticate the idtoken, it could be done in two ways.

1) Set up a fake OpenID Connect provider in your test and issue idtokens that work correctly there.

2) You are a futurist from 1000 years in the future who is familiar with quantum computers and can crack the cipher used in OpenID Connect in less than an hour to get a fake idtoken and put all that logic in the test.

koba-ninkigumi:
@awalias

My suggestion is to release this feature first.
Then, the testing of this feature should be a separate issue.

koba-ninkigumi:
@awalias 

I would also add that if you are setting up a fake OpenID Connect Provider in your testing, you will need to provide the necessary information on an externally accessible https server, per the OpenID Connect specification.
Otherwise, it will be considered a fraudulent idtoken based on the oidc specification.

koba-ninkigumi:
@awalias 

Can't you still create an idtoken in the mock and use it ? You may be wondering.
If the goture backend of supabase considers such an idtoken as a normal idtoken, it means that the goture backend of supabase is broken.

Clarification.

Writing a test that passes an idtoken is not difficult, it is impossible for me.

koba-ninkigumi:
@awalias 

One more thing to add, if you can find a way to create a mock and get a working idtoken, you can have unlimited wealth in your BitCoin wallet.
This is because getting a working OpenID Connect idtoken is the same as getting a legitimate BitCoin, in the sense that it interferes with the chain of authentication of the cryptographic key.

koba-ninkigumi:
#https://github.com/supabase/gotrue/pull/360
↑In response to this PR, we have added client_id and issuer as parameters to work with any OIDC Provider.

@JO Thanks for your generous support.

@kangmingtay Please provide support from the perspective of an implementor of OIDC for goture backend.

koba-ninkigumi:
https://github.com/supabase/supabase/pull/5181

↑ Changes have been completed to conform to this PR.

He1nr1chK:
Any updates on when this might get merged?

J0 (commented):
LGTM. Let's see if @kangmingtay and @dthyresson have any comments before we click merge though.

dthyresson:
1. There was an idea discussed to make this api an option on `signUp` just like a 3rd-party provider rather than a new method. Is this still an idea? So, the OIDC params are options.
2. I'd very much like to see documentation in a draft PR on https://github.com/supabase/supabase/blob/master/web/spec/supabase.yml to describe use and setup.

But defer to @kangmingtay and @J0 and @awalias and  @thorwebdev.

koba-ninkigumi:
> I'd very much like to see documentation in a draft PR on https://github.com/supabase/supabase/blob/master/web/spec/supabase.yml to describe use and setup.

@dthyresson Please see,
https://github.com/supabase/supabase/pull/5181/files

He1nr1chK:
Is there anything else delaying this merge? @kangmingtay @awalias

src/GoTrueClient.ts
```diff
@@ -240,7 +241,10 @@ export default class GoTrueClient {
           scopes: options.scopes,
         })
       }
-      throw new Error(`You must provide either an email, phone number or a third-party provider.`)
+      if (openIDConnect) {
+        return this._handleOpenIDConnectSignIn(openIDConnect)
+      }
+      throw new Error(`You must provide either an email, phone number or a third-party provider or OpenID Connect.`)
```
dthyresson, line 247:
```suggestion
      throw new Error(`You must provide either an email, phone number, a third-party provider or OpenID Connect.`)
```

Is it clear what ` OpenID Connect` means?

src/GoTrueApi.ts
```diff
@@ -215,6 +216,32 @@ export default class GoTrueApi {
...
+   * @param issuer The issuer of the user.
+  */
+  async signInWithOpenIDConnect({ IDToken, nonce, clientID, issuer, provider }:OpenIDConnectCredentials): Promise<{ data: Session | null; error: ApiError | null }> {
+    try {
+      const headers = { ...this.headers }
+      const queryString = '?grant_type=id_token'
+      const data = await post(
+        this.fetch,
+        `${this.url}/token${queryString}`,
+        { id_token: IDToken, nonce, client_id: clientID, issuer, provider },
```
dthyresson, line 234: Could one reuse `id_token` and `client_token` instead of `IDToken` and `clientID` so that these are always consistent?

src/GoTrueClient.ts
```diff
@@ -186,7 +187,7 @@ export default class GoTrueClient {
    * @param scopes A space-separated list of scopes granted to the OAuth application.
    */
   async signIn(
-    { email, phone, password, refreshToken, provider }: UserCredentials,
+    { email, phone, password, refreshToken, provider, openIDConnect }: UserCredentials,
```
dthyresson, line 190:
```suggestion
    { email, phone, password, refreshToken, provider, oidc }: UserCredentials,
```

I am seeing more use of `oidc`in packages like https://github.com/panva/node-oidc-provider/blob/9da61e9c9dc6152cd1140d42ea06abe1d812c286/lib/actions/authorization/device_user_flow_response.js#L15

Perhaps use the abbreviation?

src/GoTrueClient.ts
```diff
@@ -558,6 +562,27 @@ export default class GoTrueClient {
     }
   }
 
+  private async _handleOpenIDConnectSignIn(
+    { IDToken, nonce, clientID, issuer, provider }: OpenIDConnectCredentials
```
dthyresson, line 566:
```suggestion
    { id_token, nonce, client_id, issuer, provider }: OpenIDConnectCredentials
```

For consistency in naming.

src/GoTrueClient.ts
```diff
@@ -558,6 +562,27 @@ export default class GoTrueClient {
...
+  ): Promise<{
+    session: Session | null
+    user: User | null
+    error: ApiError | null
+  }> {
+    if (IDToken && nonce && ( (clientID && issuer) || provider) ) {
```
dthyresson, line 572:
```suggestion
    if (id_token && nonce && ( (client_id && issuer) || provider) ) {
```
For consistency in naming.

src/lib/types.ts
```diff
@@ -174,9 +174,18 @@ export interface UserCredentials {
   refreshToken?: string
   // (Optional) The name of the provider.
   provider?: Provider
+  openIDConnect?: OpenIDConnectCredentials
```
dthyresson, line 177:
```suggestion
oidc?: OpenIDConnectCredentials
```

Consider abbreviation?

src/lib/types.ts
```diff
@@ -174,9 +174,18 @@ export interface UserCredentials {
...
 }
 
 export interface VerifyOTPParams {
   phone: string
   token: string
 }
+
+export interface OpenIDConnectCredentials {
+  IDToken: string
```
dthyresson, line 186:
```suggestion
  id_token: string
```

src/lib/types.ts
```diff
@@ -174,9 +174,18 @@ export interface UserCredentials {
...
+  nonce: string
+  provider?: Provider
+  clientID?: string
```
dthyresson, line 189:
```suggestion
  client_id?: string
```

test/GoTrueClient.test.ts
```diff
@@ -280,6 +281,19 @@ describe('GoTrueClient', () => {
     expect(user?.email).toBe(email)
   })
 
+  test('signIn with OpenIDConnect', async () => {
+    const openIDConnect:OpenIDConnectCredentials = {
+      IDToken: 'abcde',
+      nonce: 'random value',
+      provider: 'google'  
+    }
+    const { session, user, error } = await auth.signIn({openIDConnect: openIDConnect})
```
dthyresson, line 290: Why is the the same `{openIDConnect: openIDConnect}` ?

test/GoTrueClient.test.ts
```diff
@@ -280,6 +281,19 @@ describe('GoTrueClient', () => {
     expect(user?.email).toBe(email)
   })
 
+  test('signIn with OpenIDConnect', async () => {
+    const openIDConnect:OpenIDConnectCredentials = {
```
dthyresson, line 285:
```suggestion
    const oidc: OpenIDConnectCredentials = {
```

The use `oidc` in sign in.

test/GoTrueClient.test.ts
```diff
@@ -280,6 +281,19 @@ describe('GoTrueClient', () => {
...
+      IDToken: 'abcde',
+      nonce: 'random value',
+      provider: 'google'
```
dthyresson, line 288: What happens if both `client_id` and `provider` are null? or if `id_token` and `client_id` is null?

test/GoTrueClient.test.ts
```diff
@@ -280,6 +281,19 @@ describe('GoTrueClient', () => {
     expect(user?.email).toBe(email)
   })
 
+  test('signIn with OpenIDConnect', async () => {
+    const openIDConnect:OpenIDConnectCredentials = {
+      IDToken: 'abcde',
+      nonce: 'random value',
```
dthyresson, line 287: Is there a suggested way to generate a nonce?

Should the GoTrue client offer up a `generateOidcNonce()` method?

src/GoTrueClient.ts
```diff
@@ -558,6 +562,27 @@ export default class GoTrueClient {
...
+        const { data, error } = await this.api.signInWithOpenIDConnect({IDToken, nonce, clientID, issuer, provider})
+        if (error || !data) return { user: null, session: null, error }
+        this._saveSession(data)
+        this._notifyAllSubscribers('SIGNED_IN')
+        return { user: data.user, session: data, error: null }
+      } catch (e) {
+        return { user: null, session: null, error: e as ApiError }
+      }
+    }
+    throw new Error(`You must provide a OpenID Connect provider with idtoken and nonce.`)
```
dthyresson, line 583:
```suggestion
    throw new Error(`You must provide a OpenID Connect provider with your id token and nonce.`)
```

Is Client required?

It is optional from the arguments, but is it needed to properly configure OpenID Connect?

src/GoTrueClient.ts
```diff
@@ -558,6 +562,27 @@ export default class GoTrueClient {
...
+  private async _handleOpenIDConnectSignIn(
+    { IDToken, nonce, clientID, issuer, provider }: OpenIDConnectCredentials
+  ): Promise<{
+    session: Session | null
+    user: User | null
+    error: ApiError | null
+  }> {
+    if (IDToken && nonce && ( (clientID && issuer) || provider) ) {
+      try {
+        const { data, error } = await this.api.signInWithOpenIDConnect({IDToken, nonce, clientID, issuer, provider})
```
dthyresson, line 574:
```suggestion
        const { data, error } = await this.api.signInWithOpenIDConnect({id_token, nonce, client_id, issuer, provider})
```

src/GoTrueClient.ts
dthyresson: It may be that GoTrue expects `IDToken` and `clientID` but in this client, I think `id_token` and `client_id` is more consistent.

src/GoTrueClient.ts
dthyresson: Oh, it might be:


```
this.api.signInWithOpenIDConnect({IDToken: id_token, nonce, clientID: client_id, issuer, provider})
```

for GoTrue on the api side.

dthyresson (changes requested):
Added some comment around the naming of the params, specifically for case consistency with other api methods in the client.

`client_id` vs `clientID`
`id_token` vs `IDToken`

There may be cases when calling the GoTrue api that the other case is needed but as params seen in the client docs, think the consistent with the snake case options is worth the change.

and perhaps using `oidc`.

test/GoTrueClient.test.ts
koba-ninkigumi: generateOidcNonce() is not required.
Please read the oidc specification carefully.

The ability to suggest passwords to end users is not in the scope of supabase.
Similarly, the ability to suggest a nonce to each and every developer should not be in the scope of supabase.

src/GoTrueApi.ts
koba-ninkigumi: Changed.

koba-ninkigumi:
@dthyresson
I have complied with all your requests.
Did you find anything else that needs to be changed?

src/GoTrueClient.ts
koba-ninkigumi: https://github.com/supabase/gotrue/pull/360

↑The client_id and issuer do not exist until this PR is merged.

Normally, id_token (or idtoken) identifies the oidc, but "id token" does not identify the oidc.
But since it's your request, I'll go with it.

koba-ninkigumi:
@dthyresson 
My first commit was id_token.
However, after discussion within supabase, @J0 changed the document to IDToken.
I followed suit and changed it from id_token to IDToken.
Then, @dthyresson again insisted on changing it from IDToken to id_token.
I accordingly changed it from id_token to id_token.
I hope it will not ask me to change to IDToken again.

dthyresson:
> @dthyresson
> My first commit was id_token.
> However, after discussion within supabase, @J0 changed the document to IDToken.
> I hope it will not ask me to change to IDToken again.

@koba-ninkigumi I appreciate your patience and understanding as we have a few reviewers and sorry for the frustration as we go back and forth unfortunately. 

Just be sure that we want to make a great and consistent api and experience for the users — and once this feature is released and documented it will be hard to change. So, a little time upfront while I know seems tough, it will pay off in the long term as more and more developers use your feature.

Thanks again and we’ll be able to merge soon.

J0 (approved):
Thanks for the PR! I am approving this but I hope that we can address any changes/requests that @dthyresson might have before hitting merge as is much more adept at evaluating DX and may be able to catch any blind spots we might miss.

dthyresson:
> Thanks for the PR! I am approving this but I hope that we can address any changes/requests

Yes, again many thanks for your work and patience. 

Let's get this in, but maybe test out for a few weeks and if any changes need to be made, then can address  -- and then release the docs with those updates.

dthyresson (approved):

He1nr1chK:
@koba-ninkigumi @kangmingtay @dthyresson can this function be used yet? I updated my @supabase/supabase-js package to v1.30.3 which exposes this function but when I try to call it, I get `TypeError: undefined is not a function`

koba-ninkigumi:
@He1nr1chK I'm already using it with Sign In With Google, and google-one-tap is also working fine.

He1nr1chK:
@koba-ninkigumi that's fantastic news. How are you invoking the function? Like this:
```
const { user, session, error } = await supabase.auth.signInWithOpenIDConnect({
      id_token: idToken,
      nonce,
      client_id: Platform.select({
            ios: GOOGLE_IOS_CLIENT_ID,
            android: GOOGLE_ANDROID_CLIENT_ID,
      }),
      issuer: 'https://google.com',
      provider: 'google',
});
```

koba-ninkigumi:
Please see my first post on this page.

```
const { user, session, error } = await supabase.auth.signIn({
  oidc:{
      id_token: 'your idtoken',
      nonce: 'random value',
      provider: 'google'
  }
})
```

koba-ninkigumi:
@He1nr1chK 

If you want to specify Google, you should not set issuer and client_id, but write 'google' in provider and set client_id by logging in to the supabase site.

He1nr1chK:
Hi @koba-ninkigumi, when using the method with Apple as the provider I get the following error: "oidc: invalid configuration, clientID must be provided or SkipClientIDCheck must be set". In the Supabase console my Apple OAuth is correctly configured. Any idea what might be causing this?

kangmingtay:
Hey @He1nr1chK, could you include your request format? Can you try the following request format below and see if it works for you please?

```
curl -X "POST" "https://PROJECT_REF.supabase.co/auth/v1/token?grant_type=id_token" \
     -H 'apikey: apikey' \
     -H 'Content-Type: application/json; charset=utf-8' \
     -d '{
              "client_id": "client_id",
              "nonce": "nonce",
              "id_token": "id_token",
              "issuer": "https://appleid.apple.com/"
      }'
```

He1nr1chK:
Hi @kangmingtay thanks for getting back to me. This seems to have solved the issue but now I get `error {"message": "Database error saving new user", "status": 500}` even though I have used this Apple ID to sign in before.

kangmingtay:
@He1nr1chK can you please email us at support@supabase.com with your project reference so we can help to debug further? thanks!

DanMossa:
I think this PR doesn't implement linking
https://github.com/supabase/gotrue/issues/434

He1nr1chK:
Hey @WJimmyCook, yes I did with the following function:
```
const nonce = Math.random()
    .toString(36)
    .substring(2, 10);

  return Crypto.digestStringAsync(Crypto.CryptoDigestAlgorithm.SHA256, nonce)
    .then(hashedNonce =>
      AppleAuthentication.signInAsync({
        requestedScopes: [
          AppleAuthentication.AppleAuthenticationScope.FULL_NAME,
          AppleAuthentication.AppleAuthenticationScope.EMAIL,
        ],
        nonce: hashedNonce,
      }),
    )
    .then(async appleCredential => {
      const { identityToken } = appleCredential;
      const { user, session, error } = await supabase.auth.signIn({
        oidc: {
          id_token: identityToken,
          nonce,
          client_id: '<YOUR_CLIENT_ID>',
          issuer: 'https://appleid.apple.com',
        },
      });
      if (error) {
        console.log(error);
      }
    })
    .catch(error => {
      console.log(error);
    });
``` 

I use `expo-crypto` to hash the nonce before passing the hashedNonce to Apple as is required.

WJimmyCook:
@He1nr1chK You're a legend, bro! Worked like a charm.

megacherry:
@He1nr1chK: I hope you could help me also. I always get the error "invalid nonce" - {"message": "invalid nonce", "status": 400}. My provider is google and i tried both OAuth 2.0 Client IDs (the auto created from Google Service and the self created like the documentation mentioned). Both did not work. I also tried the curl-statement from @kangmingtay but it doesn't work also (same error). I don't know where the error come from (i don't think it's from gotrue-js)

My code (simplified) looks like:

```
import {
  GoogleSignin,
  statusCodes,
} from '@react-native-google-signin/google-signin'
import {v4 as uuid} from 'uuid'

GoogleSignin.configure()
const {idToken} = await GoogleSignin.signIn()
const nonce = uuid()

const {user, error} = await supabase.auth.signIn({
        oidc: {
          id_token: idToken,
          nonce,
          provider: 'google',
        },
})
```

WJimmyCook:
@megacherry 
- Hash the nonce
- passed the hashed nonce to the google auth request
- pass the unhashed nonce to supabase
Here's the general idea
 ```
import * as Google from 'expo-auth-session/providers/google';
import * as Crypto from 'expo-crypto';
const nonce = Math.random()
    .toString(36)
    .substring(2, 10);
...
  const [hashedNonce, setHashedNonce] = useState(null);
   useEffect(() => {
        Crypto.digestStringAsync(Crypto.CryptoDigestAlgorithm.SHA256, nonce).then(hNonce => setHashedNonce(hNonce));
    }, []);
...
    const [request, response, promptAsync] = Google.useIdTokenAuthRequest({
        androidClientId,
        extraParams: { nonce: hashedNonce },
    });
...
 const { user, session, error } = await supabase.auth.signIn({
        oidc: {
            id_token: response.authentication.idToken,
            nonce,
            provider: 'google',
        },
    });
```

**Comment by WJimmyCook** [comment]
@megacherry not sure what non-expo options there are but I'm running a bare react-native project and I ended up adding expo unimodules (or whatever they call it now) so I can use any expo lib in the bare project.

**Comment by wweevv-johndpope** [comment]
I'm hitting a wall with oidc

```javascript
	const { user, session, error } = await supabase.auth.signIn({
        oidc:{
            id_token: identityToken,
            nonce: nonce,
            provider: 'apple'
        }
      })

```
doesn't seem like this got ported to v2 ? am I missing something? 

 supabase.auth.signIn is not a function

WJimmyCook:
@wweevv-johndpope I think oidc was removed from v2 but there are plans to add it back based on this comment: https://github.com/supabase/gotrue/issues/434#issuecomment-1287755049

I'm sticking with v1 until that get fixed.

ZhenFTW:
Hoping the team add it on v2 too, I've been configuring my project on supabase v2 and didn't know that this feature is not added.   If the feature is already added on v1, I would appreciate supporting it back on v2 even just an option as experimental as there is still a lot of issues related to OIDC implementation.

Issue #208, opened 2022-01-15
Feat/get session from url param


Lxstr:
# Feature request
Allows the user to optionally pass a url manually. It appears this might have been originally intended from the looks of the code.

## Is your feature request related to a problem? Please describe.

This allows the programmer to use callbacks that might otherwise be missed such as receiving a universal link from a capacitor app. i.e.
`
if (Capacitor.isPluginAvailable('App')) {
      App.addListener('appUrlOpen', async urlOpen => {
        const { data, error } = await supabase.auth.getSessionFromUrl(
          {
            storeSession: true,
          },
          urlOpen.url,
        )
      })
    }
`

This is equivalent to the auth0 method listed here [https://auth0.com/docs/quickstart/native/ionic-react/01-login](url)

This would not be possible otherwise without these changes.

## Describe the solution you'd like

Include url param in getSessionFromUrl

## Describe alternatives you've considered

This means going to the browser and back again to the native app so it is not 100% ideal but could be an intermediate step towards a fully fledged capacitor plugin that would match the features of [https://github.com/robingenz/capacitor-firebase-authentication](url) or [https://github.com/baumblatt/capacitor-firebase-auth](url). Something like this would be amazing but beyond my current skills.

hf:
Hey @Lxstr the `getSessionFromUrl` method, although public, is intended to be used in very advanced use cases and on the browser only. If you need to "intercept" a URL you can do so _only_ before the Supabase client constructor runs. Once the constructor runs, it will override the `window.location.hash = ''`. 

This should help you achieve your goal. I'll close this issue, but feel free to comment or ask further questions.

Lxstr:
@hf thanks for the response. Apologies it's been a while since I was working on this one so I can't quite visualise your suggested solution. What do you mean by before the constructor? How would that look?

PR #209, opened 2022-01-15
Feat/get session from url param

Lxstr:

Achieves #208.

Allows the user to optionally pass a url manually. It appears this might have been originally intended from the looks of the code.

Supabase/gotrue method only gets url from document url.

Capacitor native app webview does not seem to be able to get a token from a normal auth callback.

## What is the new behavior?

Force capacitor to get the token by using url param in the getSessionFromUrl method.

`
if (Capacitor.isPluginAvailable('App')) {
      App.addListener('appUrlOpen', async urlOpen => {
        const { data, error } = await supabase.auth.getSessionFromUrl(
          {
            storeSession: true,
          },
          urlOpen.url,
        )
      })
    }
`

## Additional context

This is equivalent to the auth0 method listed here [https://auth0.com/docs/quickstart/native/ionic-react/01-login](url)


Lxstr:
Would changing 
`|| null`
to
`|| undefined`
fix the errors?

hf:
Hi, thank you for your contribution! The library has advanced quite a bit from then. I'll close this PR for now, but do open a new one if you would like the same change to appear in v2.

PR #216, opened 2022-01-27
chore: adds notion and alphabetize

awalias:
to match https://github.com/supabase/supabase/pull/5018#pullrequestreview-864341046

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -1,16 +1,19 @@
 export type Provider =
+  | 'apple'
   | 'azure'
   | 'bitbucket'
+  | 'discord'
   | 'facebook'
   | 'github'
   | 'gitlab'
   | 'google'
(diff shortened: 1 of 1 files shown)


soedirgo (approved):

kangmingtay:
we also have 2 new sms providers (vonage & textlocal) and another oauth provider (linkedin), can we add those in as well?

src/lib/types.ts
```diff
@@ -1,16 +1,19 @@
 export type Provider =
```
kangmingtay, line 1: we also have 2 new sms providers (vonage & textlocal) and another oauth provider (linkedin), can we add those in as well?

kangmingtay (changes requested):

kangmingtay (approved):

PR #227, opened 2022-02-15
Handle refresh offline

### LINKED ISSUE #226: Token refresh fails silently when offline and never recovers
# Bug report

## Describe the bug
If the user is offline when their token expires then `_callRefreshToken` will fail. The failure is not handled well so there's no indication that there was a problem (except that the token no longer works) and there's no attempt to refresh again.

`_startAutoRefreshToken` uses `setTimeout(() => this._callRefreshToken(), value)`
`_callRefreshToken` does not throw errors, it returns them so the get ignored here. The token is not refreshed and a new timeout is not set.

Related: #141

## To Reproduce

1. Log in
2. Go offline
3. Wait for token to expire
4. Go online
5. Try to use the token

## Expected behavior

* The error should not be silently swallowed
* There should be another attempt to refresh the token at a later time

jake-nz:

Bug fix

Closes #226

## What is the new behavior?

Does not try to refresh the token when offline, adds an event listener to refresh once online

## Additional context

Happy to discuss alternatives. I needed to make this change for a project so I thought I'd contribute it here.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -61,6 +61,7 @@ export default class GoTrueClient {
   protected multiTab: boolean
   protected stateChangeEmitters: Map<string, Subscription> = new Map()
   protected refreshTokenTimer?: ReturnType<typeof setTimeout>
+  protected onlineListener?: () => void
 
   /**
    * Create a new client for use in the browser.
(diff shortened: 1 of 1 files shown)


hf:
I'll close this PR because in v2 (I just merged) some code that allows you to call `startAutoRefresh` and `stopAutoRefresh` as you please. The `online` event is not perfect as it just means the computer is connected to a network, not that there is internet.

PR #228, opened 2022-02-16
fix: set expiresAt before returning verifyMobileOtp

### LINKED ISSUE #118: Phone Auth Verify OTP did not stored in Local Storage
# Bug report

## Describe the bug

Phone auth OTP and verification no problem. But the `supabase.auth.token` is not stored in Local Storage. 

image

## To Reproduce

1. Your code
```
    let { session, error } = await supabase.auth.signIn({
      phone: phoneNumber
    })
```

```

    let { session, error } = await supabase.auth.verifyOTP({
      phone: phoneNumber,
      token: tokenString,
    })

```

2. Check localstorage / Get session using supabase.auth.user()  `// return null`

## Expected behavior

User token should be stored in localstorage.

## System information

- OS: Window 10
- Browser (if applies) Edge & Chrome
- Version of supabase-js: [e.g. 1.21.0]
- Version of Node.js: [e.g. 14.16.1]

kangmingtay:
* Fixes #118 

## To reproduce
* cd into the example react project and change the `GOTRUE_URL_AUTOCONFIRM` to `http://localhost:9999`
* Run gotrue locally on port 9999 (defaults to port 9999 anyway)
* Make sure to set the credentials for your sms provider in gotrue's `.env` file
* Notice the session doesn't get set in local storage


src/GoTrueClient.ts
```diff
@@ -632,7 +644,7 @@ export default class GoTrueClient {
         } else {
           this._removeSession()
         }
-      } else if (!currentSession || !currentSession.user) {
```
thorwebdev, line 635: @kiwicopple do you recall why this check was introduced? Is it fine to be removed? Currently the verify methods to not return the user object. While @kangmingtay is making a PR in gotrue to return the user object on the verify endpoints (https://github.com/supabase/gotrue/issues/141) it likely won't be available to users until next week. So we're contemplating removing this check here so this issue is resolved before the gotrue server changes go live.

thorwebdev (approved):
LGTM, thanks

PR #242, opened 2022-03-08
fix: properly default shouldCreateUser flag

emroussel:

Bug fix

The `shouldCreateUser` flag was added in #199, but passing `{ shouldCreateUser: false }` into the options of `signIn` still creates the user because this ternary will always return `true`:
```js
const shouldCreateUser = options.shouldCreateUser ? options.shouldCreateUser : true
```

## What is the new behavior?

With this change the `shouldCreateUser` can be set to `false` and the user will not be created, but it is still defaulted to `true` if no value (`undefined` or `null`) is passed.

## Additional context

Related to: https://github.com/supabase/supabase/issues/1176

### DIFF
--- a/src/GoTrueApi.ts
+++ b/src/GoTrueApi.ts
@@ -269,7 +269,7 @@ export default class GoTrueApi {
         queryString += '?redirect_to=' + encodeURIComponent(options.redirectTo)
       }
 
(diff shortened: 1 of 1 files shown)


denisdani (approved):

kangmingtay (approved):

kangmingtay:
Hey @emroussel, thanks for spotting this and making the fix!

PR #249, opened 2022-03-13
refactor: add keepCurrentSession option for signUp()

phulengo:

Add a boolean variable named `keepCurrentSession` to `signUp()` function.

Currently, there are not any options for developers to decide to keep or remove the current session after signing up.

## What is the new behavior?

Developers can actively decide whether the session should be kept or reset after signing up.
image

## Additional context

This idea was initiated when I'm building a system using Supabase where users with admin roles (store in public tables with UUID reference to the auth table) can create and manage other user account types (like staff, coordinator). However, when I use the signUp function in the Create New User page, the current session (admin) was removed and replaced by the credential of the new account. 
It's somehow really annoying, and I think this refactoring doesn't cause any conflicts to other parts, so it only improves the flexibility of usage for developers. I hope this PR will be merged as soon as possible because I've been stuck for a long time in my project before deciding to create this PR. Many thanks, GoTrue Team!

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -120,13 +120,16 @@ export default class GoTrueClient {
    * @param phone The user's phone number.
    * @param redirectTo A URL or mobile address to send the user to after they are confirmed.
    * @param data Optional user metadata.
+   * @param keepCurrentSession Optional keep current session or not
    */
   async signUp(
     { email, password, phone }: UserCredentials,
     options: {
       redirectTo?: string
       data?: object
-      captchaToken?: string
+      captchaToken?: string,
+      keepCurrentSession?:boolean
+
     } = {}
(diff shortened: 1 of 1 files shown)


pixtron:
This PR seems to introduce quite some complexity. With every additional setting the amount of test cases increases exponentially.

Even without this PR it is possible to sign up a user without changeing the local session.
The API is a public property of the client, so instead of `supabase.auth.singUp({email, password})`, you can just call `supabase.auth.api.signUpWithEmail(email, password)`. 

All the api methods you can find in here: https://github.com/supabase/gotrue-js/blob/master/src/GoTrueApi.ts

hf:
Thank you for your contribution. The API has advanced quite a bit, so please submit a new PR if you still want to see this.

PR #251, opened 2022-03-17
fix: add verify email otp to verifyOTP

kangmingtay:
* Adds support for email otp verification (https://github.com/supabase/gotrue/pull/379)


src/GoTrueApi.ts
```diff
@@ -347,6 +348,43 @@ export default class GoTrueApi {
...
 
+  /**
+   * Send User supplied Email / Mobile OTP to be verified
+   * @param email The user's email address
+   * @param phone The user's phone number WITH international prefix
+   * @param token token that user was sent to their mobile phone
+   * @param type verification type that the otp is generated for
+   * @param redirectTo A URL or mobile address to send the user to after they are confirmed.
+   */
+  async verifyEmailOrMobileOTP(
```
thorwebdev, line 376: Can we just call this `verifyOTP` and then mark `this.api.verifyMobileOTP` as deprecated?

thorwebdev (changes requested):

src/GoTrueApi.ts
kangmingtay: do we still want to keep [verifyOTP](https://github.com/supabase/gotrue-js/pull/251/files#diff-3522461172efd6058d6b8da62fc2d30d8b524d2b64894ea2c67218c52f7fdff5R267) in `GoTrueClient`?

kangmingtay: I'm with marking `this.api.verifyMobileOTP` as deprecated but it's being used in `verifyOTP` in `GoTrueClient`. How should we differentiate the `verifyOTP` in `GoTrueClient` and `GoTrueApi`?

J0 (approved):

src/GoTrueApi.ts
```diff
@@ -336,7 +337,7 @@ export default class GoTrueApi {
   }
 
   /**
-   * Send User supplied Mobile OTP to be verified
+   * @deprecated Send User supplied Mobile OTP to be verified
```
kangmingtay, line 340: marked `verifyMobileOTP` as deprecated so we can potentially remove this in the future

src/GoTrueApi.ts
```diff
@@ -364,6 +365,43 @@ export default class GoTrueApi {
...
+          `${this.url}/verify`,
+          { email, token, type: type, redirect_to: options.redirectTo },
+          { headers }
+        )
+        : await post(
+          this.fetch,
+          `${this.url}/verify`,
+          { phone, token, type: 'sms', redirect_to: options.redirectTo },
+          { headers }
+        )
```
kangmingtay, line 396: @thorwebdev i still need to check for `email` and `type` because they will be passed into the object as `{ email: undefined, phone: ..., token: ..., type: undefined }` if i don't

src/GoTrueClient.ts
```diff
@@ -258,12 +258,14 @@ export default class GoTrueClient {
...
    * Log in a user given a User supplied OTP received via mobile.
+   * @param email The user's email address.
    * @param phone The user's phone number.
    * @param token The user's password.
+   * @param type The user's verification type.
    * @param redirectTo A URL or mobile address to send the user to after they are confirmed.
    */
   async verifyOTP(
-    { phone, token }: VerifyOTPParams,
+    { email, phone, token, type = 'sms' }: VerifyOTPParams,
```
kangmingtay, line 268: default to type = 'sms' here to maintain backward compatability

thorwebdev (approved):
LGTM

PR #252, opened 2022-03-19
chore: refactor

thorwebdev:
@kangmingtay I've refactored the types and simplified the api and client a bit. 

With the types like this, `type` is optional when providing `phone` but is required when providing `email` 👍

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -185,19 +185,21 @@ export interface UserCredentials {
   oidc?: OpenIDConnectCredentials
 }
 
-export interface VerifyOTPParams {
-  email?: string
-  phone?: string
+export type VerifyOTPParams = VerifyMobileOTPParams | VerifyEmailOTPParams
+export interface VerifyMobileOTPParams {
+  email?: undefined
+  phone: string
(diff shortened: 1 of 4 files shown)


src/lib/types.ts
```diff
@@ -185,19 +185,21 @@ export interface UserCredentials {
...
 }
-export type OTPType = 
-  | 'signup' 
-  | 'invite' 
-  | 'magiclink' 
-  | 'recover' 
-  | 'email_change' 
-  | 'sms'
+export type MobileOTPType = 'sms'
+export type EmailOTPType = 'signup' | 'invite' | 'magiclink' | 'recover' | 'email_change'
```
thorwebdev, line 202: @kangmingtay can you double check that this separation is correct?

kangmingtay: hmm seems like it should be 'recovery' instead of 'recover' (https://github.com/supabase/gotrue/blob/master/api/verify.go#L24)

kangmingtay (changes requested):

kangmingtay (approved):

PR #254, opened 2022-03-22
fix: update typedocs for signUp and signIn

kangmingtay:
* `redirect_to` doesn't work for non-oauth sign-ups / sign-ins because those requests are POST requests. This means that we cannot redirect them without converting them into a GET request which would result in losing the response body.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -3,6 +3,7 @@ import { isBrowser, getParameterByName, uuid } from './lib/helpers'
 import { GOTRUE_URL, DEFAULT_HEADERS, STORAGE_KEY } from './lib/constants'
 import { polyfillGlobalThis } from './lib/polyfills'
 import { Fetch } from './lib/fetch'
+import { Logger } from './lib/logger'
 
 import type {
   ApiError,
@@ -135,7 +136,8 @@ export default class GoTrueClient {
   }> {
     try {
       this._removeSession()
-
+      Logger.warn('redirectTo param will only redirect OAuth logins.')
+      
       const { data, error } =
         phone && password
           ? await this.api.signUpWithPhone(phone!, password!, {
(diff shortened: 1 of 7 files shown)


src/GoTrueClient.ts
```diff
@@ -135,7 +136,8 @@ export default class GoTrueClient {
   }> {
     try {
       this._removeSession()
-
+      Logger.warn('redirectTo param will only redirect OAuth logins.')
```
kangmingtay, line 139: not sure why we added a `redirect_to` to the `signUp` method - it won't work here as well since these are all POST requests

kiwicopple (changes requested):
Sorry @kangmingtay - we can't include pino in these libraries because we are trying to keep them as small as possible. Perhaps just a regular console.warn will suffice?

thorwebdev:
Are we getting a lot of support requests about this? I personally haven't seen any, so wondering if this is really necessary.

kangmingtay:
> Are we getting a lot of support requests about this? I personally haven't seen any, so wondering if this is really necessary.

yeah i've seen this appear quite a few times on the github discussions / issues - it also seems confusing that the `redirect_to` param only applies when you use oauth and not other forms of authentication

GaryAustin1:
A document note on it would would be enough IMO.  I've seen it asked several times in Discord and have had to answer "I don't think that does anything"  for email/password, as I never found a clear answer, even in the source.
```
      if (email && password) {
        return this._handleEmailSignIn(email, password, {
          redirectTo: options.redirectTo,
        })
```
So you have to go dig thru gotrue to figure out the answer, or try it (which I did, so I could give an educated answer).

kangmingtay:
thanks for the feedback everyone! i've corrected the typedocs so it's clearer now

kiwicopple (approved):

PR #256, opened 2022-03-22
fix: unify localStorage acess

thorwebdev:
Fixes the `TypeError: Cannot read properties of undefined (reading 'getItem')` error in Deno land.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -713,14 +713,14 @@ export default class GoTrueClient {
 
   private _persistSession(currentSession: Session) {
     const data = { currentSession, expiresAt: currentSession.expires_at }
-    isBrowser() && this.localStorage.setItem(STORAGE_KEY, JSON.stringify(data))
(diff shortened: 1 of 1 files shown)


src/GoTrueClient.ts
```diff
@@ -605,9 +605,9 @@ export default class GoTrueClient {
    * Attempts to get the session from LocalStorage
    * Note: this should never be async (even for React Native), as we need it to return immediately in the constructor.
    */
-  private _recoverSession() {
+  private async _recoverSession() {
```
kiwicopple, line 608: I thought this one needed to be `sync` because it's used in the constructor?

(not 100% sure, just want to flag it)

thorwebdev: Do you remember where this came up? Was this the one where in svelte folks where running into issues with user being null initially or something?

 `_recoverAndRefresh()` is async and is called in the constructor, so I thought it should be fine. But could also just do the optional chaining and not the async change.

PR #258, opened 2022-03-28
fix: add methods for update password reauthentication

kangmingtay:
* Provides corresponding methods for gotrue-js for password reauthentication
```
// To send a reauthentication request
supabase.auth.reauthenticate()

// To update password with reauthentication
supabase.auth.update({
  "password": "newpassword",
  "nonce": "123456"
})
```

## Steps to test
1. Enable `Update password requires reauthentication` on the dashboard
2. Initialise supabase client
3. Sign up with any method (password-based, passwordless or oauth)
4. Request for reauthentication `supabase.auth.reauthenticate()`
5. Update your password by calling `supabase.auth.update({...})`

### DIFF
--- a/src/GoTrueApi.ts
+++ b/src/GoTrueApi.ts
@@ -832,4 +832,19 @@ export default class GoTrueApi {
       return { user: null, data: null, error: e as ApiError }
     }
   }
+
+  /**
+   * Sends a reauthentication otp to the user's email or phone number
+   * @param jwt A valid, logged-in JWT.
+   */
+  async reauthenticate(jwt: string): Promise<{ data: {} | null; error: ApiError | null }> {
+    try {
+      const data: any = await get(this.fetch, `${this.url}/reauthenticate`, {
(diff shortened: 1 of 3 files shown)


J0 (approved):

kangmingtay:
hmm actually maybe we shouldn't default `sendOTP()` to take the JWT from the session? That seems kinda strange to me - maybe we should have another option like `useJWT`:
```
async sendOTP(
    email?: string,
    phone?: string,
    options: {
      shouldCreateUser?: boolean
      useJWT?: boolean
      captchaToken?: string
    } = {}
  ): Promise<{ data: {} | null; error: ApiError | null }> 
```

Technically, calling `sendOTP()` would still send an OTP to either the user's email or phone number (email is preferred). But the main use of this would be for reauthenticating on password update which doesn't seem clear to me just by looking at the function signature. Wdyt? 

cc @awalias @inian

kangmingtay:
closing this in favour of #688

Issue #143, opened 2021-10-08
user() is null on first invocation after Magic Link sign in


joanllenas:
# Bug report

## Describe the bug

When you sign in via Magic Link from an email, the user is `null` initially.
In order to successfully initialize my app I have to use a `setTimeout`:

```ts
if (supabase.auth.user()) {
  initializeApp(supabase);
} else {
  setTimeout(() => initializeApp(supabase), 1000);
}
```

## To Reproduce

```ts
const supabase = createClient(supabaseUrl, supabaseAnonKey);
console.log("supabase.auth.token", localStorage.getItem("supabase.auth.token")); // null
console.log("User", supabase.auth.user()); // null
supabase.auth.onAuthStateChange((evt, session) => {
  console.log("onAuthStateChange.evt", evt); // 'SIGNED_IN'
  console.log("onAuthStateChange.session", session); // { access_token, user, (...) }
});
```

## Expected behavior

I would say that there are at least two possible expected behaviours:
1. Ideally, make sure that `supabase.auth.user()` is synchronous for all cases.
2. If 1. is not possible, add a new `'NOT_SIGNED_IN'` event to `supabase.auth.onAuthStateChange`.

## System information

- OS: macOS
- Browser chrome 94.0.4606.71
- Version of supabase-js: 1.24.0
- Version of Node.js: 14.17.6

## Additional context

Related to https://github.com/supabase/gotrue-js/issues/23

chipilov:
I think the problem goes deeper than this and doesn't only affect magic links.

The issue is that on initial page load, there are 3 possible scenarios:

1. There is no token stored in local storage - in this case, supabase.auth.user() will return null immediately after calling createClient() and will stay null until the user logs in explicitly;
2. There is a token stored in local storage and it has NOT expired - in this case, supabase.auth.user() will return non-null immediately after calling createClient() (as long as localStorage API used is synchronous, which is the case for browsers);
3. There is a token stored in local storage, however, it is expired and needs to be refreshed BEFORE the user is considered authenticated - in this case, supabase.auth.user() will return null immediately after calling createClient(), however, as soon as the request for refreshing the token succeeds, onAuthStateChanged handlers will be invoked and supabase.auth.user() will return non-null.

[...]

Issue #23, opened 2020-11-06
session() and user() reportedly not returning data on first invocation


awalias:
![image](https://user-images.githubusercontent.com/458736/98378900-9e155400-2081-11eb-91e3-3dcbff48f000.png)

awalias:
@thorwebdev or @kiwicopple might have ideas here

thorwebdev:
Both `client.session()` and `client.user()` are synchronous methods that only return a value when there is an active session, otherwise they are expected to return null. 

Do you have any more context as to why they are expecting a session to be existent?

SweeToxin:
Thank you @awalias and Hi @thorwebdev. 
To put some context onto this. 
I have a function that fires on app load to find if there's a user or a session and set my app storage to the loggedin user. 

1: signIn is not called initially since this goes on app load. 
2: This is used for 2 cases ( as of now ) the first is to set the user after a external auth/reset password redirect, the second one is on app reload/refresh to verify if there's a session as to not ask the user to login after each page reload. 

the behavior noticed is that when called once ( or only one of them ) it still returns null while the localstorage is already filled with the session info. 

Hope this was clear enough, if you require more info I would be glad to answer.

thorwebdev:
@SweeToxin I'm not able to reproduce this, can you provide a CodeSandbox that shows this behaviour?

There are two scenarios here:

### 1. Check on page load whether a user is logged in:

This works synchronously as expected. When GoTrue is initialized (which happens when the supabase client is created) the session is recovered synchronously from localstorage (https://github.com/supabase/gotrue-js/blob/master/src/GoTrueClient.ts#L57) and `supabase.auth.user()` returns the user object synchronously on first call. 

### 2. Login from link (confirm signup / recover password / magic link)

This is an asynchronous action as GoTrue needs to get the access_token from the URL and then retrieve the session from the server (https://github.com/supabase/gotrue-js/blob/master/src/GoTrueClient.ts#L187-L209). 

So in this scenario, `supabase.auth.user()` will initially return `null` until the session & user data has been loaded. This is where you want to use the `supabase.auth.onAuthStateChange` listener. 

So in a React application, you will want to do something like this:

```js
useEffect(() => {
    setSession(supabase.auth.session()); // will return the user object for scenario 1 // null for scenario 2
    setUser(supabase.auth.user()); // will return the user object for scenario 1 // null for scenario 2

[...]

PR #265, opened 2022-04-01
feat: add async getters for session and user.

lebrinkma:
adds async getters for session and user.

this provides a workaround for #143 and #23 without needing setTimeout.
this is an alternate solution to #147.

After `const supabase = createClient(supabaseUrl, supabaseKey)`
`supabase.auth.session()` and `supabase.auth.user()` might return `null`,
if the session is expired and currently refreshing or if the session is recovering from a magic link.

## What is the new behavior?
`refreshedSession()` resp. `refreshedUser()` wait for the pending recovery and
and return valid `session` resp. `user` data.

usage examples
```
const session = await supabase.auth.refreshedSession()
const user = await supabase.auth.refreshedUser()
```

## Additional context
for supabase-js v2 (supabase/supabase-js#170)
user()  and session() should IMHO be async functions.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -61,6 +61,7 @@ export default class GoTrueClient {
   protected multiTab: boolean
   protected stateChangeEmitters: Map<string, Subscription> = new Map()
   protected refreshTokenTimer?: ReturnType<typeof setTimeout>
+  protected refreshPending: Promise<void>
 
   /**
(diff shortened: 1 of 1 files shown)


alaister:
Hey @lebrinkma,

Thanks for this - sorry I didn't see it sooner!

This is now being tracked in #285. Any comments you have are welcome there :)

Issue #213, opened 2022-01-20
Concurrent token refreshes can lead to unintentional signing out of users when multiple tabs are open


bnjmnt4n:
# Bug report

## Describe the bug

`gotrue-js` uses `localStorage` to store the refresh token. When the user has multiple tabs open, multiple instances of the `GoTrueClient` will be instantiated, with each reading from `localStorage` the same identical access and refresh tokens. There can be cases where both tabs will attempt to refresh using the same token at the same time, which leads to the first request succeeding, but the second request failing due to token reuse detection (https://github.com/supabase/gotrue/pull/226).

## To Reproduce

1. Reduce the JWT expiry from the default 1 hour to 1 minute or lesser.
2. Login into an app using Supabase in a single tab.
3. Open the app in a second tab.
4. Open the network inspector and observe refresh token requests. Eventually both tabs will have failing requests due to reuse of the same refresh token.

## Expected behavior

There is some kind of locking mechanism which avoids concurrent refreshes using the same refresh token. https://github.com/supabase/gotrue-js/pull/203 attempts to do this, but it doesn't account for the multi-tab use case.

## System information

This is likely reproducible on other browsers, but here's the one I tested with:

- OS: Windows
- Browser (if applies): Firefox
- Version of supabase-js: 1.28.6

## Additional context

I will be testing out some changes to `gotrue-js` and will send a PR if it successfully fixes this issue.

TheOnlyBeardedBeast:
@bnjmnt4n @inian I think this issue is not frontend related issue, I went deep into the gotrue server implementation, everytime you request a new refresh token, gotrue revokes the old one. As you are in different tabs with shared refresh tokens, the tab which does the refresh first revokes the refresh token for the other tabs, so the other tabs cant get a new access token. Overcoming this issue means that only one tab can make the refresh and it needs to share the access and refresh tokens with the others. Maybe the broadcast chanell api? The second option is to not revoke the refresh token, but that would blow the security.

From the GoTrue server source code, clearly visible that it revokes the old token.
```golang
// GrantRefreshTokenSwap swaps a refresh token for a new one, revoking the provided token.
func GrantRefreshTokenSwap(tx *storage.Connection, user *User, token *RefreshToken) (*RefreshToken, error) {
	var newToken *RefreshToken
	err := tx.Transaction(func(rtx *storage.Connection) error {
		var terr error
[...]

PR #274, opened 2022-04-21
fix: refresh token race condition

thorwebdev:

Prevent token refresh race conditions and trigger `recoverAndRefresh` on visibility change.

Currently our syncing across multiple tabs causes the tabs to race with each other because the setTimeout is set at exactly the same time when the user logs in in any of the tabs, as it is synced to the other tabs via local storage. All tabs then make the refreshAccessToken call at the same time, causing them to be invalidated.

Related issues: https://github.com/supabase/gotrue-js/issues/213 https://github.com/supabase/supabase/issues/6464 https://github.com/supabase/supabase/issues/5990

## What is the new behavior?

The new strategy is:

- add a `localStorage` lock check to tell other tabs that a refresh request is in progress
- add a 0-500ms random offset to the timeout to prevent most race conditions
- use `visibilitychange` event to clearTimeout and restart autorefresh

## Additional context

cc @GaryAustin1 @bnjmnt4n

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,7905 @@
 {
   "name": "@supabase/gotrue-js",
   "version": "0.0.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "@supabase/gotrue-js",
+      "version": "0.0.0",
+      "license": "MIT",
+      "dependencies": {
+        "cross-fetch": "^3.0.6"
+      },
+      "devDependencies": {
+        "@types/faker": "^5.1.6",
+        "@types/jest": "^26.0.13",
+        "@types/jsonwebtoken": "^8.5.6",
+        "@types/node": "^14.14.7",
+        "@types/node-fetch": "^2.5.7",
+        "@typescript-eslint/eslint-plugin": "^5.0.0",
+        "@typescript-eslint/parser": "^5.0.0",
+        "eslint": "^8.4.1",
+        "eslint-config-prettier": "^8.3.0",
+        "eslint-config-standard": "^16.0.3",
+        "eslint-plugin-import": "^2.25.3",
+        "eslint-plugin-node": "^11.1.0",
+        "eslint-plugin-promise": "^5.2.0",
+        "faker": "^5.3.1",
+        "genversion": "^3.0.1",
+        "jest": "^27.3.1",
+        "jsonwebtoken": "^8.5.1",
+        "npm-run-all": "^4.1.5",
+        "prettier": "2.4.1",
+        "rimraf": "^3.0.2",
+        "semantic-release-plugin-update-version-in-files": "^1.1.0",
+        "ts-jest": "^27.0.7",
+        "typedoc": "^0.22.7",
+        "typescript": "^4.0.2"
+      }
+    },
+    "node_modules/@babel/compat-data": {
+      "version": "7.16.0",
+      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.16.0.tgz",
+      "integrity": "sha512-DGjt2QZse5SGd9nfOSqO4WLJ8NN/oHkijbXbPrxuoJO3oIPJL3TciZs9FX+cOHNiY9E9l0opL8g7BmLe3T+9ew==",
+      "dev": true,
+      "engines": {
+        "node": ">=6.9.0"
+      }
+    },
+    "node_modules/@babel/core": {
+      "version": "7.16.0",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.16.0.tgz",
+      "integrity": "sha512-mYZEvshBRHGsIAiyH5PzCFTCfbWfoYbO/jcSdXQSUQu1/pW0xDZAUP7KEc32heqWTAfAHhV9j1vH8Sav7l+JNQ==",
+      "dev": true,
+      "dependencies": {
+        "@babel/code-frame": "^7.16.0",
+        "@babel/generator": "^7.16.0",
+        "@babel/helper-compilation-targets": "^7.16.0",
+        "@babel/helper-module-transforms": "^7.16.0",
+        "@babel/helpers": "^7.16.0",
+        "@babel/parser": "^7.16.0",
+        "@babel/template": "^7.16.0",
+        "@babel/traverse": "^7.16.0",
+        "@babel/types": "^7.16.0",
+        "convert-source-map": "^1.7.0",
+        "debug": "^4.1.0",
+        "gensync": "^1.0.0-beta.2",
+        "json5": "^2.1.2",
+        "semver": "^6.3.0",
+        "source-map": "^0.5.0"
+      },
+      "engines": {
+        "node": ">=6.9.0"
+      },
+      "funding": {
+        "type": "opencollective",
+        "url": "https://opencollective.com/babel"
+      }
+    },
+    "node_modules/@babel/core/node_modules/@babel/code-frame": {
+      "version": "7.16.0",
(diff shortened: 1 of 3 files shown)


src/GoTrueClient.ts
```diff
@@ -643,7 +644,7 @@ export default class GoTrueClient {
       const { currentSession, expiresAt } = data
       const timeNow = Math.round(Date.now() / 1000)
 
-      if (expiresAt < timeNow) {
+      if (expiresAt < timeNow + 10) {
```
thorwebdev, line 647: Maybe this should be 60s to give you more buffer? However when you set JWT expiry to 30s for example it then refreshes constantly. @kangmingtay maybe we should enforce a minimum expiry of 120 seconds?

liaujianjie:
Awesome stuff!! Does change this handle the case where a tab is closed mid-refresh? A lock could be acquired and not released.

thorwebdev:
@liaujianjie it does now, thanks for pointing it out! 💚

src/GoTrueClient.ts
thorwebdev: I've changed it to 60 seconds to be consistent with the setTimeout `refreshDurationBeforeExpires`. However this means if your JWT expiry is below 60 seconds you're constantly calling gotrue. So we should probably enforce a minimum JWT expiry?

package-lock.json
```diff
@@ -1,8 +1,7905 @@
 {
   "name": "@supabase/gotrue-js",
   "version": "0.0.0",
-  "lockfileVersion": 1,
```
inian, line 4: remove this change Thor? Lock files should be v1 for now

src/GoTrueClient.ts
```diff
@@ -634,7 +635,7 @@ export default class GoTrueClient {
    */
   private async _recoverAndRefresh() {
     try {
-      const json = isBrowser() && (await this.localStorage.getItem(STORAGE_KEY))
+      const json = isBrowser() && (await this.localStorage?.getItem(STORAGE_KEY))
```
inian, line 638: extract localstorage.(get, set, removeitem) into a utility cos we seem to be doing isbrowser() everytime before accessing it and will lead to errors if we dont do that guard. (can be in a separate PR)

inian (commented):
I think the flow is still a bit more complex than it needs to be, lets discuss offline

GaryAustin1:
@thorwebdev I just took a quick glance an the visibility handler.  You probably do not want to automatically refresh on every visible event.  You should do the expiration test for what ever time you agree to before renewing.

That event fires for any tab change on a browser. Desktop/mobile does not matter.  

If the user tabs to check the stock market or news constantly, getting a new token will happen every time.  
I know in my environment I've got 15 tabs open and move between 5 of them constantly during the day.

GaryAustin1:
If you are attempting to fix gotrue.js token refresh in general (versus just token race condition), I believe this issue of token refresh fails if the device is offline https://github.com/supabase/gotrue-js/issues/226 (they also have a PR that seems overly complex to me) might need be addressed as well.  I've not researched the online event handler.  Also wondering what happens if the server/internet connection is down right at moment of token refresh timer going off.

src/GoTrueClient.ts
```diff
@@ -673,7 +674,27 @@ export default class GoTrueClient {
...
+        isBrowser() && (await this.localStorage?.getItem(`${STORAGE_KEY}-tokenRefreshLock`))
+      if (tokenRefreshLock === 'IN_PROGRESS') {
+        // Release lock after 5s in case original lock setter has dropped off.
+        if (isBrowser())
+          setTimeout(
+            () => {
+              this.localStorage?.removeItem(`${STORAGE_KEY}-tokenRefreshLock`)
+              this._recoverAndRefresh()
+            },
+            5 * 1000 // remove lock after 5s
```
kangmingtay, line 688: would it help if we randomise this `setTimeout` duration too? in the event there are multiple tabs waiting for the lock, there could be multiple instances of `setTimeout` invoked at the same time?

maybe it's better if the original lock setter calls `setTimeout` to remove the lock after 5 seconds after it acquires the lock [here](https://github.com/supabase/gotrue-js/pull/274/files#diff-3522461172efd6058d6b8da62fc2d30d8b524d2b64894ea2c67218c52f7fdff5R693)

GaryAustin1: On the lock with a timer... What happens if the tab that set the lock gets closed right after that?
I'll admit I'm not up to speed on the need for multiple tabs and what the user can do with these tabs, but they can certainly close one at any point.

kangmingtay: ok i just tested out the following and closed the tab before the timeout expired:
```
setTimeout(function(){
  console.log("Hello from setTimeout")
}, 5000);
```
seems like once you close the tab, the timeout gets cleared so yeah it needs to be called by another tab

thorwebdev:
I switched course and decided to not use the lock but rather rely on a combination of

a) random timeout offset: https://github.com/supabase/gotrue-js/pull/274/files#diff-3522461172efd6058d6b8da62fc2d30d8b524d2b64894ea2c67218c52f7fdff5R716
b) check if the token is still valid within the margin and in that case not refresh: https://github.com/supabase/gotrue-js/pull/274/files#diff-3522461172efd6058d6b8da62fc2d30d8b524d2b64894ea2c67218c52f7fdff5R683

It does limit the chance of the race condition happening, but it does feel a bit hacky... Would appreciate your thoughts!

GaryAustin1:
> I switched course and decided to not use the lock but rather rely on a combination of
> 
> a) random timeout offset: https://github.com/supabase/gotrue-js/pull/274/files#diff-3522461172efd6058d6b8da62fc2d30d8b524d2b64894ea2c67218c52f7fdff5R716 b) check if the token is still valid within the margin and in that case not refresh: https://github.com/supabase/gotrue-js/pull/274/files#diff-3522461172efd6058d6b8da62fc2d30d8b524d2b64894ea2c67218c52f7fdff5R683
> 
> It does limit the chance of the race condition happening, but it does feel a bit hacky... Would appreciate your thoughts!

Upfront I'll say I'm lost on where this is as far as the "final" code as I'm new to github tracking... 

I'm out of time to keep looking tonight, but my scan of what I think the latest GoTrueClient.ts has dropped the visibility event handler.  If understand correctly this PR was to fix some race condition, but then several issues with token expiring while tab is not in focus or in sleep mode got "pulled in".  The visibility handler code was a key part of that possible solution.  I think that is gone now, but as I said, this is a maze for me at the moment...

My view is that the Supabase client should take care of refreshing the jwt token in all cases where the refresh token is valid without user dev interaction.  This can be done right before every authorization header set up in supabase.js, on "token expired" error with a retry on the actual database request, or with visibility handler to deal with dead timer when device MAY have been sleeping (you can not assume "hidden" state means timers have stopped). 

IMO the refresh token timer should be an optimistic event to save time on an actual database call or realtime status, BUT should NOT be the final solution.  Supabase.js and realtime.js should on their own be responsible for verifying token about to expire and refresh it or deal with error gracefully (not easy for realtime).  

At a minimum the visibility handler is needed to deal with dead timer on mobile (and apparently other situations) when the tab does not have focus or device is asleep and token expires.

Once again sorry, it is late but wanted to to throw this out.

thorwebdev:
Thanks, Gary, I’m also not happy with this PR. Agreed, I think there’s a couple of things we need to fix

1. The race condition needs to be handled server-side. There is just too many edge cases, including network request issues, that just can’t be handled on the client (cc @kangmingtay)
2. The visibility handler to recover the session when the window becomes visible.
3. The retries for failed refresh token requests

thorwebdev:
1. is being addressed server-side here: https://github.com/supabase/gotrue/pull/466
2. & 3. are being addressed here: https://github.com/supabase/gotrue-js/pull/278

PR #278, opened 2022-05-04
fix: token refresh retry offline + recover on visible

### LINKED ISSUE #226: Token refresh fails silently when offline and never recovers

### LINKED ISSUE #272: Token refresh time is likely set too "short" if less than 60 seconds left before it expires.
# Bug report

In looking into another issue I found this code in gotrue.js:
```
  if (expiresAt) {
      const expiresIn = expiresAt - timeNow
      const refreshDurationBeforeExpires = expiresIn > 60 ? 60 : 0.5
      this._startAutoRefreshToken((expiresIn - refreshDurationBeforeExpires) * 1000)
```
This code was modified in June 21 here:  https://github.com/supabase/gotrue-js/pull/101 to deal with jwt expire set to less than 60 seconds.

This comment does not seem valid:
"0.5s is chosen to support JWT expiry of 1s, even though that should be an edge case. That should usually still be enough time to reach the server before expiry."

image

This is from Texas to us-east instance and the request are slightly over to slightly less than 500msec in just one random sample today for CORS with preflight requests.

If I login the client with less than 61 seconds on my token, refresh will be set to occur 1 (for 61) to .5 seconds before the token expires.  If I issue a request right as the token is getting ready to expire, we are at the mercy of https response getting a new token back into local storage within .5 seconds or there is a window of invalid token.

I agree it is a small window, and would appear to be a random error in the scheme of things.
Edit: Also if a realtime subscription is running it could lose subscription and have to reconnect if heart beat occurs in the window

## To Reproduce

This is a concern at this point based on looking at the code and token refresh times.

## Expected behavior

Token should be refreshed with plenty of time for round trip of https requests.  This may mean it is unrealistic to support token expiration setting of 5 (or some number) or less.

thorwebdev:

In combination with https://github.com/supabase/gotrue/pull/466 this PR should resolve the most critical token refresh issues,

- Fixes #226 
- Fixes https://github.com/supabase/supabase/issues/6464
- Maybe Fixes #272 

cc @GaryAustin1


src/GoTrueClient.ts
```diff
@@ -643,11 +650,19 @@ export default class GoTrueClient {
       const { currentSession, expiresAt } = data
       const timeNow = Math.round(Date.now() / 1000)
 
-      if (expiresAt < timeNow) {
+      if (expiresAt < timeNow + EXPIRY_MARGIN) {
         if (this.autoRefreshToken && currentSession.refresh_token) {
           const { error } = await this._callRefreshToken(currentSession.refresh_token)
           if (error) {
             console.log(error.message)
+            if (error.message === 'Failed to fetch') {
```
inian, line 658: There is a better way to find out if a fetch failed due to a network error - It throws. If we can pass that info all the way back up that would be a more robust way of checking this. For example, Safari seems to have "Load failed" as the error message if a network request fails

inian: Another way is to trigger the request when the browser comes online via https://developer.mozilla.org/en-US/docs/Web/API/Window/online_event similar to how we do the visibility change

inian (changes requested):

src/GoTrueClient.ts
GaryAustin1: Online event is only good for the devices network connection though.   If the connection to the server is down anywhere (or the server is down/very slow) then you still need a way of retrying.

src/GoTrueClient.ts
```diff
@@ -703,7 +718,7 @@ export default class GoTrueClient {
     if (expiresAt) {
       const timeNow = Math.round(Date.now() / 1000)
       const expiresIn = expiresAt - timeNow
-      const refreshDurationBeforeExpires = expiresIn > 60 ? 60 : 0.5
+      const refreshDurationBeforeExpires = expiresIn > EXPIRY_MARGIN ? EXPIRY_MARGIN : 0.5
       this._startAutoRefreshToken((expiresIn - refreshDurationBeforeExpires) * 1000)
     }
```
GaryAustin1, line 724: I have an open issue that the first part addresses the .5 seconds as not enough time to complete refresh.
https://github.com/supabase/gotrue-js/issues/272

I have a trace showing failure to do preflight and refresh request within 500msec in that issue (even though who ever came up with it felt it was enough time)

So if it is possible for the time left to expiration is EXPIRY_MARGIN + .5 or 1 or less, the token will expire for a brief amount of time if code can run and use the token before this task is finished (which I have not traced thru entirely).

I also think the rounding error mentioned in that still exists here and in by cutting it so close with the .5 as Date.now() could get rounded up adding up to .5 seconds you don't have if on the margin.

src/GoTrueClient.ts
thorwebdev: I've switched to a NETWORK_FAILURE constant so this works across browser and retries for any rejected fetches.

src/GoTrueClient.ts
thorwebdev: We've made the following decisions:

1. We will default JWT expiry to 15mins in the future (once we've verified all the token refresh issues have been resolved)
2. We will allow a minimum JWT expiry of 30s.
3. We will default the expiry margin to 10s.
4. In case of network failure we will retry every second. 

So if you set JWT to the smallest possible value (30s), the client library will refresh it every 20s.

src/lib/fetch.ts
```diff
@@ -52,7 +39,12 @@ async function _handleRequest(
         return result.json()
       })
       .then((data) => resolve(data))
-      .catch((error) => handleError(error, reject))
+      .catch(() =>
+        reject({
+          message: NETWORK_FAILURE,
```
inian, line 44: Since we are throwing even if `result.ok` is false, lets have a generic message like "Request Failed"? Cos this will get triggered both for network errors and server errors like timeouts etc

src/GoTrueClient.ts
```diff
@@ -634,20 +636,24 @@ export default class GoTrueClient {
...
-      }
-
-      const data = JSON.parse(json)
+      const data = await getItemAsync(this.localStorage, STORAGE_KEY)
+      if (!data) return null
       const { currentSession, expiresAt } = data
       const timeNow = Math.round(Date.now() / 1000)
 
-      if (expiresAt < timeNow) {
+      if (expiresAt < timeNow + EXPIRY_MARGIN) {
```
inian, line 644: a counter and stop pinging gotrue or stop pinging after expiry time is reached cos gotrue is not going to refresh that token anyway

inian (approved):

src/GoTrueClient.ts
thorwebdev: Added exponential backoff, starting with 200ms and then exponentially backing off for 10 retries

GaryAustin1:
@thorwebdev I can confirm the new supabase.js 1.35.3  refreshes the token when coming back from sleep in my test suite from https://github.com/supabase/supabase/issues/6464

I do notice one issue though so far.  My test code has a visibility handler to generate a select when tab goes visible.  That fails with JWT expired.  I put a delay of 5 seconds in to the select and it works.  I put a delay of 100msec in and it did not work (still get JWT expired)

So at a minimum anyone using a visibility handler needs to check for token refreshed before executing a function.

Worse case is though that ANY CODE THAT RUNS before the refresh finishes will error. 

I will try and work on a test for this not using a visibility handler in my code, but I am starting a 10 day road trip in the morning, so may not be able to get to it quickly.

Select is in doStuff.
```
        if (document.visibilityState === "visible" ) {
            //await checkSession()
        setTimeout(function(){ doStuff(); }, 100);
       }
```
image

GaryAustin1:
@thorwebdev If there is a better place for me to document findings on this pull request let me know.

I've not been able to figure out how to test the issue above without a visibility handler yet.

The following trace bothers me, and although I guess not a bug is certainly different in how gotrue.js will run.
I expected the new code to not do anything if the token was not within expire time, but it generates a signin event on every hidden to visible transition, which on a desktop tab could be 100's of times while the app is in use in an hour.   On the desktop the refresh also normally occurs in background, so you should never have a stale token.  Why even issue a signIn in event in the visibility handler, if token is good to go?  Note: the visibility handler messaged below is mine, not gotrue's, but occur on same event.
 
tokenrefreshdesktop

PR #286, opened 2022-05-17
Add query params

### LINKED ISSUE #279: WorkOS Authentication: Pass in query params to signIn method
# Bug report

GoTrue-js library currently does not allow for query params to be passed to the `signIn` method. This impedes usage of the WorkOS provider.

## To Reproduce

SignIn with WorkOS as provider

## Expected behavior

User should be able to pass query params and sign in with WorkOS as provider.

## Screenshots

NA

J0:

Fix #279 

Can't pass in query Params 

## What is the new behavior?

Can pass in query params

TODOs:
- Figure out whether to handle OIDC endpoint
- amend queryParams and decide whether to pass in  additional values as array, string, orother

### DIFF
--- a/src/GoTrueApi.ts
+++ b/src/GoTrueApi.ts
@@ -67,6 +67,7 @@ export default class GoTrueApi {
     options: {
       redirectTo?: string
       scopes?: string
+      queryParams?: string
     }
   ) {
     const urlParams: string[] = [`provider=${encodeURIComponent(provider)}`]
@@ -75,6 +76,9 @@ export default class GoTrueApi {
     }
     if (options?.scopes) {
       urlParams.push(`scopes=${encodeURIComponent(options.scopes)}`)
+    }
+    if(options?.queryParams) {
+      urlParams.push(options.queryParams)
     }
     return `${this.url}/authorize?${urlParams.join('&')}`
   }
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -194,6 +194,7 @@ export default class GoTrueClient {
       shouldCreateUser?: boolean
       scopes?: string
       captchaToken?: string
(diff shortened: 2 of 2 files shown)


src/GoTrueApi.ts
```diff
@@ -76,6 +77,11 @@ export default class GoTrueApi {
     if (options?.scopes) {
       urlParams.push(`scopes=${encodeURIComponent(options.scopes)}`)
     }
+    if(options?.queryParams) {
+      for(const [param, value] of Object.entries(options.queryParams)) {
+        urlParams.push(`${encodeURIComponent(param)}=${encodeURIComponent(value)}`)
```
inian, line 82: Consider using URLSearchParams here. Since realtime uses it [here](https://github.com/supabase/realtime-js/blob/05d098c228d4303151d58448bf3e05cfb1c77225/src/RealtimeClient.ts#L451), I don't think compatibility across different environments should be an issue

inian (approved):

src/GoTrueApi.ts
J0: Makes sense, will use URLSearchParams

liaujianjie:
Just to clarify, the usage with this change will look like this right?

```tsx
supabaseClient.auth.signIn(
  { email: "john@acme-inc.com", provider: "workos" },
  undefined,
  { queryParams: { connection: "conn_xxxxxxxxx" } },
)
```

J0:
Oh actually, I think it will look more like this:

```js
supabaseClient.auth.signIn(
  { provider: "workos" },
  { queryParams: { connection: "conn_xxxxxxxxx", organization: "...", workos_provider: "...." } },
)
```

Have yet to test and document the flow but I'll include the usage instructions in the docs

Issue #255, opened 2022-03-22
Broken link in README


whalesync-ryder:
The top-level readme has a broken link:

https://github.com/supabase/gotrue-js#docs

There's a 404 at "Using gotrue-js: https://supabase.io/docs/gotrue/client/initializing"

J0:
Thanks for reporting this issue!

PR #287, opened 2022-05-23
Fix broken links

inian:
Also points to Supabase Gotrue since the client supports features that the original library doesnt


J0:
Fixes #255

J0 (approved):

PR #293, opened 2022-06-06
build(release-next): sets up the next branch as an npm prerelease

alaister:

Sets up the next branch as an npm prerelease

There is currently no prerelease pipeline.

## What is the new behavior?

Pushes to the next branch will be included as an npm prerelease under the `next` tag.

## Additional context

Prereleases will look like:
Screen Shot 2022-06-06 at 4 41 22 pm

If this works as intended and we're happy I'll also add this to `supabase-js`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - master
+      - next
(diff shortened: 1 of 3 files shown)


thorwebdev (approved):
LGTM

PR #298, opened 2022-06-07
fix: ignore no-extra-semi

alaister:

Build fix

Build fails due to no extra semi rule

## What is the new behavior?

Ignore the rule where needed


thorwebdev (approved):
LGTM

PR #299, opened 2022-06-07
fix: getSession() read from localStorage every time

alaister:

Bug fix

When the session is expired the new `getSession()` handler does not refresh it as `_recoverSession()` does not set `this.currentSession` if it has expired, so `getSession()` thinks the user is logged out.

## What is the new behaviour?

`getSession()` reads from local storage (or in-memory storage) on every call, rather than relying on `this.currentSession`.
This also sets us up for the future, when we will be able to remove all the cross tab synchronisation logic.

## Additional context

This PR also deprecates a bunch of stuff, most notably `setAuth()`. As we're moving away from `this.currentSession`, `setAuth()` will not be needed to synchronise sessions across gotrue instances, as it can now be done via storage.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -53,12 +53,20 @@ export default class GoTrueClient {
   api: GoTrueApi
   /**
    * The currently logged in user or null.
+   * @deprecated use `getUser()` instead
    */
   protected currentUser: User | null
   /**
    * The session object for the currently logged in user or null.
+   * @deprecated use `getSession()` instead
    */
   protected currentSession: Session | null
+
+  /**
+   * The session object for the currently logged in user or null.
+   * Only used if persistSession is false.
+   */
+  protected inMemorySession: Session | null
 
   protected autoRefreshToken: boolean
   protected persistSession: boolean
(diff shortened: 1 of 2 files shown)


soedirgo (approved):
I've approved for now, but tbh not familiar with gotrue-js - might want a second pair of eyes for this, either @kangmingtay or @thorwebdev?

alaister:
Thanks @soedirgo!

Since this is going into `next` I'm going to merge this in now so I can test in some of my real apps.

Another look over would very helpful @kangmingtay and @thorwebdev.
@inian your input is very welcome here too – I remember you mentioned you liked the idea of reading from `localStorage` every time.

src/GoTrueClient.ts
```diff
@@ -478,6 +491,7 @@ export default class GoTrueClient {
   /**
    * Overrides the JWT on the current client. The JWT will then be sent in all subsequent network requests.
    * @param access_token a jwt access token
+   * @deprecated you can use `getSession()` instead as it reads from storage every time (always fresh)
```
kangmingtay, line 494: hmm just wondering, why are we deprecating `setAuth()` here? wouldn't this be useful for a user using gotrue-js on it's own and they need to set a custom JWT?

alaister: Good point. Fixed in #300

We should also consider actually persisting the sessions in `setAuth()`, although it will mean we'll have to make it async.

PR #300, opened 2022-06-07
fix: destructure currentSession from persistedSession

alaister:

Bug fix

`_persistSession()` makes a new object in `localStorage` for some reason so when reading it back out in `getSession()` needs to pull out the `currentSession` field.

## What is the new behaviour?

Pulls out the `currentSession` field from the `persistedSession`


thorwebdev (approved):
LGTM

PR #301, opened 2022-06-08
feature): [gotrue-js] Refactor return and error types

dthyresson:

This PR refactors the return types for the `gotrue-js` client to match the pattern used by `storage-js`.

See: https://github.com/supabase/storage-js/pull/60

> There are some type errors, and functions return everything with | null, making it difficult for end-users to work with.

> Gotrue-js types are hard to work with and would benefit from union types everywhere. See what we’ve done in storage-js for inspiration 

See: https://github.com/supabase/storage-js/pull/60

## What is the new behavior?

* Functions now return type unions, which are more correct and "easier to work with".
* Updates tests
* Adds new error types
* Updates cross-fetch version to match storage

## Additional context

Tests pass.

```terminal
gotrue-js dt-error-types % npm run test:suite

> @supabase/gotrue-js@0.2.0 test:suite
> jest --runInBand

 PASS  test/GoTrueClient.test.ts
 PASS  test/GoTrueApi.test.ts
 PASS  test/lib/utils.test.ts

Test Suites: 3 passed, 3 total
Tests:       2 skipped, 63 passed, 65 total
Snapshots:   0 total
Time:        7.159 s, estimated 8 s
Ran all test suites.
```

### DIFF
--- a/src/lib/errors.ts
+++ b/src/lib/errors.ts
@@ -1,4 +1,4 @@
-export class AuthApiError extends Error {
+export class AuthError extends Error {
   protected __isAuthError = true
 
   constructor(message: string) {
@@ -11,7 +11,7 @@ export function isAuthError(error: unknown): error is AuthError {
   return typeof error === 'object' && error !== null && '__isAuthError' in error
 }
 
-export class AuthError extends AuthApiError {
+export class AuthApiError extends AuthError {
   status: number
 
   constructor(message: string, status: number) {
@@ -29,7 +29,7 @@ export class AuthError extends AuthApiError {
   }
 }
 
-export class AuthUnknownError extends AuthApiError {
+export class AuthUnknownError extends AuthError {
   originalError: unknown
 
   constructor(message: string, originalError: unknown) {
--- a/src/lib/errors.ts
+++ b/src/lib/errors.ts
@@ -3,7 +3,7 @@ export class AuthError extends Error {
 
   constructor(message: string) {
     super(message)
-    this.name = 'AuthApiError'
+    this.name = 'AuthError'
   }
 }
 
@@ -16,7 +16,7 @@ export class AuthApiError extends AuthError {
 
   constructor(message: string, status: number) {
     super(message)
-    this.name = 'AuthError'
+    this.name = 'AuthApiError'
     this.status = status
   }
 
--- a/src/lib/errors.ts
+++ b/src/lib/errors.ts
@@ -1,0 +1,40 @@
+export class AuthApiError extends Error {
+  protected __isAuthError = true
+
+  constructor(message: string) {
+    super(message)
+    this.name = 'AuthApiError'
+  }
+}
+
+export function isAuthError(error: unknown): error is AuthError {
+  return typeof error === 'object' && error !== null && '__isAuthError' in error
+}
+
+export class AuthError extends AuthApiError {
+  status: number
+
+  constructor(message: string, status: number) {
+    super(message)
+    this.name = 'AuthError'
+    this.status = status
+  }
+
+  toJSON() {
+    return {
+      name: this.name,
+      message: this.message,
+      status: this.status,
+    }
+  }
+}
+
+export class AuthUnknownError extends AuthApiError {
+  originalError: unknown
+
+  constructor(message: string, originalError: unknown) {
+    super(message)
+    this.name = 'AuthUnknownError'
+    this.originalError = originalError
+  }
+}
--- a/src/lib/fetch.ts
+++ b/src/lib/fetch.ts
@@ -1,4 +1,4 @@
-import { AuthError, AuthUnknownError } from './errors'
+import { AuthApiError, AuthUnknownError } from './errors'
 import { resolveResponse } from './helpers'
 
 export type Fetch = typeof fetch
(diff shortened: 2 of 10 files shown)


dthyresson:
@kangmingtay -- 

* This is a Draft WIP PR where tests pass locally, but CI fails. Fixing.

1. In some places I left the `| null` like
```ts
Promise<
    | {
        session: Session | null
        user: User | null
        provider?: Provider
        url?: string | null
        error: null
      }
    | {
        session: Session | null
        user: User | null
        provider?: Provider
        url?: string | null
        error: AuthError
      }
  >
```

But perhaps I should enumerate the possibilities (I did so when there were fewer nullable cases)?

2. There are many repeated, common return types. Should I create types from these so less copy paste and easier to make changes moving forward?

For example:

```ts
Promise<
    | {
        data: User
        error: null
      }
    | {
        data: Session
        error: null
      }
    | { data: null; error: AuthError }
  >
```

I could make this a `UserOrSession` or `DataAsUserOrSession` something along those lines and reduce repeated code.

dthyresson:
Also what changes to the documentation need to be made? Any new error handling?

dthyresson:
Also, should the release and package be updated to 2?

```
{
  "name": "@supabase/gotrue-js",
  "version": "0.2.0",
```

and RELEASES?

```
# Releases

Releases are handled by Semantic release. This document is for forcing and documenting any non-code changes.

### v1.27 

- Fix: https://github.com/supabase/gotrue-js/pull/184

### v1.12.7
```

src/lib/fetch.ts
```diff
@@ -9,85 +10,99 @@ export interface FetchOptions {
...
+  method: RequestMethodType,
+  options?: FetchOptions,
+  parameters?: FetchParameters,
+  body?: object
+) => {
   const params: { [k: string]: any } = { method, headers: options?.headers || {} }
 
   if (method === 'GET') {
     return params
   }
```
kangmingtay, line 44: hmm why don't we return `parameters` if it's a get request?

src/lib/fetch.ts
```diff
@@ -9,85 +10,99 @@ export interface FetchOptions {
   noResolveJson?: boolean
 }
 
-export type RequestMethodType = 'GET' | 'POST' | 'PUT' | 'PATCH' | 'DELETE'
+export interface FetchParameters {
+  signal?: AbortSignal
+}
```
kangmingtay, line 15: just wondering what is this for and why do we add it to every fetch request?

kangmingtay (changes requested):

src/lib/fetch.ts
alaister: It's a bit like a context in Go - allows us to cancel a request. We have them in `postgrest-js` so I think it makes sense to have them here too

https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal

kangmingtay:
> 1. In some places I left the `| null` like
> 
> ```ts
> Promise<
>     | {
>         session: Session | null
>         user: User | null
>         provider?: Provider
>         url?: string | null
>         error: null
>       }
>     | {
>         session: Session | null
>         user: User | null
>         provider?: Provider
>         url?: string | null
>         error: AuthError
>       }
>   >
> ```
> 
> But perhaps I should enumerate the possibilities (I did so when there were fewer nullable cases)?
> 

Would it make sense to just have everything in the `Session` type since it also contains the `User` type? We could also add the `Provider` type inside `Session` so that there are fewer possibilities to consider?

kangmingtay:
> 2. There are many repeated, common return types. Should I create types from these so less copy paste and easier to make changes moving forward?
> 
> For example:
> 
> ```ts
> Promise<
>     | {
>         data: User
>         error: null
>       }
>     | {
>         data: Session
>         error: null
>       }
>     | { data: null; error: AuthError }
>   >
> ```
> 
> I could make this a `UserOrSession` or `DataAsUserOrSession` something along those lines and reduce repeated code.

Looking at this, it makes sense to remove the type check for `User` since `Session` has a `User` type in it. Also, if the `User` is null, then i guess the `Session` would be null too? 

@alaister also has some thoughts on why he prefers not to have custom types for these

dthyresson:
> Looking at this, it makes sense to remove the type check for `User` since `Session` has a `User` type in it. Also, if the `User` is null, then i guess the `Session` would be null too?

One of the more confusing aspects of this api is the use of user vs session vs data.

Sometimes data is a user, sometimes a session, sometimes it returns user and sometimes session.

Is this the time to refactor to be consistent and *not* use “data” when a user or session is returned?

src/lib/errors.ts
```diff
@@ -0,0 +1,40 @@
...
+    super(message)
+    this.name = 'AuthApiError'
+  }
+}
+
+export function isAuthError(error: unknown): error is AuthError {
+  return typeof error === 'object' && error !== null && '__isAuthError' in error
+}
+
+export class AuthError extends AuthApiError {
```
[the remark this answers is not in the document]
dthyresson: I went back and for on this naming -- I just wasn't sure what to expose to the user -- `AuthError` (base) or `AuthApiError`(just by client or api aka admin).

* AuthError
* AuthApiError < AuthError

or 

* AuthNNNNError (some base error)
* AuthError < AuthNNNNError

Another option is to have separate error types:

* AuthError (some base error)
* AuthClientError < AuthError -> for client 
* AuthApiError -> for admin api

And then is Client and Api also confusing? :)

src/lib/fetch.ts
dthyresson: Seems to be in the prior implementation as well. I used the fetch code from the storage client lib.

src/lib/errors.ts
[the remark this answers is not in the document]
kangmingtay: hmm i think regardless of whether it's a client or admin method, it should be an `AuthError`? Then any internal gotrue api errors will be classified as `AuthApiError`? Wdyt?

kangmingtay (approved):

PR #303, opened 2022-06-16
fix: allow custom storage key in localstorage

kangmingtay:
* Allows passing of a custom storage key in the constructor
* Fixes https://github.com/supabase/supabase-js/issues/332 

## To-Do
* Supabase-js needs to pass a custom storage key when instantiating the `GoTrueClient` class


thorwebdev (approved):
One nit, PTAL

PR #304, opened 2022-06-23
fix: refactor sign in methods

kangmingtay:
* Split up the `signIn` method into their respective use cases (e.g. password sign-ins, passwordless sign-ins, oauth sign-ins)
* Removed `signIn` with refreshToken because the `getSession()` method already handles that and it seems incorrect to "sign-in with a refresh token" when the user is just refreshing the session
* Removed sign in with oidc capabilities since we need to rework the implementation in gotrue first

## To-Dos
- [x] Get tests to pass
- [x] Captcha token field in gotrue should be renamed from `hcaptcha_token` to `captcha_token` for generalisability 
- [x] ~Add a `is_new_user` field in the response for the passwordless and oauth endpoints in gotrue so that the developer can tell if it was a new user that signed up or not~


src/GoTrueClient.ts
```diff
@@ -216,78 +219,82 @@ export default class GoTrueClient {
...
-        // currentSession and currentUser will be updated to latest on _callRefreshToken using the passed refreshToken
-        const { error } = await this._callRefreshToken(refreshToken)
-        if (error) throw error
-
-        return {
-          user: this.currentUser,
-          session: this.currentSession,
-          error: null,
-        }
+      throw new Error(`You must provide either an email or phone number and a password.`)
```
silentworks, line 258: Since this is for v2 it would be good to have proper error codes and responses. And to keep these consistent between libraries along with a glossy explaining what each error means and what could be the cause.

src/GoTrueClient.ts
```diff
@@ -216,78 +219,82 @@ export default class GoTrueClient {
...
+   * @param provider One of the providers supported by GoTrue.
+   * @param oidc See OpenIdCredentials type specification
+   * @param redirectTo A URL to send the user to after they are confirmed (OAuth logins only).
+   * @param scopes A space-separated list of scopes granted to the OAuth application.
+   * @param queryParams An object of query params 
+   */
+  async signInWithOAuth(
+    { provider, oidc, options = {} }: SignInWithOAuthCredentials,
+  ): Promise<
+    | {
```
silentworks, line 280: Can we create proper interfaces for the responses instead of writing them out like this inside of each method?

src/GoTrueClient.ts
```diff
@@ -216,78 +219,82 @@ export default class GoTrueClient {
...
+        session: Session | null
+        user: User | null
+        provider?: Provider
+        url?: string | null
+        error: null
+      }
+    | {
+        session: Session | null
```
inian, line 288: why will a session or user be returned in case of AuthError?

inian (changes requested):
The types can be made tighter as part of this PR

test/GoTrueClient.test.ts
```diff
@@ -322,7 +322,7 @@ describe('GoTrueClient', () => {
     expect(initialSession).not.toBeNull()
 
     const refreshToken = initialSession?.refresh_token
-    const { error, user, session } = await authWithSession.signIn({ refreshToken })
+    const { error, user, session } = await authWithSession.refreshSession()
```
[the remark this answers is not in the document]
pixtron: Did you consider [this use case](https://supabase.com/docs/reference/javascript/auth-signin#sign-in-using-a-refresh-token-eg-in-react-native)?

src/lib/types.ts
```diff
@@ -181,6 +181,36 @@ export interface UserCredentials {
...
+export interface SignInWithOAuthCredentials {
+  provider?: Provider
+  oidc?: OpenIDConnectCredentials
+  options?: SignInWithOAuthOptions
+}
+export interface SignInWithOAuthOptions {
+  redirectTo?: string
+  scopes?: string
+  queryParams?: { [key: string]: string }
+}
```
alaister, line 213: We can use union types here for more descriptive APIs.

```javascript
export type SignInWithPasswordOptions = {
  captchaToken?: string
}

export type SignInWithPasswordCredentials = {
  email?: string
  password: string
  options?: SignInWithPasswordOptions
} | {
  phone?: string
  password: string
  options?: SignInWithPasswordOptions
}

export type SignInWithPasswordlessOptions = {
  // The redirect url embedded in the email link
  emailRedirectTo?: string
  shouldCreateUser?: boolean
  captchaToken?: string
}

export type SignInWithPasswordlessCredentials = {
  email?: string
  options?: SignInWithPasswordlessOptions
} | {
  email?: string
  options?: SignInWithPasswordlessOptions
}

export type SignInWithOAuthCredentials = {
  provider?: Provider
  oidc?: OpenIDConnectCredentials
  options?: SignInWithOAuthOptions
}

export type SignInWithOAuthOptions = {
  redirectTo?: string
  scopes?: string
  queryParams?: { [key: string]: string }
}
```

src/GoTrueClient.ts
```diff
@@ -216,78 +219,82 @@ export default class GoTrueClient {
...
+   * @param scopes A space-separated list of scopes granted to the OAuth application.
+   * @param queryParams An object of query params 
+   */
+  async signInWithOAuth(
+    { provider, oidc, options = {} }: SignInWithOAuthCredentials,
+  ): Promise<
+    | {
+        session: Session | null
+        user: User | null
+        provider?: Provider
```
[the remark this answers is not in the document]
kangmingtay: yup, i've also decided to remove the oidc flow from v2 till we've reworked the implementation on gotrue

src/GoTrueClient.ts
kangmingtay: added an [`AuthResponse`](https://github.com/supabase/gotrue-js/pull/304/files#diff-b8b77f5be18cf56dca425b3a5b8e9d2e754dd37fe0e3612b95cd4e9bccda08a5R29-R37) type

test/GoTrueClient.test.ts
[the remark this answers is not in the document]
kangmingtay: thanks, i've updated [`refreshSession`](https://github.com/supabase/gotrue-js/pull/304/files#diff-3522461172efd6058d6b8da62fc2d30d8b524d2b64894ea2c67218c52f7fdff5R473) to take in an optional refresh token to handle this case

src/GoTrueClient.ts
kangmingtay: resolved in https://github.com/supabase/gotrue-js/pull/341

J0 (approved):
Looks good to me, Thanks! 💚 Should make things clearer + types clear up user questions about when they can use certain options(e.g. redirectTo)

inian (approved):

vbylen:
hi @kangmingtay, I can see you added the types for `signInWithPasswordless()`, but not the implementation. Is that something you plan on adding later? 

Great work by the way.

kangmingtay:
hey @vbylen, you should use `signInWithOtp()` instead, it uses the `SignInWithPasswordlessCredentials` and is the method you want to use for passwordless sign-ins. We won't be adding a `signInWithPasswordless()` method for now.

vbylen:
> hey @vbylen, you should use `signInWithOtp()` instead, it uses the `SignInWithPasswordlessCredentials` and is the method you want to use for passwordless sign-ins. We won't be adding a `signInWithPasswordless()` method for now.

Thanks for clarifying!

GaryAustin1:
I’m just throwing this here as I happened to be here, the new release notes for v2 use passwordless and not OTP. Customers are encountering.  I’ll try and get a issue generated soon.

kangmingtay:
thanks for catching this @GaryAustin1! i've made a PR to amend it already

FezVrasta:
@kangmingtay the refresh token login is used by the React Native example to work with Magic Links, what's the right way to do this now that you removed this functionality? Thanks!

PR #306, opened 2022-06-27
fix: import cross-fetch conditionally

soedirgo:


alaister (approved):
Thanks, Bobbie! lgtm!

PR #309, opened 2022-07-06
Possible change of error message when getting session fails

### LINKED ISSUE #240: Annoy error message without a error.
# Chore

## Describe the chore

https://github.com/supabase/gotrue-js/blob/e440a90099b5c138ac841348e1f1bb5b02f58608/src/GoTrueClient.ts#L109

the library is reporting an error and it should be a warn because it doesn't stop the execution of the app.

J0:

Changes  the console log on error in getting session from URL from error -> to warn in order to fix  #240 . Seems more a DX thing so looping in Thor and Silentworks.

## What is the new behavior?

## Additional context

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -112,7 +112,7 @@ export default class GoTrueClient {
       // Handle the OAuth redirect
       this.getSessionFromUrl({ storeSession: true }).then(({ error }) => {
         if (error) {
-          console.error('Error getting session from URL.', error)
+          console.warn('Error getting session from URL.', error)
         }
       })
     }
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -112,7 +112,7 @@ export default class GoTrueClient {
       // Handle the OAuth redirect
       this.getSessionFromUrl({ storeSession: true }).then(({ error }) => {
         if (error) {
-          console.warn('Error getting session from URL.', error)
+          throw new Error('Error getting session from URL.')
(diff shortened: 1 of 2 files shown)


kangmingtay (commented):
should we just throw the error instead? @thorwebdev @silentworks

silentworks (approved):

silentworks:
A warn seems fair to me in this context.

kangmingtay:
Actually, hang on i think we should throw this error instead based on this [issue](https://github.com/supabase/gotrue/issues/536)

pixtron:
@kangmingtay it was a good intention to throw the error. the thing is: the constructor is sync and the error is thrown when the async getSessionFromUrl resolves. At this point the constructor already returned and therefore there is no chance to catch that error. Probably best is to revert back to logging it?. There is a [workaround](https://github.com/supabase/gotrue/issues/536#issuecomment-1200156629) how catch these errors. Check this simplified example:

```js
class Client {
  constructor() {
    this.getSessionFromUrl().then(({error}) => {
      if (error) {
        throw error;
      }
    });
  }
  
  async getSessionFromUrl() {
    try {
      await new Promise(resolve => setTimeout(resolve, 1000));
      throw new Error('No expires_in detected.');
      
      return { session: { access_token: 'xxx' }, error: null}
    } catch(e) {
      return { session: null, error: e };
    }
  }
}

try { 
  const client = new Client();
} catch(error) {
  console.log('can\'t catch it here', error);
}
```

PR #327, opened 2022-07-19
Fix: only save session when persist session is true

alaister:

Bug fix

If the user has `persistSession` set to `false`, `gotrue-js` will still load in an existing session from `localStorage` if there is one.

## What is the new behaviour?

`gotrue-js` will only save the session from `localStorage` if the user has `persistSession` set to `true`.


J0 (approved):
Good catch!

J0:
Oh yeah just in case:

Build is failing because of an update to GoTrue -- changes similar to  https://github.com/supabase/gotrue-js/commit/9a0db1456d2c1f258995c8901063b8303dc987ef
need to be put in.

PR #337, opened 2022-07-29
Reduce some duplication in GoTrueApi

karlseguin:
Refactor / DRY

A lot of copy/paste code inside GoTrueAPI

## What is the new behavior?
Refactored repetitive code inside of functions so that they can be re-used.

## Additional context
!!
**I am not a typescript/javascript developer.** 
!!

Also, I think these changes are pretty subjective, so...I wouldn't have my feelings hurt to have it rejected :)

Repeating yourself is often useful, but I think having 5 places do:
```
const session = { ...data }
if (session.expires_in) session.expires_at = expiresAt(data.expires_in)
return { data: session, error: null }
```

Isn't great. As a reader, I need to read all 5 closely to figure out if there's a [subtle] difference between them or not.Whereas, having them all use that same function makes the intent "these are all the same" much more obvious

### DIFF
--- a/src/GoTrueApi.ts
+++ b/src/GoTrueApi.ts
@@ -1,4 +1,4 @@
-import { Fetch, get, post, put, remove } from './lib/fetch'
+import { Fetch, FetchOptions, RequestMethodType, request } from './lib/fetch'
 import {
   Session,
   Provider,
@@ -12,6 +12,15 @@ import {
 import { COOKIE_OPTIONS } from './lib/constants'
 import { setCookies, getCookieString } from './lib/cookies'
 import { expiresAt, resolveFetch } from './lib/helpers'
+
+interface ClientFetchOptions extends FetchOptions {
+  jwt?: string
+  body?: object
+  redirectTo?: string
+  xerr?: (err: ApiError) => any
+  xform?: (data: any) => any
+  query?: { [key: string]: string }
+}
 
 import type { ApiError } from './lib/types'
 export default class GoTrueApi {
@@ -41,17 +50,6 @@ export default class GoTrueApi {
     this.fetch = resolveFetch(fetch)
   }
 
-  /**
-   * Create a temporary object with all configured headers and
-   * adds the Authorization token to be used on request methods
-   * @param jwt A valid, logged-in JWT.
-   */
-  private _createRequestHeaders(jwt: string) {
-    const headers = { ...this.headers }
-    headers['Authorization'] = `Bearer ${jwt}`
-    return headers
-  }
-
   private cookieName() {
     return this.cookieOptions.name ?? ''
   }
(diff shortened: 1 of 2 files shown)


src/GoTrueApi.ts
```diff
@@ -70,18 +68,17 @@ export default class GoTrueApi {
...
     if (options?.scopes) {
-      urlParams.push(`scopes=${encodeURIComponent(options.scopes)}`)
+      query['scopes'] = options.scopes
     }
-    if (options?.queryParams) {
-      const query = new URLSearchParams(options.queryParams)
-      urlParams.push(`${query}`)
-    }
-    return `${this.url}/authorize?${urlParams.join('&')}`
+    const qs = new URLSearchParams(query).toString()
```
alaister, line 80: This is a small one but seeing as the purpose of this is to clean up code, I like the idea of removing the intermediary `query` object in favour of using the `URLSearchParams` directly. For example:
```javascript
const qs = new URLSearchParams(options?.queryParams).toString()

qs.set('provider', provider)

if (options?.redirectTo) {
  qs.set('redirect_to', options.redirectTo)
}
```

alaister (changes requested):
Hey @karlseguin,
Thanks for doing this – I really like a lot of these changes, and I think it makes `gotrue-js` easier to understand for someone looking at it for the first time.

I have somewhat of a large ask though: Would you mind working off the [next](https://github.com/supabase/gotrue-js/tree/next) branch as there are a lot of changes that conflict with this PR.

There are also some changes related to how errors are handled, which may make your `_userResponse` and `_userResponseErr` functions no longer needed.

Let me know if you have any questions about the new changes in `next`.

Thanks again!

karlseguin:
Hi @alaister 

Is there a reason for having both Api and Client? I see some methods, like `verifyMobileOTP` are only accessible directly from GoTrueApi. Are callers expected to reach into `auth.api.X`?

I feel like a lot of the code can be described as:
```
client_magicLink(email, optionsl) {
   api_magicLink(email, {shouldCreateUser: options.shouldCreateUser})
}

api_magicLink(email, options) {
   fetch({email,  gotrue_meta_security: { captcha_token: options.captchaToken })
}
```
And I wonder why client couldn't call fetch directly and skip the intermediary representation.

I think this is most obvious with errors:
1 - fetch raises an exception
2  - api catches and returns an error
3 - client detects the error and raises an exception
4 - client catches the exception and returns an error

If client called fetch directly, it would be:
1 - fetch raises an exception
2 - client catches and returns an error

Same outcome, but 1 less  if + throw + catch

karlseguin:
I understand now that Api has extra functionality for server-side/service roles. A bit of a shame that it's exposed to through the Client. I'd need to put more thought into this now that I get it.

PR #339, opened 2022-08-01
fix: resolve & reset deferred upon refresh error

### LINKED ISSUE #334: Dangling Promises in _callRefreshToken
# Bug report

## Describe the bug

Upon an error the method [`_callRefreshToken`](https://github.com/supabase/gotrue-js/blob/fb7fae5d763c7173f16aa8c6a8747013d8ba4d24/src/GoTrueClient.ts#L887-L923) does not resolve the `refreshingDeferred`, leading to dangling promises. If an error during a refresh token request occurs the deferred is kept for ever, and all subsequent calls to `_callRefreshToken` will [await  the deferred](https://github.com/supabase/gotrue-js/blob/fb7fae5d763c7173f16aa8c6a8747013d8ba4d24/src/GoTrueClient.ts#L890-L892) for an infinite time as it will never resolve.

## To Reproduce

1. clone the [repro](https://github.com/pixtron/supabase-repro)
2. npm install
3. npm start
4. Set breakpoints on these lines:
	 - [`if (this.refreshingDeferred)`](https://github.com/supabase/gotrue-js/blob/v1.23.0-next.6/src/GoTrueClient.ts#L890)
	 - [`const { session, error } = await this.api.refreshAccessToken(refresh_token)`](https://github.com/supabase/gotrue-js/blob/fb7fae5d763c7173f16aa8c6a8747013d8ba4d24/src/GoTrueClient.ts#L902)
5. Signup with email and password (or login if you already have an account)
6. Disable all network interfaces
7. Wait 2 minutes until token expires and auto refresh is triggered (there will be no retries #333)
8. Enable network again
9. Sign out
10. Sign in again
11. Wait 2 minutes for refresh 

The condition [`if (this.refreshingDeferred)`](https://github.com/supabase/gotrue-js/blob/v1.23.0-next.6/src/GoTrueClient.ts#L890) will be true and `this.refreshingDeferred` from the former refresh request in step 7 is returned and never rejected nor resolved.

## Expected behavior

The deferred should be resolved with the error result and then be reset to null.

## Code sandbox
I created a [sandbox ](https://codesandbox.io/s/current-deferred-m3crnh?file=/src/Client.ts:828-1590) to simulate the behavior.

#### Auth Error 
1. Click two times on the Button "Refresh Auth Error"
2. Wait a second
3. We only get a result for request #1 NOK: we should get a result for req #2 as well
4. Click again on any Button
5. Request #1 is triggered, but never gets a result NOK

#### Generic Error
Same scenario as Auth Error but with Button "Refresh Generic Error", throws the error instead of resolving with an error response. 

#### Success case works as expected
1. Click tow times on the Button "Refresh Success"
2. Wait a second
3. Both requests get resolved with the result of req #1 OK
 
## Suggested Fix

1. Move [`this.refreshingDeferred = null`](https://github.com/supabase/gotrue-js/blob/fb7fae5d763c7173f16aa8c6a8747013d8ba4d24/src/GoTrueClient.ts#L913) into a `finally` block so it gets reset in success and error case.
2. [If the error is an `AuthError`](https://github.com/supabase/gotrue-js/blob/fb7fae5d763c7173f16aa8c6a8747013d8ba4d24/src/GoTrueClient.ts#L917), resolve `this.refreshingDeferred` with the error result in the catch block.
3. [If the error is not an `AuthError`](https://github.com/supabase/gotrue-js/blob/fb7fae5d763c7173f16aa8c6a8747013d8ba4d24/src/GoTrueClient.ts#L921), reject `this.refreshingDeffered` with the thrown error
4. Move [`return await this.refreshingDeferred.promise`](https://github.com/supabase/gotrue-js/blob/fb7fae5d763c7173f16aa8c6a8747013d8ba4d24/src/GoTrueClient.ts#L890-L892) out of the try block and don't await it (`return this.refreshingDeferred.promise`). This is to avoid multiple entries into catch and finally block, when the error is not an `AuthError` and therefore re thrown (see 3).

I created a [sandbox simulation with the suggested fix](https://codesandbox.io/s/fixed-deferred-mghg60?file=/src/Client.ts:828-1933) 

## System information

- OS: NA
- Browser: NA
- Version of supabase-js: NA
- Version of gotrue-js: 1.23.0-next.6
- Version of Node.js: NA

## Additional Context

This regression seems to have been introduced in 1.23.0-next with [#285](https://github.com/supabase/gotrue-js/pull/285/files#diff-3522461172efd6058d6b8da62fc2d30d8b524d2b64894ea2c67218c52f7fdff5R756-R782).

pixtron:

Bug fix

Currently `refreshingDeferred` does not get resolved/rejected upon error. Also the deferred does not get reset to null during an error. See #334 

## What is the new behavior?

The deferred gets resolved upon AuthError and rejected upon other errors, and finally reset to null. Like this the deferred Promise behaves like the Promise returned by `_callRefreshToken` itself.

## Additional context

Closes #334

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -803,14 +803,14 @@ export default class GoTrueClient {
       return result
     } catch (error) {
       if (isAuthError(error)) {
-        const result = { session: null, error };
+        const result = { session: null, error }
 
-        this.refreshingDeferred?.resolve(result);
+        this.refreshingDeferred?.resolve(result)
 
-        return result;
+        return result
       }
 
-      this.refreshingDeferred?.reject(error);
+      this.refreshingDeferred?.reject(error)
       throw error
     } finally {
       this.refreshingDeferred = null
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -44,17 +44,10 @@ import type {
   SignInWithPasswordlessCredentials,
   AuthResponse,
   OAuthResponse,
+  CallRefreshTokenResult,
 } from './lib/types'
 
 polyfillGlobalThis() // Make "globalThis" available
-
-type CallRefreshTokenResult = {
-  session: Session
-  error: null
-} | {
-  session: null
-  error: AuthError
-}
 
 const DEFAULT_OPTIONS = {
   url: GOTRUE_URL,
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -37,6 +37,14 @@ import type {
 
 polyfillGlobalThis() // Make "globalThis" available
 
+type CallRefreshTokenResult = {
+  session: Session
+  error: null
+} | {
+  session: null
+  error: AuthError
+}
+
 const DEFAULT_OPTIONS = {
   url: GOTRUE_URL,
   storageKey: STORAGE_KEY,
@@ -83,10 +91,7 @@ export default class GoTrueClient {
   protected refreshTokenTimer?: ReturnType<typeof setTimeout>
   // eslint-disable-next-line @typescript-eslint/no-inferrable-types
   protected networkRetries: number = 0
-  protected refreshingDeferred: Deferred<{
-    session: Session
-    error: null
-  }> | null = null
+  protected refreshingDeferred: Deferred<CallRefreshTokenResult> | null = null
 
   /**
    * Create a new client for use in the browser.
@@ -884,17 +889,16 @@ export default class GoTrueClient {
     }
   }
 
-  private async _callRefreshToken(refresh_token = this.currentSession?.refresh_token) {
-    try {
-      // refreshing is already in progress
-      if (this.refreshingDeferred) {
-        return await this.refreshingDeferred.promise
-      }
+  private async _callRefreshToken(
+    refresh_token = this.currentSession?.refresh_token
+  ): Promise<CallRefreshTokenResult> {
+    // refreshing is already in progress
+    if (this.refreshingDeferred) {
+      return this.refreshingDeferred.promise
+    }
 
-      this.refreshingDeferred = new Deferred<{
-        session: Session
-        error: null
-      }>()
+    try {
+      this.refreshingDeferred = new Deferred<CallRefreshTokenResult>()
 
       if (!refresh_token) {
         throw new Error('No current session.')
@@ -910,15 +914,20 @@ export default class GoTrueClient {
       const result = { session, error: null }
 
       this.refreshingDeferred.resolve(result)
-      this.refreshingDeferred = null
 
       return result
     } catch (error) {
       if (isAuthError(error)) {
-        return { session: null, error }
+        const result = { session: null, error };
+
+        this.refreshingDeferred?.resolve(result);
+
+        return result;
       }
 
       throw error
+    } finally {
+      this.refreshingDeferred = null
     }
   }
 
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -925,6 +925,7 @@ export default class GoTrueClient {
         return result;
       }
 
+      this.refreshingDeferred?.reject(error);
       throw error
     } finally {
       this.refreshingDeferred = null
--- a/test/GoTrueClient.test.ts
+++ b/test/GoTrueClient.test.ts
@@ -1,4 +1,4 @@
-import { AuthError } from '../src/lib/errors';
+import { AuthError } from '../src/lib/errors'
 import {
   authClient as auth,
   authClientWithSession as authWithSession,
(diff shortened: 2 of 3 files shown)


src/GoTrueClient.ts
```diff
@@ -37,6 +37,14 @@ import type {
 
 polyfillGlobalThis() // Make "globalThis" available
 
+type CallRefreshTokenResult = {
```
pixtron, line 40: This type is solely used in private methods and properties. Not sure if it nevertheless should go to `lib/types.ts`? Both is fine for me.

src/GoTrueClient.ts
```diff
@@ -884,17 +889,16 @@ export default class GoTrueClient {
...
-      // refreshing is already in progress
-      if (this.refreshingDeferred) {
-        return await this.refreshingDeferred.promise
-      }
+  private async _callRefreshToken(
+    refresh_token = this.currentSession?.refresh_token
+  ): Promise<CallRefreshTokenResult> {
+    // refreshing is already in progress
+    if (this.refreshingDeferred) {
+      return this.refreshingDeferred.promise
```
pixtron, line 897: Moved that out of the try block and removed the `await`. Otherwise a rejection would fall into `catch` and `finally` multiple times. As `_callRefreshToken` throws non `AuthErorr`'s that should be the expected behavior.

test/GoTrueClient.test.ts
```diff
@@ -1,4 +1,5 @@
 import { OpenIDConnectCredentials } from '../src'
+import { AuthError } from '../src/lib/errors';
```
pixtron, line 2: Errors are not exported by the library. Would it make sense to export them so a consumer can import errors and compare them with `instanceof AuthError` or `isAuthError(error)`?

wattroll:
Thank you for looking into this. This fixes the problem for me.

- Coding wise, would it make sense to move the actual `callRefreshToken` logic into another private function (e.g. `callRefreshTokenConcurrently`) that will not be concerned with concurrency. While the `callRefreshToken` will be only creating/resolving/rejecting the deferred promises as appropriate to share the concurrent calls. That is just to keep things more robust against future changes. Just to make it less likely that somebody who touches error-matching logic, or session saving, or subscriber notifications will introduce another dangling promise bug.

- Have you considered adressing the problem of this being called with different refresh tokens simultaneously? e.g. `_callRefreshToken(token_a)` and `_callRefreshToken(token_b)` should not resolve to the same result. I don't think this is exposed to the users now, at least not in any common path, but it's a potential issue.

For reference here is an earlier implementation of shared token refreshes, where these two points are addressed https://github.com/supabase/gotrue-js/pull/203/files#diff-3522461172efd6058d6b8da62fc2d30d8b524d2b64894ea2c67218c52f7fdff5

pixtron:
Regarding @wattroll's points:

> would it make sense to move the actual callRefreshToken logic into another private function (e.g. callRefreshTokenConcurrently) that will not be concerned with simultaneous calls. The callRefreshToken will then only be creating/resolving/rejecting the deferred promises as appropriate to share the concurrent calls.

The implementation in #203 does not protect from race condition. `_callRefreshTokenConcurrently()` could accidentally be called internally without going through `_callRefreshToken()` and like this introduce concurrent calls to `this.api.refreshAccessToken()`. Therefore I prefer @alaister's implementation introduced into `next` with #285, regarding this point.

> Have you considered adressing the problem of this being called with different refresh tokens simultaneously? e.g. _callRefreshToken(token_a) and _callRefreshToken(token_b) should not resolve to the same result. I don't think this is exposed to the users now, at least not in any common path, but it's a potential issue.

This indeed is a potential issue. But the solution to have a promise/deferred for each token would shift the race condition down to `saveSession()`. When calling `_callRefreshToken(token_a)` and `_callRefreshToken(token_b)` concurrently the last one that resolves would write to the storage. As the storage in `gotrue-js` currently is a single user storage it probably would make more sense to abort pending refresh requests, when `_callRefreshToken()` is called concurrently with another refresh token (should be possible to do that by passing an abort signal to the fetch request).

My intention for this PR is solely to fix a Bug (see #334) introduced with [#285](https://github.com/supabase/gotrue-js/pull/285/files#diff-3522461172efd6058d6b8da62fc2d30d8b524d2b64894ea2c67218c52f7fdff5R756-R782). Probably it makes sense to open a new Issue with @wattroll 's points to discuss them further?

src/GoTrueClient.ts
alaister: Yeah let's put this in `lib/types.ts` to keep things consistent 😄

test/GoTrueClient.test.ts
alaister: Good point – we definitely should at least be exporting  `isAuthError()`

alaister (changes requested):
Thanks, @pixtron! This looks great!

Would you mind running prettier (`npm run format`) on your code and moving `CallRefreshTokenResult` to `lib/types.ts`, then I'll merge this in.

Thanks for adding tests too!

pixtron:
Moved `CallRefreshTokenResult` to `lib/types.ts` and comited prettier changes affecting code introduced with this PR. I didn't dare to commit prettier changes to other areas of the code base.

Prettier [changed](https://github.com/supabase/gotrue-js/pull/339/commits/7e20b31d269aea0221999946890cda488606928f#diff-375078a96c498c67bf7c49a00a71c6af7d8c8195e19958c18bc8862eac65cdc3L146-R154) this line

```ts 
const { session: session3, error: error3 } = await authWithSession._callRefreshToken(session?.refresh_token)
```
to 3 lines, which resulted in the TS error `Type 'undefined' is not assignable to type 'string'` during tests. When the code is on one line this error is not thrown. I decided to use a non null assertion as there is already a `expect(session).not.toBeNull()` a couple of lines above.

alaister (approved):
Thanks for making those changes.
I agree that the non-null assertion is fine there.

Cheers!

PR #341, opened 2022-08-03
fix: add custom error classes

kangmingtay:
* Use a custom error class that extends `AuthError` instead of throwing a generic error

### DIFF
--- a/src/GoTrueApi.ts
+++ b/src/GoTrueApi.ts
@@ -13,7 +13,7 @@ import { COOKIE_OPTIONS } from './lib/constants'
 import { setCookies, getCookieString } from './lib/cookies'
 import { expiresAt, resolveFetch } from './lib/helpers'
 
-import { isAuthError, AuthError, AuthApiError } from './lib/errors'
+import { isAuthError, AuthError, AuthApiError, AuthSessionMissingError, AuthEventMissingError, AuthNoCookieError } from './lib/errors'
 
 export default class GoTrueApi {
   protected url: string
@@ -632,9 +632,9 @@ export default class GoTrueApi {
     }
     const { event, session } = req.body
 
-    if (!event) throw new Error('Auth event missing!')
+    if (!event) throw new AuthEventMissingError()
     if (event === 'SIGNED_IN') {
-      if (!session) throw new Error('Auth session missing!')
+      if (!session) throw new AuthSessionMissingError()
       setCookies(
         req,
         res,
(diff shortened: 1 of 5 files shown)


src/lib/cookies.ts
```diff
@@ -112,7 +114,7 @@ function serialize(
  */
 function isSecureEnvironment(req: any) {
   if (!req || !req.headers || !req.headers.host) {
-    throw new Error('The "host" request header is not available')
+    throw new AuthUnknownError('The "host" request header is not available', new Error('The "host" request header is not available'))
```
silentworks, line 117: You have an `new Error` inside of your `new AuthUnknownError`.

silentworks (approved):
LGTM

src/lib/cookies.ts
kangmingtay: yeah just passing it through because the `AuthUnknownError` takes in a message and an error, since this code path doesn't have an actual underlying error, i'm just creating one and wrapping it with the `AuthUnknownError` class

src/GoTrueClient.ts
```diff
@@ -238,11 +238,9 @@ export default class GoTrueClient {
    * @param provider One of the providers supported by GoTrue.
    * @param redirectTo A URL to send the user to after they are confirmed (OAuth logins only).
    * @param scopes A space-separated list of scopes granted to the OAuth application.
-   * @param queryParams An object of query params 
+   * @param queryParams An object of query params
    */
-  async signInWithOAuth(
-    credentials: SignInWithOAuthCredentials,
-  ): Promise<OAuthResposne> {
+  async signInWithOAuth(credentials: SignInWithOAuthCredentials): Promise<OAuthResposne> {
```
J0, line 243: Should this be  `OAuthResponse` ? Not sure if this is done to deconflict with a keyword
```suggestion
  async signInWithOAuth(credentials: SignInWithOAuthCredentials): Promise<OAuthResponse> {
```

src/GoTrueApi.ts
```diff
@@ -697,9 +697,9 @@ export default class GoTrueApi {
     }
     const { event, session } = req.body
 
-    if (!event) throw new Error('Auth event missing!')
+    if (!event) throw new AuthEventMissingError()
```
J0, line 700: More a question than a review comment: when will we have auth events go missing? Not too sure how the events portion works.

If there's time, can I trouble you for a short overview pre/post sync?

src/lib/cookies.ts
J0: Think this is intended -- `AuthUnknownError` [seems to take in a string and constructor](https://github.com/supabase/gotrue-js/pull/341/files#diff-33a1fa952d3520ae3b8ff121fe89f8d401e10efd4cbf34f180663caa16c0f153L35)

J0 (approved):

src/GoTrueClient.ts
kangmingtay: ah good catch this is unintended, thanks!

PR #366, opened 2022-08-04
fix: remove unnecessary events and multi-tab

kangmingtay:
* Remove unnecessary events emitted 
* Remove multi-tab stuff because `getSession()` will always ensure that a valid session is returned or no session at all


alaister (approved):
lgtm!

PR #370, opened 2022-08-08
fix: cleanup localStorage session format

alaister:
Currently, the item in `localStorage` is stored with a surrounding object with a duplicated `expiresAt` timestamp.

This removes the surrounding object and stores the session directly. It also adds a check to ensure the item in `localStorage` is a valid session.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -366,7 +366,7 @@ export default class GoTrueClient {
     if (this.persistSession) {
       const maybeSession = await getItemAsync(this.localStorage, this.storageKey)
 
-      if (this._isValidSession(maybeSession)) {
+      if (this._doesSessionExist(maybeSession)) {
         currentSession = maybeSession
       } else {
         await this._removeSession()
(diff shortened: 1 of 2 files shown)


src/GoTrueClient.ts
```diff
@@ -601,6 +605,17 @@ export default class GoTrueClient {
     }
   }
 
+  private _isValidSession(maybeSession: unknown): maybeSession is Session {
+    const isValidSession =
+      typeof maybeSession === 'object' &&
+      maybeSession !== null &&
+      'access_token' in maybeSession &&
+      'refresh_token' in maybeSession &&
+      'expires_at' in maybeSession
```
alaister, line 614: Is this check sufficient enough to ensure the data in `localStorage` is actually a session?

kangmingtay: hmm why do we need to check if the data in `localStorage` is a valid session? i think a valid session should be defined as one that has a valid refresh token and an access token that hasn't expired yet

kangmingtay (approved):

src/GoTrueClient.ts
alaister: We need to check as after people upgrade, some people might have the old session format stored in `localStroage`, which will break us trying to read in that session. This check is just to make sure that what's there is probably a session

kangmingtay: hmm would renaming it to `_doesSessionExist` make it clearer?

alaister: Good call – renaming!

Issue #326, opened 2021-10-13
onAuthStateChange should fire on initial load


mikob:
# Feature request

## Is your feature request related to a problem? Please describe.
Currently it seems there is no deterministic way to know if a supabase session is done with initial loading.

If the user is signed out,`auth.session()` will return `null` and `onAuthStateChange` won't fire. If the user is signed-in, `auth.session()` will return `null`, but then `onAuthStateChange` will fire soon after with the session data.

Consider this code:
```
        const sess = supabase.auth.session();
        log("session event", sess);

        changeSession(sess);

        supabase.auth.onAuthStateChange((_event, session) => {
            log("session event", _event);
            changeSession(session);
        });
```

### A clear and concise description of what you want and what your use case is.
I need to load certain user history with knowledge of whether the user is signed-in or not. Yes it's possible to first fire assuming they aren't logged in, and then backtracking and firing again (once `onAuthStateChange` is hit) saying actually they are signed in, but it forces me to fire changeSession a function which is resource intensive on my end twice. 

## Describe the solution you'd like

Either `onAuthStateChange` should fire initially by default (how firebase auth works BTW), whether the user is signed in or not (perhaps a special INITIAL event) or should take a parameter {initial: true} to allow this behavior.

Stuk:
I think the best approach is to represent the initial initialization state as a promise. The API would look something like:

```ts
const client = new GoTrueClient(options);

// any time later
await client.initialized();
if (client.user()) {
  // signed in
} else {
  // not signed in
  client.signIn(credentials);
}
```

This avoids race conditions between the the construction of the client and adding event listeners, and allows one to check the initialization state at any point in the future.

I have created a pull request for this here: https://github.com/supabase/gotrue-js/pull/371

hf:
With v2 this is handled. `getSession` will reuse the initial promise so any waiters on it will resolve properly.

PR #371, opened 2022-08-09
Add initialized method to client

Stuk:
Returns a promise that is resolved when the initializing of the user session is complete.

Bug fix for https://github.com/supabase/gotrue-js/issues/326

It is not possible to know when the client has finished reading the saved session after construction.

## What is the new behavior?

The promise returned from `initialized()` is resolved when the saved session, if one exists, has been restored:

```ts
const client = new GoTrueClient(options);

// any time later
await client.initialized();
if (client.user()) {
  // signed in
} else {
  // not signed in
  client.signIn(credentials);
}
```

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -51,6 +51,10 @@ export default class GoTrueClient {
    */
   api: GoTrueApi
   /**
+   * Resolved when the initializing of the user session, if one exists, is complete.
+   */
+  protected initializedPromise: Promise<void>;
+  /**
    * The currently logged in user or null.
    */
   protected currentUser: User | null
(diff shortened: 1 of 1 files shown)


alaister:
Hey @Stuk,
You might like to checkout the `next` branch as we're moving away from the `.session()` and `.user()` methods in favour of asynchronous `.getSession()` and `.getUser()` methods that will always return a valid, refreshed session if there is one.

You can try these out by installing `@supabase/supabase-js@next` in your project.

With this in mind, unfortunately, I don't think these changes will be relevant with the release of supabase-js v2, so I'm going to close this PR. Of course, feel free to reopen it if there's something I'm missing.

Thanks for contributing!

Stuk:
Sounds good. Would be useful to communicate that on the linked issue

PR #372, opened 2022-08-11
feat: refactor storage key

kangmingtay:
* Expose an interface where the developer can use their own custom storage option. Defaults to localstorage.
* Rename `localstorage` property to a more generic name - `storage` 
* Remove `storageKey` property
* Add custom storage interface and a `LocalStorage` class that implements the `CustomStorage` interface

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -13,7 +13,7 @@ import {
   DEFAULT_HEADERS,
   EXPIRY_MARGIN,
   NETWORK_FAILURE,
-  STORAGE_KEY,
+  PERSISTENCE_KEY,
 } from './lib/constants'
 import { polyfillGlobalThis } from './lib/polyfills'
 import { Fetch } from './lib/fetch'
@@ -46,12 +46,13 @@ import type {
   OAuthResponse,
   CallRefreshTokenResult,
 } from './lib/types'
+import { LocalStorage } from './lib/storage/localStorage'
 
 polyfillGlobalThis() // Make "globalThis" available
 
 const DEFAULT_OPTIONS = {
   url: GOTRUE_URL,
-  storageKey: STORAGE_KEY,
+  persistenceKey: PERSISTENCE_KEY,
   autoRefreshToken: true,
   persistSession: true,
   detectSessionInUrl: true,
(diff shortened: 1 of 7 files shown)


kangmingtay:
The DX does feel abit weird to me though because now the user has to implement some class that implements the `CustomStorage` interface. 

I designed the `CustomStorage` interface to intentionally restrict storage access to the only 1 key. This key would be provided in the constructor of any class that implements `CustomStorage`

alaister:
This looks great!

I get what you're saying about the DX feeling a bit odd (maybe too class-based?)
What do you think about just requiring an object with the methods instead of a full-blown class? For example:
```
{
  getItem() {...},
  setItem(value) {...},
  removeItem() {...}
}
```

src/GoTrueClient.ts
```diff
@@ -46,12 +46,13 @@ import type {
...
   CallRefreshTokenResult,
 } from './lib/types'
+import { LocalStorage } from './lib/storage/localStorage'
 
 polyfillGlobalThis() // Make "globalThis" available
 
 const DEFAULT_OPTIONS = {
   url: GOTRUE_URL,
-  storageKey: STORAGE_KEY,
+  persistenceKey: PERSISTENCE_KEY,
```
alaister, line 55: Is this used anywhere?

src/GoTrueClient.ts
```diff
@@ -89,33 +85,30 @@ export default class GoTrueClient {
...
     cookieOptions?: CookieOptions
     fetch?: Fetch
   }) {
     const settings = { ...DEFAULT_OPTIONS, ...options }
     this.inMemorySession = null
-    this.storageKey = settings.storageKey
     this.autoRefreshToken = settings.autoRefreshToken
     this.persistSession = settings.persistSession
-    this.localStorage = settings.localStorage || globalThis.localStorage
+    this.storage = new LocalStorage({ key: PERSISTENCE_KEY })
```
inian, line 111: shouldn't you take this from the constructor and use `LocalStorage` as the default?

kangmingtay:
After discussions with @inian, we've decided to keep the existing implementation but rename `this.localstorage` to `this.storage` 

cc @alaister

PR #373, opened 2022-08-11
feat: return types should always contain `data`

kangmingtay:
* Common methods should always return an object with `{ data: {...}, error: ... }`


alaister (approved):
lgtm!

Issue #333, opened 2022-07-28
Backoff upon error during auto refresh


pixtron:
# Bug report

## Describe the bug

Backoff during auto refresh upon network failure fails in 1.23.0-next.  
 
## To Reproduce

1. clone the [repro](https://github.com/pixtron/supabase-repro)
2. npm install
3. npm start
4. Signup with email and password (or login if you already have an account)
5. Disable all network interfaces
6. Wait 2 minutes until token expires and auto refresh is triggered

The refresh request fails due to the client being offline. The client does not backoff and retry with further requests.
		
## Expected behavior

The client should backoff and retry to refresh the token.

## Suggested fix

This regression seems to have been introduced in 1.23.0-next in `lib/fetch.ts` with [#301](https://github.com/supabase/gotrue-js/pull/301/files#diff-437f2c15df7c90ad8154c5de1677ec0838d51859bcc0a0cefd8a0424b5736f31L17-R27).

1. Introduce a new Error `AuthNetworkFailureError exends AuthError`
2. Throw `AuthNetworkFailureError` instead of [`AuthUnknownError`](https://github.com/supabase/gotrue-js/blob/fb7fae5d763c7173f16aa8c6a8747013d8ba4d24/src/lib/fetch.ts#L30). Fetch only throws on a network error (unfortunately there seems to be no way to distinguish between network errors and CORS though).
3. Replace the check [`error?.message === NETWORK_FAILURE.ERROR_MESSAGE`](https://github.com/supabase/gotrue-js/blob/fb7fae5d763c7173f16aa8c6a8747013d8ba4d24/src/GoTrueClient.ts#L989) with `error instanceof AuthNetworkFailureError`
4. Remove constant [`NETWORK_FAILURE.ERROR_MESSAGE`](https://github.com/supabase/gotrue-js/blob/fb7fae5d763c7173f16aa8c6a8747013d8ba4d24/src/lib/constants.ts#L8)

## System information

- OS: NA
- Browser: NA
- Version of supabase-js: NA
- Version of gotrue-js: 1.23.0-next.6
- Version of Node.js: NA

## Additional context
 
The same mal function can probably be observed in [`_recoverAndRefresh`](https://github.com/supabase/gotrue-js/blob/fb7fae5d763c7173f16aa8c6a8747013d8ba4d24/src/GoTrueClient.ts#L847-L862).

kangmingtay:
closing this since it's been resolved by the PRs above! thanks @pixtron once again :)

Issue #289, opened 2022-05-26
Error messages not being parsed correctly


awalias:
recently we got a report from usage of supabase-js:

`FetchError: invalid json response body at https://fscmlciotnxwjzflqzqs.supabase.co/auth/v1/admin/users reason: Unexpected token < in JSON at position 0`

it's suspected that the error in question was either due to a failed request:
```
{
  "code": 422,
  "msg": "Email address already registered by another user"
}
// POST https://{REF}.supabase.in/auth/v1/admin/users
// HTTP/1.1 422 Unprocessable Entity
// Date: Thu, 26 May 2022 00:46:27 GMT
// Content-Type: application/json
// Content-Length: 69
```

or 

```
Status
524
MethodPOST
Timestamp1653499427722000
IP Address173.70.226.178
Origin CountryUS
Clientsupabase-js/1.35.3
```

a 524 is a cloudflare specific error

we should try and reproduce both and make sure we are handling the errors correctly in the client

pixtron:
`FetchError: invalid json response body ...` can be observed when `response.json()` is called on a response that is not valid json. Both `1.22.22` and `1.23.0-next.16` do currently not catch this error. 

See `1.23.0-next.16` 
https://github.com/supabase/gotrue-js/blob/67a6575102be9c50763ff5d06fd2ee44b18a69f3/src/lib/fetch.ts#L26-L28

PR #379 will catch this error and reject with an `AuthUnknownError`.

hf:
This has been fixed. Thank you @pixtron!

PR #379, opened 2022-08-13
Fix retry token refresh upon network failure and retry refresh upon 5xx errors

pixtron:

Bug fix

1. Since #301 the error `NETWORK_FAILURE.ERROR_MESSAGE` is not thrown anymore, without this message failed refresh request are not retryed anymore:
https://github.com/supabase/gotrue-js/blob/67a6575102be9c50763ff5d06fd2ee44b18a69f3/src/GoTrueClient.ts#L870-L877
2. When the server/cloudflare proxy sends a body with no or invalid json, the error handling throws an exception (see #289) and does not reject the promise.
3. 5xx server responses could be retryed as well (eg. cloudflares `524: A timeout occurred`)  

Resolves: #333
Related to: #289 

## What is the new behavior?

This PR introduces a new error `AuthRetryableFetchError` Exponential backoff will now retry if this error is returned. 
`handleError` will reject with `AuthRetryableFetchError` when  
- when fetch throws an error (behaviour as it was pre #301)
- the server responds with a 5xx status (new behaviour)

Additionaly when `response.json()` throws an error (eg. invalid json) the error is caught and the promise is rejected with an `AuthUnknownError`.

In order to test the error handling this PR adds a new dev dependency `jest-mock-server`, which let's us setup a mock server that responds with errors or aborts a request without sending any response.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,6 +28,7 @@
         "faker": "^5.3.1",
         "genversion": "^3.1.1",
         "jest": "^28.1.3",
+        "jest-mock-server": "^0.1.0",
         "jsonwebtoken": "^8.5.1",
         "npm-run-all": "^4.1.5",
         "prettier": "2.7.1",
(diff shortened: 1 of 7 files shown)


kangmingtay (approved):
good catch @pixtron, this is very nice!

pixtron:
just recognized i mentioned the wrong PR in the description. It's #301 that introduced the regression. Edited the description as well.

alaister (approved):
Thank you @pixtron!

PR #380, opened 2022-08-13
fix: refactor return types

kangmingtay:
* Return `{ data, error }` for more methods like `getUser` and `updateUser` 
* Differentiate `auth.getUser` and `auth.api.getUser` by renaming the former to `auth.getUserFromSession` 
* Remove `verifyMobileOTP`

### DIFF
--- a/test/GoTrueClient.test.ts
+++ b/test/GoTrueClient.test.ts
@@ -23,23 +23,20 @@ describe('GoTrueClient', () => {
     test('setSession should return no error', async () => {
       const { email, password } = mockUserCredentials()
 
-      const {
-        error,
-        data: { session },
-      } = await authWithSession.signUp({
+      const { error, data } = await authWithSession.signUp({
         email,
         password,
       })
       expect(error).toBeNull()
-      expect(session).not.toBeNull()
+      expect(data.session).not.toBeNull()
 
-      await authWithSession.setSession(session?.refresh_token as string)
+      await authWithSession.setSession(data.session?.refresh_token as string)
       const {
         data: { user },
         error: updateError,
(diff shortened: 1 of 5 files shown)


test/GoTrueClient.test.ts
```diff
@@ -33,8 +34,12 @@ describe('GoTrueClient', () => {
       expect(session).not.toBeNull()
 
       await authWithSession.setSession(session?.refresh_token as string)
-      const { user } = await authWithSession.update({ data: { hello: 'world' } })
+      const {
+        data: { user },
+        error: updateError,
```
alaister, line 39: We'll have to make a point in the docs of destructuring _after_ we check for the error. This will throw a JS error if `data` is `null` as it won't be able to destructure out the user. Wherever `{ data: { user }, error }` needs to be changed to:
```typescript
const { data, error } = authWithSession.xxx()

if (error) {
  throw error
}

const { user } = data
```

alaister (changes requested):
This looks great! I'm a fan of the consistency with the `data` return and also the renaming of the `getUser` method; it makes things clearer.

I could only find one thing that needs changing:

test/GoTrueClient.test.ts
alaister: Thanks - I think you've resolved every instance bar this one! (Unless GitHub is refusing to update because this part is marked for review or something?)

kangmingtay: fix in https://github.com/supabase/gotrue-js/pull/381

alaister (approved):
In that case looks good!

PR #381, opened 2022-08-15
fix: remove duplicated code in GoTrueApi

kangmingtay:
* `GoTrueApi` contains 2 types of methods
  * Methods meant for just wrapping gotrue's api
  * Methods that should only be used when using a service role key / admin key
* With the introduction of `getSession` and `this.inMemorySession` / `this.persistSession`, the methods in `GoTrueClient` can also be used in the server-side context
* This change helps to reduce the amount of duplicated code and tests in gotrue-js 
* Moved `api.updateUser`, `api.getUser`, `api.resetPasswordForEmail` and `api.refreshAccessToken` to `GoTrueClient`
* Remove `getUserFromSession` - user can just call `getSession` instead
* Renamed `api` namespace to `admin` for clarity - methods in the `GoTrueAdminApi` class should only be used with a service role key / admin key in a trusted server-side environment

* Referenced the pattern used in #337


src/lib/types.ts
```diff
@@ -262,7 +262,7 @@ export interface VerifyMobileOTPParams {
   email?: undefined
   phone: string
   token: string
-  type?: MobileOTPType
+  type: MobileOTPType
```
kangmingtay, line 265: this was optional in the past for backward compatibility reasons because we didn't support email otps so the user didn't have to pass a type since it would always be `type = sms`

src/GoTrueClient.ts
```diff
@@ -387,34 +432,40 @@ export default class GoTrueClient {
...
-        data: {
-          user: null
+  async getUser(): Promise<UserResponse> {
+    try {
+      let access_token = ''
+      if (!('Authorization' in this.headers)) {
+        const { session, error } = await this.getSession()
+        if (error) {
+          throw error
         }
```
kangmingtay, line 444: @alaister i'm getting the `Authorization` header from the `this.headers` if there's one - this is to account for the server-side usecase where one would create a new client per request and pass the `Authorization: Bearer token` in the header

src/GoTrueClient.ts
```diff
@@ -387,34 +432,40 @@ export default class GoTrueClient {
...
-
-    return { data: { user: session.user }, error: null }
   }
 
   /**
    * Updates user data, if there is a logged in user.
    * @param attributes The data you want to update.
    */
-  async update(attributes: UserAttributes): Promise<UserResponse> {
+  async updateUser(attributes: UserAttributes): Promise<UserResponse> {
```
kangmingtay, line 468: @alaister choosing not to ignore the value set in `this.headers` because `updateUser` actually updates the user in gotrue and also updates the session with the updated user details - maybe we should have a separate method for just updating the user so it can be used in a server-side context?

src/GoTrueClient.ts
```diff
@@ -423,7 +474,11 @@ export default class GoTrueClient {
       if (!session) {
         throw new AuthSessionMissingError()
       }
-      const { data, error: userError } = await this.api.updateUser(session.access_token, attributes)
+      const { data, error: userError } = await _request(this.fetch, 'PUT', `${this.url}/user`, {
+        body: attributes,
+        jwt: session.access_token,
+        xform: _userResponse,
```
kangmingtay, line 480: In v2.1, this will be refactored out instead of using it as a callback here

src/GoTrueClient.ts
```diff
@@ -887,4 +892,33 @@ export default class GoTrueClient {
...
   }
+
+  /**
+   * Generates the relevant login URL for a third-party provider.
+   * @param provider One of the providers supported by GoTrue.
+   * @param options.redirectTo A URL or mobile address to send the user to after they are confirmed.
+   * @param options.scopes A space-separated list of scopes granted to the OAuth application.
+   * @param options.queryParams An object of key-value pairs containing query parameters granted to the OAuth application.
+   */
+  private _getUrlForProvider(
```
kangmingtay, line 903: @alaister made this private because i don't see a usecase where someone would want to get the `/authorize` endpoint separately

src/GoTrueClient.ts
```diff
@@ -475,7 +523,7 @@ export default class GoTrueClient {
 
   /**
    * Gets the session data from a URL string
-   * @param options.storeSession Optionally store the session in the browser
+   * @param options.storeSession Optionally store the session in the browser or in-memory
```
kangmingtay, line 526: @alaister should we make `getSessionFromUrl` a private method too? not too sure why is there even an option here to `storeSession` - if this is set to `false` then the session wouldn't even be saved in the `this.inMemorySession` property

was thinking we can make this method private, remove `storeSession` and rely on the `this.persistSession` property to do this thing

src/GoTrueClient.ts
alaister: Does it update the JWT as well as the user's details? If it does we'll either need a new method or a config param for this one

src/GoTrueClient.ts
alaister: Nice. We can always add it back if needed!

src/GoTrueClient.ts
alaister: Sounds simpler to me! Once again we can always add it back if we discover a valid use case in the future. So yep, I'm a fan too

src/GoTrueClient.ts
```diff
@@ -588,73 +636,70 @@ export default class GoTrueClient {
...
-    password: string,
-    options: {
-      captchaToken?: string
-    } = {}
-  ) {
+  /**
+   * Generates a new JWT.
+   * @param refreshToken A valid refresh token that was returned on login.
+   */
+  async refreshAccessToken(refreshToken: string): Promise<AuthResponse> {
```
kangmingtay, line 677: @alaister should this be made private too? seems like it can be handled by `setSession` which also saves the session

alaister: This one I think we should leave public as users may want to manually refresh a session if they've updated the user info server-side for example

alaister: Can you remember why we had to pass a `refreshToken` to it, though, instead of using the one in the session?

src/GoTrueClient.ts
alaister: Sure, makes sense!
Do you think it's cleaner if you do something like `this.headers['Authorization'] ?? session.access_token` instead of having the whole if and extra variable?

alaister (approved):

src/GoTrueClient.ts
kangmingtay: on hindsight, maybe not because we would avoid invoking `getSession()` for no reason this way

src/GoTrueClient.ts
kangmingtay: decided to make it private since we have `setSession` does basically the same thing but also updates the session stored in memory / in the browser

src/GoTrueClient.ts
```diff
@@ -503,10 +551,10 @@ export default class GoTrueClient {
       const timeNow = Math.round(Date.now() / 1000)
       const expires_at = timeNow + parseInt(expires_in)
 
-      const { data, error } = await this.api.getUser(access_token)
+      const { data, error } = await this.getUser()
```
pixtron, line 554: @kangmingtay doesn't this change break magic links, oauth callbacks et all? Usually there shouldn't be a session in storage at this point. If there is a session in storage we can't be sure it is a session for the same user as the access_token in the url.

src/GoTrueClient.ts
pixtron: @kangmingtay i use the public method to have a predictable session initialization (wanted to create a PR for that issue anyway sooner or later). See the code below as an short explanation of the issue:

```ts
try {
      const supabase = createClient(SUPABASE_URL, SUPABASE_ANON_KEY, {
        detectSessionInUrl: false, 
        // if not disabled refresh session and getSessionFromUrl run concurrently 
        // (both might trigger a SIGN_IN when there is an access_token in the url and an 
        // old session in the storage)
        // also it's not possible to catch errors during getSessionFromUrl and  
        // getSessionFromUrl won't run when the url contains (?error=).
      });
      // as session refresh runs async at this point we don't know for sure if 
      // there is a session in storage or not. 

      const {session, error} = await supabase.auth.getSession();
      // at this point we know if there is a session in storage,
      // neverthless try to grab a fresh session from url
      
      const { session: urlSession, error: urlError } = await this._getSessionFromUrl();
      // at this point we know for sure if the user has a session or not

      if (
        urlSession !== null && session !== null
        &&
        urlSession.user?.id !== session.user?.id)
      {
        // if we have a session from storage and one from url and the user is
        // not the same in both sessions, the old session has been terminated (user signed out)
        // fire event --> ({type: 'USER_CHANGED', session: null};
      }

      if (urlSession || session) {
        // fire event --> {type: 'SIGNED_IN', session: urlSession || session};
      } else {
        // fire event --> {type: 'SIGNED_OUT', session: null};
      }
    } catch (err) {
      // fire event --> {type: 'SIGNED_OUT', session: null};
    } finally {
      supabase.auth.onAuthStateChange((event, session) => {
        // fire event --> {type: event, session};
      });
    }
```

src/GoTrueClient.ts
kangmingtay: yeah you're right, we're putting up a fix here (https://github.com/supabase/gotrue-js/pull/384) - thanks for pointing this out, many changes were made here and the goal was to simplify the structure of the code base as much as possible and this was unfortunately overlooked :/

src/GoTrueClient.ts
kangmingtay: hmm just wondering why do you have to manually call `getSessionFromUrl()` here? when you do `createClient`, this instantiates a `GoTrueClient` which will always run `_getSessionFromUrl` in the [constructor](https://github.com/supabase/gotrue-js/blob/next/src/GoTrueClient.ts#L133-L140)

pixtron: There are various reasons to disable `getSessionFromUrl` in the constructor:

1. `getSessionFromUrl` and  `_recoverAndRefresh` run concurrently in the constructor. If there is a session in storage (User A) and a valid access_token in the url (User B) we can't predict the outcome. Either the refreshed session is written last to storage or the session from url. 
2. After constructing `GoTrueClient` we don't know if `getSessionFromUrl` will resolve with a Session. There is no way to await it. `await getSession()` will immediately return when there is no session in storage, and not wait for `getSessionInUrl` to resolve/reject.
3. The [AuthUnknownError thrown](https://github.com/supabase/gotrue-js/blob/421a168ad67cd89eca4d1251c7a729b912efd526/src/GoTrueClient.ts#L137) thrown from `getSessionFromUrl` can't be caught as it's thrown in a `then` at a point when the constructor already successfully returned. See [this codesandbox](https://codesandbox.io/s/getsessionfromurl-error-txi7p6?file=/src/index.ts). 
4. `getSessionFromUrl` is only called if the url contains an `access_token`, but not if the url contains `error` and `error_description` (eg expired magic link).

Ideally gotrue-js would export a `async createGoTrueClient()` that does create the client and awaits initialization, but thats maybe something for v3. 
Initialization should probably first check if it's a callback url (login attempt), if true clear storage and call `getSessionFromUrl` otherwise call `recoverAndRefresh`.

PR #384, opened 2022-08-16
feat: update signIn methods, getUser and getSession

kangmingtay:
## What kind bugs does this PR fix?
* `signInWithPassword` was not sending a 'sign-in' event
  *  added sign-in event after a successful response
* `signInWithOAuth` ran into issues getting session from url because it was using `getUser()` which depended on an existing session to be set
  * fixed by allowing getUser to take in an optional access token jwt  
* `getSession` should also return { data, error }
*  Nit: Lowercase OTP to Otp


J0 (approved):

alaister (approved):
lgtm!

PR #385, opened 2022-08-16
Docs: Update react example to use gotrue-js/next

kangmingtay:
## What kind of change does this PR introduce
* This react example uses the latest changes from gotrue-js/next


example/react/src/App.js
```diff
@@ -9,37 +9,42 @@ const auth = new GoTrueClient({
 })
 
 function App() {
-  let [session, setSession] = useState(auth.session())
+  let [session, setSession] = useState()
   let [email, setEmail] = useState(localStorage.getItem('email') ?? '')
   let [phone, setPhone] = useState(localStorage.getItem('phone') ?? '')
   let [password, setPassword] = useState('')
   let [otp, setOtp] = useState('')
   let [rememberMe, setRememberMe] = useState(false)
```
alaister, line 18: I think we'll need to add a `useEffect` here that calls `getSession()` and then sets the result

alaister (approved):
We should consider adding a loading state to this example to prevent the logged out flash, but this works as a first pass!

PR #386, opened 2022-08-16
fix: typedocs docs

inian:


alaister:
@kangmingtay would you mind taking a look at this one since you were just doing some docs stuff?

kangmingtay (approved):

PR #387, opened 2022-08-16
fix: cleanup type docs

kangmingtay:
* Make the types and definitions pretty

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -37,11 +37,11 @@ import type {
   SignInWithOAuthCredentials,
   SignInWithPasswordCredentials,
   SignInWithPasswordlessCredentials,
+  SignUpWithPasswordCredentials,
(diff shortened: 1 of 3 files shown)


src/GoTrueClient.ts
```diff
@@ -141,55 +141,59 @@ export default class GoTrueClient {
 
   /**
    * Creates a new user.
-   * @type UserCredentials
    * @param options.redirectTo The redirect URL attached to the signup confirmation link. Does not redirect the user if it's a mobile signup.
```
inian, line 144: change to emailRedirectTo

inian (approved):

src/GoTrueClient.ts
kangmingtay: whoops, small bug, made it consistent with the sign-in methods and also made the types tighter in the process

PR #391, opened 2022-08-17
fix: wait for _getSessionFromUrl()

alaister:

Bug fix

`_getSessionFromUrl()` is not waited for in `getSession()`, leading to the session either coming from the URL or localStorage, depending on which one returns quicker. This is indeterministic behaviour.

## What is the new behaviour?

`getSession()` waits for `_getSessionFromUrl()`, meaning we always preference the session in the URL over localStorage.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -149,7 +149,7 @@ export default class GoTrueClient {
    */
   async signUp(credentials: SignUpWithPasswordCredentials): Promise<AuthResponse> {
     try {
-      this._removeSession()
+      await this._removeSession()
 
       let res: AuthResponse
       if ('email' in credentials) {
@@ -193,7 +193,7 @@ export default class GoTrueClient {
       const user: User | null = data.user
 
       if (data.session) {
-        this._saveSession(data.session)
+        await this._saveSession(data.session)
         this._notifyAllSubscribers('SIGNED_IN', session)
       }
 
@@ -212,7 +212,7 @@ export default class GoTrueClient {
    */
   async signInWithPassword(credentials: SignInWithPasswordCredentials): Promise<AuthResponse> {
     try {
-      this._removeSession()
+      await this._removeSession()
 
       let res: AuthResponse
       if ('email' in credentials) {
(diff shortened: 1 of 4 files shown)


kangmingtay (approved):

hf (commented):
I would also add the block to _saveSession and _removeSession to make sure any calls are properly ordered with the URL parsing logic.

```typescript
    try {
      // make sure we've read the session from the url if there is one
      await this.gettingSessionFromUrlPromise
    } catch (_error) {
      // ignore the error as it will be thrown by the constructor
    }
```

alaister:
> I would also add the block to _saveSession and _removeSession to make sure any calls are properly ordered with the URL parsing logic.

I think this will cause an infinite loop as `_getSessionFromUrl()` calls `_saveSession()`

pixtron:
@alaister there is still some issues. Right now im a bit in a hurry as I have to finish some specs for a client until tonight. I will create a PR towards your branch until 7am Melbourne Time.

alaister:
Thanks, @pixtron!
We'll aim to get this in around 1 pm Melbourne time, so please focus on your client work!

Thanks again for contributing ⚡️

src/GoTrueClient.ts
```diff
@@ -123,20 +124,22 @@ export default class GoTrueClient {
...
     this.url = settings.url
     this.headers = settings.headers
     this.fetch = resolveFetch(settings.fetch)
 
     if (settings.detectSessionInUrl && isBrowser() && !!getParameterByName('access_token')) {
       // Handle the OAuth redirect
-      this._getSessionFromUrl().then(({ error }) => {
+      this.gettingSessionFromUrlPromise = this._getSessionFromUrl().then(({ error }) => {
         if (error) {
           throw new AuthUnknownError('Error getting session from URL.', error)
```
pixtron, line 136: Unfortunately this error can't be caught and it might break an app: see this [codesandbox](https://codesandbox.io/s/getsessionfromurl-error-txi7p6?file=/src/index.ts)
The only way to return async errors from a constructor is via a promise.

pixtron:
Created two PR's: #393 and #394 The second builds up on the first. See description in PR. And comment above.

PR #409, opened 2022-08-22
fix: move isBrowser localStorage check + other small fixes

alaister:

Bug fix

If you provide a custom storage adapter server side, it acts as a no-op as `gotrue-js` only calls storage implementations in the browser.

## What is the new behaviour?

Storage is now always called if `persistSession` is `true`.

## Additional context

Cookie storage is now a viable option to use server side.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -26,6 +26,7 @@ import {
   setItemAsync,
   uuid,
 } from './lib/helpers'
+import localStorageAdapter from './lib/local-storage'
 import { polyfillGlobalThis } from './lib/polyfills'
 import type {
   AuthChangeEvent,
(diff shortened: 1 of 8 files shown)


src/GoTrueClient.ts
```diff
@@ -937,9 +938,10 @@ export default class GoTrueClient {
     }
 
     try {
-      window?.addEventListener('visibilitychange', () => {
+      window?.addEventListener('visibilitychange', async () => {
         if (document.visibilityState === 'visible') {
-          this._recoverAndRefresh()
+          await this.initializePromise
+          await this._recoverAndRefresh()
```
alaister, line 944: This is unrelated, but I've included it in this PR. If the user quickly switches tabs while the session is still initializing, it won't call `_recoverAndRefresh` prematurely.

src/GoTrueClient.ts
```diff
@@ -545,9 +538,14 @@ export default class GoTrueClient {
         return { data: { session: null, user: null }, error: error }
       }
 
-      await this._saveSession(data.session!)
+      if (!data.session) {
+        return { data: { session: null, user: null }, error: null }
+      }
```
alaister, line 543: Do you agree this is better? Are there any problems with doing it this way?

kangmingtay (approved):

PR #410, opened 2022-08-22
Added options with redirectTo for user update

mghizzo:

Accept `options` object with `redirectTo` and send it as a query parameter to the endpoint.

While GoTrue accepts a query parameter for url redirect, there is no way to set it inside gotrue-js.
It's probably not the best solution, it may be better to create a different method for user email change, as mentioned in the issue.

## What is the new behavior?

Accept `options` object with `redirectTo` and send it as a query parameter to the endpoint.

## Additional context

Look at [this issue](https://github.com/supabase/gotrue/issues/962)

### DIFF
--- a/src/GoTrueApi.ts
+++ b/src/GoTrueApi.ts
@@ -843,10 +843,17 @@ export default class GoTrueApi {
    */
   async updateUser(
     jwt: string,
-    attributes: UserAttributes
+    attributes: UserAttributes,
+    options?: {
+      redirectTo?: string
+    }
   ): Promise<{ user: User | null; data: User | null; error: ApiError | null }> {
     try {
-      const data: any = await put(this.fetch, `${this.url}/user`, attributes, {
+      const urlParams: string[] = []
+      if (options?.redirectTo) {
+        urlParams.push(`redirect_to=${encodeURIComponent(options.redirectTo)}`)
+      }
+      const data: any = await put(this.fetch, `${this.url}/user?${urlParams.join('&')}`, attributes, {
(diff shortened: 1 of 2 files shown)


kangmingtay:
Hey @atzmanuel, thanks for flagging this out! this would need some fixing on gotrue's `PUT /user` endpoint since i would prefer to name the `redirect_to` param to something more explicit so it's clear that it will only apply to a user email update.

atzmanuel:
> Hey @atzmanuel, thanks for flagging this out! this would need some fixing on gotrue's `PUT /user` endpoint since i would prefer to name the `redirect_to` param to something more explicit so it's clear that it will only apply to a user email update.

Hey @kangmingtay, thanks for your reply. I understand that it's not clear that it will apply only to email change but,
I don't think that renaming parameter on gotrue's endpoint is possible since  it uses a global API method called `getReferrer` which is shared across the project.
An alternative solution may be creating another method like `changeEmail` which accept `redirectTo` param and leave `update()` method as it is. 
The method `changeEmail` will require a mandatory parameter `newEmail` and an optional `redirectTo` parameter and will fire the same `PUT /user` endpoint.
Let me know what do you think about this solution.

silentworks:
Closing this one out as this was fixed in #610.

JonathanPicques:
@silentworks I may be mistaken, but the pull request was about "Remove system user" and there is no mention to redirect in that PR's commit, is it really fixed?

kangmingtay:
@JonathanPicques i think @silentworks was referring to this PR instead: https://github.com/supabase/gotrue-js/pull/610

PR #414, opened 2022-08-25
feat: add mfa client library binding(deactivated by default)

J0:
Introduces the additional MFA methods for admin and regular endpoints. 

Admin:
- `deleteFactor`
- `listFactors`

Regular:
- `verify`
- `enroll`
- `unenroll`
- `challenge`
-  `listFactors`
-  `getAuthenticatorAssuranceLevel`

Pull Requests involved:
- #440
- #446
- #472
- #475
- #478
- #479
- #480
- #486
- #506
- #508
- #510


src/GoTrueAdminApi.ts
```diff
@@ -227,4 +235,8 @@ export default class GoTrueAdminApi {
       throw error
     }
   }
+
+  async _deleteFactor() {
```
hf, line 239: Mark it as `private` to be hidden from TypeScript users? (Note there's no such thing as "private" in JavaScript, it's only enforced by TypeScript, thus the `_` prefix.)

src/GoTrueAdminApi.ts
```diff
@@ -213,7 +223,7 @@ export default class GoTrueAdminApi {
    *
    * This function should only be called on a server. Never expose your `service_role` key in the browser.
    */
-  async deleteUser(uid: string): Promise<UserResponse> {
+  private async deleteUser(uid: string): Promise<UserResponse> {
```
kangmingtay, line 226: hmm why is this private?

J0: Yup, good catch — It’s not meant to be private—  was testing something and forgot to change back. Let me also delete the unused admin functions

kangmingtay:
@J0 hmm can we omit the docs stuff and keep that in a separate PR?

hf (approved):
🚀

PR #420, opened 2022-08-31
fix: handle custom fetch error responses

pixtron:

Bug fix

If a custom fetch implementation does not return a Response that is a instance of native fetch Response or a cross-fetch Response the `instanceof` check in the error handler fails.

## What is the new behavior?

The fetch responses are checked with a utility method

## Additional context

See [discussion in discord](https://discord.com/channels/839993398554656828/1011280048495542402)

### DIFF
--- a/src/lib/fetch.ts
+++ b/src/lib/fetch.ts
@@ -1,4 +1,4 @@
-import { expiresAt, resolveResponse } from './helpers'
+import { expiresAt, looksLikeFetchResponse } from './helpers'
 import {
   AuthResponse,
   GenerateLinkProperties,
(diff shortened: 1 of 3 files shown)


pixtron:
If the custom fetch response returns a response that has a `json()` method that doesn't return a Promise (no `then()`). This part would fail. That could for example happen in a project not using typescript. Not sure if that should be covered or not.
https://github.com/supabase/gotrue-js/blob/3d89361deff28eedb4f319ed00245e9b2f9f4928/src/lib/fetch.ts#L37-L45

alaister (approved):
Thank you again @pixtron!
Will merge this in on Monday.

PR #421, opened 2022-08-31
Update examples to use .env file

silentworks:

Cleanup the examples so that you only need to update the .env file instead of copying and pasting code from the README


example/next/utils/initSupabase.js
```diff
@@ -1,11 +1,13 @@
 import { GoTrueClient } from '@supabase/gotrue-js'
 
-const supabaseURL = 'http://localhost:9999'
+const supabaseURL = process.env.NEXT_PUBLIC_SUPABASE_URL
+const supabaseAnon = process.env.NEXT_PUBLIC_SUPABASE_ANON_KEY
 
 const auth = new GoTrueClient({
-  url: `${supabaseURL}`,
+  url: `${supabaseURL}/auth/v1`,
```
kangmingtay, line 7: hmm why do we add the `/auth/v1` here?

example/react/src/App.js
```diff
@@ -2,10 +2,15 @@ import React, { useState } from 'react'
 import { GoTrueClient } from '@supabase/gotrue-js'
 import './tailwind.output.css'
 
-const GOTRUE_URL_AUTOCONFIRM = 'http://localhost:9998'
+const supabaseURL = process.env.REACT_APP_SUPABASE_URL
+const supabaseAnon = process.env.REACT_APP_SUPABASE_ANON_KEY
 
 const auth = new GoTrueClient({
-  url: GOTRUE_URL_AUTOCONFIRM,
+  url: `${supabaseURL}/auth/v1`,
```
kangmingtay, line 9: ditto

kangmingtay (changes requested):

example/next/utils/initSupabase.js
silentworks: That's the auth endpoint.

kangmingtay: ah yeah i thought the `supabaseURL` included that already - it might lead to confusing behaviour if someone's using gotrue-js outside of supabase and trying out this next example but i'm ok with going ahead with this and changing it later if we need to since it's just an example

hf:
@silentworks What's the status of this PR?

silentworks:
Seems out of sync right now, but let me fix the conflicts and we should be able to get this merged in.

silentworks:
Closed in favor of #605

PR #425, opened 2022-09-02
refactor: rename `GoTrueAdminApi` to `GoTrueAdminClient`

Thore1954:


kangmingtay:
hey @Thore1954, would you mind explaining the purpose of this change?

kangmingtay:
hey @Thore1954, unfortunately, we've decided not to proceed with this change as it seems more of a "preference" rather than an actual issue.

Issue #205, opened 2022-01-08
GoTrue JS Client sends incorrect content-type in the request header


chipilov:
# Bug report

## Describe the bug

The network requests generated from calls to ```signUp()```, ```signIn()``` and ```generateLink()``` send the following request header:
```
content-type: text/plain;charset=UTF-8
```

However, the payload of the requests is JSON so the content-type header seems incorrect. 

NOTE: I have NOT tested the other methods in the GoTrue JS client (e.g. signOut, etc) - they might suffer from the same issue.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Init supabase js client: 
2. Send a sign-up request via the client: ```await supabase.auth.signUp({ email, password });```

## Expected behavior

The request to ```https://projectid.supabase.co/auth/v1/signup``` contains a request header ```content-type: application/json```

## Actual behavior

The request to ```https://projectid.supabase.co/auth/v1/signup``` contains a request header ```content-type: text/plain;charset=UTF-8```

## System information

- Supabase JS version: 1.28.5
- GoTrue JS version: 1.21.7

hf:
Closing since this has been fixed with PR #429 and already fixed in v2. Thanks @TeeWallz!

PR #429, opened 2022-09-05
fix: send `application/json` in Content-Type header

TeeWallz:
Resolves issue #205 

supabase/gotrue:v2.16.3 does not permit requests with content-type `text/plain` and is therefore replaced with `application/json`


hf (approved):
Looks good, I'll port this to v2 just now (if it applies).

hf:
This is already fixed in v2.

PR #440, opened 2022-09-15
fix: mfa verify should set session & emit auth event

kangmingtay:
* `mfa.verify` returns a user and a session on a successful verify
* `mfa.verify` updates the current session to the newly returned session (with aal2)
* Related to: https://github.com/supabase/gotrue/pull/682


src/GoTrueClient.ts
```diff
@@ -1040,23 +1040,30 @@ export default class GoTrueClient {
...
     try {
-      return await _request(
+      const { data, error } = await _request(
         this.fetch,
         'POST',
         `${this.url}/user/${user.id}/factor/${params.factorId}/verify`,
         {
           body: { code: params.code, challenge_id: params.challengeId },
           headers: this.headers,
+          xform: _sessionResponse,
```
J0, line 1055: What does is the xform property used for again?

src/lib/types.ts
```diff
@@ -529,7 +525,7 @@ export type DeleteFactorParams = {
 }
 
 export interface GoTrueMFAApi {
-  verify(params: MFAVerifyParams): Promise<AuthMFAResponse>
+  verify(params: MFAVerifyParams): Promise<AuthResponse>
```
J0, line 528: Might make sense to standardise all methods to either `AuthMFAResponse` or `AuthResponse`. Should we update AuthResponse and change the rest of the methods to AuthResponse instead as well? I can do it in a separate PR.

J0 (approved):

src/GoTrueClient.ts
kangmingtay: it acts as a callback function and is used to transform the API response returned into a desirable format

Issue #442, opened 2022-09-17
2nd Tab starting 10 seconds before expiration of session can cause 400 error on token refresh and loss of refresh tokens


GaryAustin1:
# Bug report

## Describe the bug
If a tab is running and a second tab starts up with createClient (or possibly becoming visible ... not tested) 10 seconds before expiration of session, a timing rounding error appears to cause refreshing of tokens to fail because of a reuse violation.

This is somewhat random, but by building a timer to delay the createClient call of the second tab to line up with the 10 second expiration time frame it can be made to fail with a "few" trys.

I suspect the failure is due to rounding error and network timing variability right at the 10 second boundary in this code:
https://github.com/supabase/gotrue-js/blob/a15296288d15919d2ec8a64654a545a375fb3a57/src/GoTrueClient.ts#L873
```
    const expiresAt = session.expires_at
    if (expiresAt) {
      const timeNow = Math.round(Date.now() / 1000)
      const expiresIn = expiresAt - timeNow
      const refreshDurationBeforeExpires = expiresIn > EXPIRY_MARGIN ? EXPIRY_MARGIN : 0.5
      this._startAutoRefreshToken((expiresIn - refreshDurationBeforeExpires) * 1000, session)
```

A failing case:
![10secondviolation](https://user-images.githubusercontent.com/54564956/190873852-9477fb26-075f-4d44-938a-464a4e14e85c.JPG)

Note: The 2nd tab starts with 211msec left before the session expires on it's refresh timer based on the above formula in this case.

The refresh timers set up on alternating 10 seconds time frames.

In case #3 one tab uses the refresh token and the next tab tries to refresh the same token 10.116 seconds later and violates the default reuse setting of 10 seconds, resulting in an invalid token error and logout of user.

I suspect the same error could occur on a tab coming back visible right at the 10 second boundary also.

## To Reproduce

This is somewhat difficult to reproduce constantly...
Basically I wrote javascript to init the client and monitor token refresh.  I use a localStorage variable to set the time the first tab starts and use that to delay the start of the client init on the second tab by 20 seconds (with a 30 second token refresh).  The first tab also does a signIn, the second does not and relies on the Supabase session in localStorage.

```
console.log('start')
async function app (supabase1) {
    let supabase = supabase1.createClient(SUPABASE_URL, SUPABASE_KEY)
    if (firstTab) {
        console.log('first tab signin', firstTab)
        const {data} = await supabase.auth.signInWithPassword({
                email: 'email',
[...]

PR #444, opened 2022-09-19
Support refresh token syncing for multiple tabs

miguelespinoza:

This change introduces a mechanism to sync refresh tokens at the client level, rather than depending on the backend.
The current behavior of supabase-js v2 unfortunately continues to log out users. This was reproducible with at least three tabs. More information here: https://github.com/supabase/gotrue-js/issues/454 and https://github.com/supabase/gotrue-js/issues/442

### Implementation Details

The sync logic uses `this.storage` to cache the status of the refresh token request. The only properties stored are `refresh_token` and `status`. Check `SyncTokenRefreshStorage`.

- `TOKEN_REFRESHING` is set on status before the API request is made.
- `TOKEN_REFRESHED` is set on status after the API request resolved.

For any tab that attemps to refresh the token.

- If status in storage is `TOKEN_REFRESHING`. It waits until token has refreshed, then pulls latest session from storage
- If status in storage is `TOKEN_REFRESHED`. it pulls latest session from storage
- In both cases, the latestSession is validated, to check if it’s not about to expire.

The user is logged out with at least three tabs: https://github.com/supabase/gotrue-js/issues/454
Notice a burst of network requests to the API. Eventually, this causes a 400 error, logging out the user. Check timestamps.
* This screenshot represents 8 instances, lasting only 9 minutes before timing out.

cleanshot_09_19_at_14_54@2x

## What is the new behavior?
The tabs depend on only one request to succeed. The remainder tabs wait until the refresh token is retrieved from the API to resolve. Providing a clean POST request of one to the API. Check timestamps.

cleanshot_09_19_at_15_12@2x

## Additional context
* Multi-requests can still happen if the timer for tabs have an identical cycle. This is where we would fallback on the backend with the reuse interval logic
* Another idea I had was to idle the supabase instance if the user put the tab in the background. I was concerned this approach would introduce unwanted behaviors for devs that use supabase requests while in the background. Let me know if this is worth pursuing. Just throwing that idea out there.
* For debugging I found `mitmproxy` extremely helpful to view the requests
   * Configuration details for mac: https://blogs.aaddevsup.xyz/2018/04/tracing-all-network-machine-traffic-using-mitmproxy-for-mac-osx/
   * Run `mitmproxy --mode socks5 --showhost --view-filter="~u supabase.co ~m POST"`
* Here are the auth configuration changes I tested. 30 second JWT expiry and 10 second reuse interval
* Definitely a preliminary change. I'd love to get feedback and potential gotchas. So far with testing, no 400 errors. I plan on testing this further with a production app I'm building. As this was causing me a bit of headache. The production app is currently running supabase-js v1, but I still found issues in v2.

Edit: Tue Sept 20 - Included Implementation Details

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -88,6 +88,8 @@ export default class GoTrueClient {
   protected storage: SupportedStorage
   protected stateChangeEmitters: Map<string, Subscription> = new Map()
   protected refreshTokenTimer?: ReturnType<typeof setTimeout>
+  protected refreshTokenSyncIntervalTimer?: ReturnType<typeof setTimeout>
+  protected refreshTokenSyncTimeoutTimer?: ReturnType<typeof setTimeout>
   protected networkRetries = 0
   protected refreshingDeferred: Deferred<CallRefreshTokenResult> | null = null
   /**
@@ -900,23 +902,23 @@ export default class GoTrueClient {
 
   private async _waitOnRefreshTokenSync(): Promise<void> {
     return new Promise((resolve) => {
-      const refreshTokenInterval = setInterval(async () => {
+      this.refreshTokenSyncIntervalTimer = setInterval(async () => {
         const refreshTokenSync = (await getItemAsync(
           this.storage,
           this.refreshTokenSyncStorageKey
         )) as SyncTokenRefreshStorage | undefined
 
         if (refreshTokenSync?.status === 'TOKEN_REFRESHED') {
-          clearInterval(refreshTokenInterval)
-          clearTimeout(refreshTokenTimeout)
+          clearInterval(this.refreshTokenSyncIntervalTimer)
+          clearTimeout(this.refreshTokenSyncTimeoutTimer)
           resolve()
         }
       }, 100)
 
       // Stop interval if tokenSync.status does not change
-      const refreshTokenTimeout = setTimeout(async () => {
+      this.refreshTokenSyncTimeoutTimer = setTimeout(async () => {
         await removeItemAsync(this.storage, this.refreshTokenSyncStorageKey)
-        clearInterval(refreshTokenInterval)
+        clearInterval(this.refreshTokenSyncIntervalTimer)
         resolve()
       }, 1500)
     })
@@ -967,6 +969,14 @@ export default class GoTrueClient {
 
     if (this.refreshTokenTimer) {
       clearTimeout(this.refreshTokenTimer)
+    }
+
+    if (this.refreshTokenSyncIntervalTimer) {
+      clearInterval(this.refreshTokenSyncIntervalTimer)
+    }
+
+    if (this.refreshTokenSyncTimeoutTimer) {
+      clearTimeout(this.refreshTokenSyncTimeoutTimer)
     }
   }
 
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -43,6 +43,7 @@ import type {
   SignUpWithPasswordCredentials,
   Subscription,
   SupportedStorage,
+  SyncTokenRefreshStorage,
   User,
   UserAttributes,
   UserResponse,
@@ -70,6 +71,11 @@ export default class GoTrueClient {
    * The storage key used to identify the values saved in localStorage
    */
   protected storageKey: string
+
+  /**
+   * The storage key used to keep the refresh token synced with tabs
+   */
+  protected refreshTokenSyncStorageKey: string
 
   /**
    * The session object for the currently logged in user. If null, it means there isn't a logged-in user.
@@ -105,6 +111,7 @@ export default class GoTrueClient {
     const settings = { ...DEFAULT_OPTIONS, ...options }
     this.inMemorySession = null
     this.storageKey = settings.storageKey
+    this.refreshTokenSyncStorageKey = `${this.storageKey}-rts`
     this.autoRefreshToken = settings.autoRefreshToken
     this.persistSession = settings.persistSession
     this.storage = settings.storage || localStorageAdapter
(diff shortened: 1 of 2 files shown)


src/GoTrueClient.ts
```diff
@@ -855,6 +900,36 @@ export default class GoTrueClient {
...
+      }, 100)
+
+      // Stop interval if tokenSync.status does not change
+      this.refreshTokenSyncTimeoutTimer = setTimeout(async () => {
+        await removeItemAsync(this.storage, this.refreshTokenSyncStorageKey)
+        clearInterval(this.refreshTokenSyncIntervalTimer)
+        resolve()
+      }, 1500)
+    })
+  }
```
miguelespinoza, line 925: I saw there's an exponential backoff for network retry. 
I didn't think it was required here. A request to refresh a token takes on average 150-300ms.

src/GoTrueClient.ts
```diff
@@ -822,19 +831,55 @@ export default class GoTrueClient {
...
-      this._notifyAllSubscribers('TOKEN_REFRESHED', data.session)
+      if (!syncedSession) {
+        await setItemAsync(this.storage, this.refreshTokenSyncStorageKey, {
+          refresh_token: refreshToken,
+          status: 'TOKEN_REFRESHING',
+        })
 
-      const result = { session: data.session, error: null }
+        const { data, error } = await this._refreshAccessToken(refreshToken)
+        if (error) throw error
```
miguelespinoza, line 869: I'm looking closer into this case when an error occurs with the request.
So far with a network disconnection, calls resolve appropriately.
I just want to make sure everything is well accounted for, so just sharing what's next on my list to validate

miguelespinoza:
I'll hold on updating any test until we're all cool with these changes.
In the meantime, I'll be testing this change very soon in production 🤞

miguelespinoza:
Follow-up conversation from Discord: https://discord.com/channels/839993398554656828/1019328795129421944

> On chrome/firefox and I'm sure mobile, timers get throttled in background.
https://developer.chrome.com/blog/timer-throttling-in-chrome-88/#intensive-throttling
You have potentially alot of timers that will get throttled to 1 minute (realtime breaks in background because of loss of heartbeat for instance).  Need to think thru what that could do here.  Also on mobile these tabs might not run at all until they become visible again.  gotrue-js added a check for coming back from hidden recently.
From @GaryAustin1


This is a good assessment. I'd like to avoid any unnecessary work that could use up resources. The chrome documentation talks about event changes. This is what the multiTab option supported. Could we get an explanation for why this was removed?

Another add-on could be idling the API request to refresh token if the tab is in the background? Any concerns here? @GaryAustin1 mentions realtime logic takes a break when in the background. We could apply the same behavior to the refresh token logic.

GaryAustin1:
Just to clarify on realtime in background... It does not take a break.  I have code using a visibility handler to make it "take a break" by stopping it before the background timing errors occur (5 minutes).

miguelespinoza:
There's been an awesome coversation on Discord. Started [here](https://discord.com/channels/839993398554656828/1019328795129421944/1022065258606632981)
So here's a recap:

* gut feeling from a few members, this will cause timing issues due to throttling from browsers. It's a very good point
* "Wouldn't it be better to just, check here if there is a new session in storage?" Maybe that's the trick, but I'm curious why was multiTab removed. Was that to put full dependency on the backend with reuse interval support?
*  "if i pass a custom InMemoryStorage i won't need the checks at all." Yup, I made a suggestion to add a flag `refreshTokenSync` that enables this logic
*  Remove the interval polling all together and just check if the session has refreshed before making a network request.
   * 💡  This could be a smaller change, so this could be a **candidate solution**
   * This requires a collaboration with backend team to understand why a 400 returns with multiple tabs. Issue reported here: https://github.com/supabase/supabase/issues/8959


> Another thing: if i pass a custom InMemoryStorage i won't need the checks at all. So my thinking goes that handling multitab is related to the storage in use. Therefore the actual storage should implement the handling of expired refresh tokens. Eg for local storage this could be solved with storage events (https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event) or eventually with a SharedWorker (https://developer.mozilla.org/en-US/docs/Web/API/SharedWorker?retiredLocale=de).
For your case this could potentially solved with storage.onChanged event (https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/onChanged) or a SharedWorker as well. Not sure how that would be solved in ract native (is there even a multitab scenario?). 

Relying on localStorage events was supported in v1, but removed in v2. I don't know it's history, so it would be great to understand the decision for removing `multiTab` support. With these issues encountered, it almost feels like we need to re-introduce `_listenForMultiTabEvents`


All in all, this information is helping towards finding an optimal solution. 
* I have a feeling this requires another review in the backend to understand the 400 errors: https://github.com/supabase/supabase/issues/8959
* This change may not be adequate, mainly because of the interval polling, but maybe we take some pieces like checking if there's a new session in storage before making a request. That's a small change
* Re-introduce multiTab support

miguelespinoza:
Here's my realization. For a chrome extension that uses supabase on multiple tabs. I need to opt out of the `autoRefreshToken` option and handle that on my own, using browser native methods. 

* Instead of a `setTimeout`. Use `chrome.alarm` to trigger an event 10 seconds before token expiration.
* On alarm received, refresh token manually
* Update that in `chrome.storage`. Which is passed to supabase using the `storage` option
* Update remainder tabs using `storage.onChange` picks up the new token in other tabs. I need to figure out if/how we can bring back `_listenForMultiTabEvents` and make it abstract so any storage engine can support this behavior. 

If this works, I wouldn't require the PR changes anymore. But I'm sure others would benefit from a fix on the library because this has been reproduced in a regular project: https://github.com/supabase/gotrue-js/issues/442

miguelespinoza:
Here's a WIP implementation of refresh token logic for browser extensions.
https://gist.github.com/miguelespinoza/59d8dc57aaeae27ffac87989f06e8f0e

It's not complete, but the core is in the gist. Two gotchas:
1. Setting `autoRefreshToken: false` causes the library to remove the token if the extension starts with network disconnection
   * It's due to this [else statement](https://github.com/supabase/gotrue-js/blob/master/src/GoTrueClient.ts#L677-L679) when calling `_recoverAndRefresh`.
   * Is the team open to an option like `refreshTokenHandledByClient`, to allow the app to handle refresh token logic? I'm open to suggestions
2. I originally started to add support for `browser.storage.onChange`. This was going to listen to token changes in storage. But I noticed that a supabase request calls `auth.getSession()`. Is it safe to say I don't need to handle token syncing, because `supabse.from...` queries for an up-to-date auth token?

This trace would indicate `fetch` ends up calling `supabase.auth.getSession()` on each request. I have to pinpoint where in the library this happens..

cleanshot_09_21_at_15_29@2x

src/GoTrueClient.ts
```diff
@@ -822,19 +831,55 @@ export default class GoTrueClient {
...
+        syncedSession = data.session
+
+        await setItemAsync(this.storage, this.refreshTokenSyncStorageKey, {
+          refresh_token: syncedSession.refresh_token,
+          status: 'TOKEN_REFRESHED',
+        })
+      }
+
+      await this._saveSession(syncedSession)
+      this._notifyAllSubscribers('TOKEN_REFRESHED', syncedSession)
```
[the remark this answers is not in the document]
kangmingtay: I think this implementation is fine, imo `TOKEN_REFRESHING` seems like an implementation detail and we don't have to add it to the `AuthChangeEvent` type.

src/GoTrueClient.ts
```diff
@@ -105,6 +113,7 @@ export default class GoTrueClient {
     const settings = { ...DEFAULT_OPTIONS, ...options }
     this.inMemorySession = null
     this.storageKey = settings.storageKey
+    this.refreshTokenSyncStorageKey = `${this.storageKey}-rts`
```
kangmingtay, line 116: is it possible to actually storage this information in the same storage used for the session?

kangmingtay (commented):
Hey @miguelespinoza, thanks for the detailed explanation and taking the time to investigate this issue. I've tested this PR on my end with the `JWT_EXPIRY` set to 20s, `REUSE_INTERVAL` set to 10s with 4 tabs and it's looking good. 

We got rid of the multi-tab property because when we tested the v2 implementation initially, we couldn't reproduce the error. Also, i don't think the `multiTab` property would've helped in this scenario to prevent a refresh outside the reuse interval. It was mainly used to add a listener to the storage event.

kangmingtay:
One thing i'm trying to figure out is how the client can get to the point where one of the refreshes happen outside the reuse interval. 10 seconds is a long time and i suspect that the reason for this is because the session passed into the `_startAutoRefreshToken` method is outside of the reuse interval at the point in time where the [callback is invoked](https://github.com/supabase/gotrue-js/blob/a15296288d15919d2ec8a64654a545a375fb3a57/src/GoTrueClient.ts#L911-L923).

kangmingtay:
Hey @miguelespinoza, i think i've figured out the issue, seems like the session passed into the `_startAutoRefreshToken`  method can possibly become stale and fall outside the reuse interval. I've made a PR [here](https://github.com/supabase/gotrue-js/pull/461). Would be awesome if you can try it out and see if it fixes the problem for you!

miguelespinoza:
Thanks for submitting a PR. I think that would be a better change than this. Your change is small, and it's a change I advocated for after submitting this PR.

I've since gone with a different solution commented [here](https://github.com/supabase/gotrue-js/pull/444#issuecomment-1253750718).
I'm using the `chrome.alarms` API to mimic the timer in the library. It'll be a stable solution moving forward when I migrate my extension to MV3 which uses a service worker and shies away from timers. More details [here](https://developer.chrome.com/docs/extensions/mv3/migrating_to_service_workers/). This also has the added benefit of only one instance handling the token refresh logic rather than the first come first serve style.

Others had this issue too, so I'm learning on them to help with testing 😁. 

While I have your attention, I'm working on another project and had to downgrade from v2 to v1, because of this tangential auth issue: https://github.com/supabase/gotrue-js/issues/450. Just sharing if not already on your radar

kangmingtay:
hey @miguelespinoza, yup we're gonna look into #450 next, thanks for taking the time and effort to make this PR though!

PR #445, opened 2022-09-21
feat: add `setSession` support for a SSR context

hf:
Adds support for calling `setSession()` in a server-side rendering context. When used in SSR, the access and refresh tokens are typically sent from the browser to the server. Both of them represent the user's session. 

Users can thus call:

```typescript
await client.setSession({ 
  access_token: req.cookies['my-access-token'], 
  refresh_token: req.cookies['my-refresh-token'] 
})
```

To correctly extract the session information from the two cookies. The client can then be used to call other APIs and automatic refresh token management will take place.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -44,6 +44,24 @@ const DEFAULT_OPTIONS = {
   headers: DEFAULT_HEADERS,
 }
 
+const decodeBase64URL = (value: string): string => {
+  try {
+    // atob is present in all browsers and nodejs >= 16
+    // but if it is not it will throw a ReferenceError in which case we can try to use Buffer
+    // replace are here to convert the Base64-URL into Base64 which is what atob supports
+    // replace with //g regex acts like replaceAll
+    return atob(value.replace(/[-]/g, '+').replace(/[_]/g, '/'))
+  } catch (e) {
+    if (e instanceof ReferenceError) {
+      // running on nodejs < 16
+      // Buffer supports Base64-URL transparently
+      return Buffer.from(value, 'base64').toString('utf-8')
+    } else {
+      throw e
+    }
+  }
+}
+
 export default class GoTrueClient {
   /**
    * Namespace for the GoTrue API methods.
@@ -381,24 +399,97 @@ export default class GoTrueClient {
   }
 
   /**
+   * Sets the session from the provided session information. The access_token
+   * is reused if it is not expired, otherwise a new access token is fetched by
+   * refreshing the session with the provided refresh_token.
+   *
+   * This method is useful when using in a server-side rendered context.
+   *
+   * @param params.refresh_token A valid refresh token (typically obtained from a cookie)
+   * @param params.access_token An access token (typically obtained from a cookie)
+   */
+  async setSession(params: {
+    refresh_token: string
+    access_token: string
+  }): Promise<{ session: Session | null; error: ApiError | null }>
+
+  /**
    * Sets the session data from refresh_token and returns current Session and Error
    * @param refresh_token a JWT token
    */
   async setSession(
     refresh_token: string
+  ): Promise<{ session: Session | null; error: ApiError | null }>
+
+  async setSession(
+    arg0: string | { access_token: string; refresh_token: string }
   ): Promise<{ session: Session | null; error: ApiError | null }> {
-    try {
-      if (!refresh_token) {
-        throw new Error('No current session.')
-      }
+    let session: Session
+
+    if (typeof arg0 === 'string') {
+      // using the refresh_token string API
+      const refresh_token = arg0
+
       const { data, error } = await this.api.refreshAccessToken(refresh_token)
       if (error) {
         return { session: null, error: error }
       }
 
-      this._saveSession(data!)
+      session = data!
+    } else {
+      // using the object parameter API
+
+      const timeNow = Math.round(Date.now() / 1000)
+
+      let { refresh_token, access_token } = arg0
+      let expires_at = 0
+      let expires_in = 0
+
+      const tokenParts = access_token.split('.')
+      if (tokenParts.length !== 3) throw new Error('access_token is not a proper JWT')
+
+      const bodyJSON = decodeBase64URL(tokenParts[1])
+
+      let parsed: any = undefined
+      try {
+        parsed = JSON.parse(bodyJSON)
+      } catch (e) {
+        throw new Error('access_token is not a proper JWT, invalid JSON in body')
+      }
+
+      if (typeof parsed === 'object' && parsed && typeof parsed.exp === 'number') {
+        expires_at = parsed.exp
+        expires_in = timeNow - parsed.exp
+      } else {
+        throw new Error('access_token is not a proper JWT, missing exp claim')
+      }
+
+      if (timeNow > expires_at) {
+        const { data, error } = await this.api.refreshAccessToken(refresh_token)
+        if (error) {
+          return { session: null, error: error }
+        }
+
+        session = data!
+      } else {
+        const { user, error } = await this.api.getUser(access_token)
+        if (error) throw error
+
+        session = {
+          access_token,
+          expires_in,
+          expires_at,
+          refresh_token,
+          token_type: 'Bearer',
+          user: user!,
+        }
+      }
+    }
+
+    try {
+      this._saveSession(session)
       this._notifyAllSubscribers('SIGNED_IN')
-      return { session: data, error: null }
+      return { session, error: null }
     } catch (e) {
       return { error: e as ApiError, session: null }
     }
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -480,7 +480,7 @@ export default class GoTrueClient {
           expires_in,
           expires_at,
           refresh_token,
-          token_type: 'Bearer',
+          token_type: 'bearer',
           user: user!,
         }
       }
--- a/test/GoTrueClient.test.ts
+++ b/test/GoTrueClient.test.ts
@@ -16,7 +16,7 @@ describe('GoTrueClient', () => {
   })
 
   describe('Sessions', () => {
-    test('setSession should return no error', async () => {
+    test('setSession(refreshToken) should return no error', async () => {
       const { email, password } = mockUserCredentials()
 
       const { error, session } = await authWithSession.signUp({
@@ -26,7 +26,38 @@ describe('GoTrueClient', () => {
       expect(error).toBeNull()
       expect(session).not.toBeNull()
 
-      await authWithSession.setSession(session?.refresh_token as string)
+      const { error: sessionError } = await authWithSession.setSession(session!.refresh_token!)
+      expect(sessionError).toBeNull()
+
+      const { user } = await authWithSession.update({ data: { hello: 'world' } })
+
+      expect(user).not.toBeNull()
+      expect(user?.user_metadata).toStrictEqual({ hello: 'world' })
+    })
+
+    test('setSession({ access_token, refresh_token }) should return no error', async () => {
+      const { email, password } = mockUserCredentials()
+
+      const { error, session } = await authWithSession.signUp({
+        email,
+        password,
+      })
+      expect(error).toBeNull()
+      expect(session).not.toBeNull()
+
+      const { session: setSession, error: sessionError } = await authWithSession.setSession({
+        refresh_token: session!.refresh_token!,
+        access_token: session!.access_token!,
+      })
+      expect(sessionError).toBeNull()
+      expect(setSession).not.toBeNull()
+      expect(setSession!.user).not.toBeNull()
+      expect(setSession!.expires_in).not.toBeNull()
+      expect(setSession!.expires_at).not.toBeNull()
+      expect(setSession!.access_token).not.toBeNull()
+      expect(setSession!.refresh_token).not.toBeNull()
+      expect(setSession!.token_type).toStrictEqual('Bearer')
+
       const { user } = await authWithSession.update({ data: { hello: 'world' } })
 
       expect(user).not.toBeNull()


src/lib/version.ts
```diff
@@ -1,2 +1,2 @@
-// generated by genversion
+// Generated by genversion.
```
hf, line 1: Probably not supposed to be in the PR.

src/GoTrueClient.ts
```diff
@@ -380,25 +380,112 @@ export default class GoTrueClient {
...
+          refresh_token,
+          token_type: 'Bearer',
+          user: user!,
+        }
+      }
+    } else {
+      throw new Error('Invalid argument, must be string or object')
+    }
+
+    try {
```
kangmingtay, line 485: should we wrap everything above in a single try block and do the error handling in the catch?

src/GoTrueClient.ts
```diff
@@ -380,25 +380,112 @@ export default class GoTrueClient {
...
+   * @param params.refresh_token A valid refresh token (typically obtained from a cookie)
+   * @param params.access_token An access token (typically obtained from a cookie)
+   * @param params.provider_token A provider token (from social login, for example), optional
+   * @param params.expires_in A timeout in seconds when the access token will expire, if not provided taken from the `exp` claim in the access token
+   */
+  async setSession(params: {
+    refresh_token: string
+    access_token: string
+    expires_in?: number
+    provider_token?: string
```
kangmingtay, line 399: can we assign this to a type here so we don't have to check the types below (lines 424-429)?

src/GoTrueClient.ts
```diff
@@ -380,25 +380,112 @@ export default class GoTrueClient {
...
+          parsed = JSON.parse(bodyJSON)
+        } catch (e) {
+          throw new Error('access_token is not a proper JWT, invalid JSON in body')
+        }
+
+        if (typeof parsed === 'object' && parsed && typeof parsed.exp === 'number') {
+          expires_in = Math.max(0, parsed.exp - timeNow)
+          expires_at = parsed.exp
+        } else {
+          throw new Error('access_token is not a proper JWT, missing exp claim')
```
kangmingtay, line 456: note: this is ok for v1 but we decided to start returning a standard typed error object going forward for v2

src/GoTrueClient.ts
```diff
@@ -380,25 +380,112 @@ export default class GoTrueClient {
...
+        if (tokenParts.length !== 3) throw new Error('access_token is not a proper JWT')
+
+        const bodyJSON = atob(tokenParts[1].replaceAll(/[-]/g, '+').replaceAll(/[_]/g, '/'))
+
+        let parsed: any = undefined
+        try {
+          parsed = JSON.parse(bodyJSON)
+        } catch (e) {
+          throw new Error('access_token is not a proper JWT, invalid JSON in body')
+        }
```
kangmingtay, line 450: i feel like we should just leave this validation to gotrue and not include it in the client lib

src/GoTrueClient.ts
```diff
@@ -380,25 +380,112 @@ export default class GoTrueClient {
...
+      (typeof arg0.provider_token === 'string' || !arg0.provider_token)
+    ) {
+      // using the object parameter API
+      let { provider_token, refresh_token, access_token, expires_in } = arg0
+
+      const timeNow = Math.round(Date.now() / 1000)
+      let expires_at = 0
+
+      if (expires_in) {
+        expires_at = timeNow + expires_in
```
kangmingtay, line 438: why should we allow passing in an `expires_in` time - seems like this is never validated against the `exp` claim in the jwt

kangmingtay (changes requested):
I think we're trying to do too much here and i'm not sure what this PR is trying to change - the current method just requires a refresh token, forces a refresh and then sets the current session. 

Why do we also want to support setting a session via an access token? If no refresh token is being passed in, it could result in a incomplete session returned.

hf:
Here's a Mermaid description of the issue:

```mermaid
sequenceDiagram
    participant G as GoTrue
    participant S as Server
    participant B as Browser
    participant C as Cookies
    participant L as Local Storage
    
    B-->>S: GET /login
    S-->>B: 200 OK Render /login

    B-->>G: POST /sign-in { email, password }
    G-->>B: 303 See Other server/authenticate...

    B-->>S: GET /authenticate
    S-->>B: 200 OK Render /authenticate

    activate B

        B-->>B: getSessionFromURL
        B-->>L: putItem(session, ...)
        B-->>C: document.cookie = 'my-access-token...'
        B-->>C: document.cookie = 'my-refresh-token...'

    deactivate B

    note over G, L: After 1 week the user comes back to /route on server -- no issues

    activate B
        B-->>S: GET /route
        B-->>S: Cookie: my-access-token ...
        B-->>S: Cookie: my-refresh-token ...

        activate S
            S-->>S: setSession(req.cookies['my-refresh-token'])
            S-->>G: POST /verify?grant_type=refresh_token
            G-->>S: 200 OK access token, refresh token
            S-->>S: return session information to app

            S-->>B: 200 OK prerendered content
        deactivate S

        B-->>B: _refreshSession
        B-->>L: getItem(session)
        L-->>B: session { access_token, refresh_token, expires_at }
        B-->>G: POST /verify?grant_type=refresh_token
        G-->>B: 200 OK access token, refresh token 
        B-->>L: putItem(session)
    deactivate B
    

note over G, L: Immediately after login user navigates to /route -- issue after 1 hour

   activate B
        B-->>S: GET /route
        B-->>S: Cookie: my-access-token ...
        B-->>S: Cookie: my-refresh-token ...

        activate S
            S-->>S: setSession(req.cookies['my-refresh-token'])
            S-->>G: POST /token?grant_type=refresh_token
            G-->>S: 200 OK access token, refresh token
            S-->>S: return session information to app

            note over B, S: Start reuse timer!

            S-->>B: 200 OK prerendered content
        deactivate S

        B-->>B: _refreshSession
        B-->>L: getItem(session)
        L-->>B: session { access_token, refresh_token, expires_at }

        note over L, B: now is before expires_at!
        note over L, B: No need to renew refresh token

        note over G, L: 1 hour passes, app notices expires_at and tries to refresh token
        B-->>G: POST /token?grant_type=refresh_token
        G-->>B: 401 Unauthorized refresh token already used

        note over G, B: 1 hour passed after the first validation of the refresh token!

    deactivate B

note over G, L: With setSession({ access_token, refresh_token }) fix -- no-issue

   activate B
        B-->>S: GET /route
        B-->>S: Cookie: my-access-token ...
        B-->>S: Cookie: my-refresh-token ...

        activate S
            S-->>S: setSession({ refresh_token: req.cookies['my-refresh-token']), access_token: ... })
            S-->>S: access token is not expired no post to GoTrue

            S-->>S: return session information to app

            note over B, S: Start reuse timer!

            S-->>B: 200 OK prerendered content
        deactivate S

        B-->>B: _refreshSession
        B-->>L: getItem(session)
        L-->>B: session { access_token, refresh_token, expires_at }

        note over L, B: now is before expires_at!
        note over L, B: No need to renew refresh token

        note over G, L: 1 hour passes, app notices expires_at and tries to refresh token
        B-->>G: POST /token?grant_type=refresh_token
        G-->>B: 200 OK access_token, refresh_token

        note over G, B: 1 hour passed after the first validation of the refresh token!

        B-->>B: getSessionFromURL
        B-->>L: putItem(session, ...)
        B-->>C: document.cookie = 'my-access-token...'
        B-->>C: document.cookie = 'my-refresh-token...'

        note over L, G: Ok but what if the user closes the tab until 1 hour passes and then navigates to /route

        B-->>S: GET /route
        B-->>S: Cookie: my-access-token ...
        B-->>S: Cookie: my-refresh-token ...

        activate S
            S-->>S: setSession({ refresh_token: req.cookies['my-refresh-token']), access_token: ... })
            S-->>S: access token IS expired, refreshing in GoTrue
            S-->>G: POST /token?grant_type=refresh_token
            G-->>S: 200 OK access token, refresh token
            S-->>S: return session information to app

            note over B, S: Start reuse timer!

            S-->>B: 200 OK prerendered content
        deactivate S

        B-->>B: _refreshSession
        B-->>L: getItem(session)
        L-->>B: session { access_token, refresh_token, expires_at }
        B-->>B: now is after expires_at!
        B-->>G: POST /token?grant_type=refresh_token
        note over B, G: this occurs milliseconds after the refresh from the server
        note over B, G: thus the refresh_token although used is within a reuse period
        G-->>B: 200 OK access_token, refresh_token

        B-->>B: getSessionFromURL
        B-->>L: putItem(session, ...)
        B-->>C: document.cookie = 'my-access-token...'
        B-->>C: document.cookie = 'my-refresh-token...'

        note over G, L: All is good in the world!

    deactivate B

```

src/GoTrueClient.ts
hf: Nope, that's unfortunately a limitation with TypeScript and function overloading as described here: https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads

Tho it may be possible to do away a lot of the `typeof` statements (but not all) with some type changes on `arg0`.

src/GoTrueClient.ts
hf: OK but this is not a typed error -- this is literally programmer error where they are passing some invalid string as `access_token`.

src/GoTrueClient.ts
hf: We can't leave this to GoTrue, as the library is trying to figure out if it should be calling the refresh token endpoint. Remember, once that's called the refresh token is "invalidated."

src/GoTrueClient.ts
hf: It is validated later on line 406:

```typescript
      if (timeNow > expires_at) {
```

kangmingtay: isn't `timeNow` just the current time and not the `exp` claim in the jwt? if `expires_in` is passed, then `timeNow` will remain as the current time since line 453 never happens

src/GoTrueClient.ts
```diff
@@ -380,25 +380,112 @@ export default class GoTrueClient {
...
   async setSession(
     refresh_token: string
-  ): Promise<{ session: Session | null; error: ApiError | null }> {
-    try {
-      if (!refresh_token) {
-        throw new Error('No current session.')
-      }
+  ): Promise<{ session: Session | null; error: ApiError | null }>
+
+  async setSession(arg0: any): Promise<{ session: Session | null; error: ApiError | null }> {
```
kangmingtay, line 410: we can make the types tighter here so it's easier to do the type-checking later in the code
```suggestion
  async setSession(arg0: string | Session): Promise<{ session: Session | null; error: ApiError | null }> {
```

src/GoTrueClient.ts
```diff
@@ -380,25 +380,112 @@ export default class GoTrueClient {
...
 
-      this._saveSession(data!)
+      session = data!
+    } else if (
+      typeof arg0 === 'object' &&
+      arg0 &&
+      typeof arg0.refresh_token === 'string' &&
+      typeof arg0.access_token === 'string' &&
+      (typeof arg0.expires_in === 'number' || !arg0.expires_in) &&
+      (typeof arg0.provider_token === 'string' || !arg0.provider_token)
```
kangmingtay, line 429: we can do something similar for v1 by introducing a private `_isValidSession` function to do this so it's clearer 

referenced in v2: https://github.com/supabase/gotrue-js/blob/477d5869077ad5ca11269d3d3e3b1b64d2601b9c/src/GoTrueClient.ts#L737-L746

src/GoTrueClient.ts
```diff
@@ -380,25 +380,112 @@ export default class GoTrueClient {
...
+      const timeNow = Math.round(Date.now() / 1000)
+      let expires_at = 0
+
+      if (expires_in) {
+        expires_at = timeNow + expires_in
+      } else {
+        const tokenParts = access_token.split('.')
+        if (tokenParts.length !== 3) throw new Error('access_token is not a proper JWT')
+
+        const bodyJSON = atob(tokenParts[1].replaceAll(/[-]/g, '+').replaceAll(/[_]/g, '/'))
```
kangmingtay, line 443: are the regex checks necessary? also, `atob` is deprecated, lets use this instead:

```js
const payload = JSON.parse(
          Buffer.from(currentSession.access_token.split('.')[1], 'base64').toString()
        )
```

kangmingtay (changes requested):

src/GoTrueClient.ts
hf: There's no native Base64 support on the browser except `btoa` / `atob`. The `replaceAll` converts the Base64-URL encoding in the access token into a Base64 which is what `btoa` supports.

src/GoTrueClient.ts
hf: Yes.

src/GoTrueClient.ts
hf: Probably not `Session` but I will change into this yes, nice catch.

kangmingtay (approved):

kangmingtay (approved):

src/GoTrueClient.ts
hf: Actually TypeScript is smart enough not to need this if I change the type of `arg0: string | { access_token: string; refresh_token: string; }`.

src/GoTrueClient.ts
```diff
@@ -380,25 +398,98 @@ export default class GoTrueClient {
...
+      } else {
+        const { user, error } = await this.api.getUser(access_token)
+        if (error) throw error
+
+        session = {
+          access_token,
+          expires_in,
+          expires_at,
+          refresh_token,
+          token_type: 'Bearer',
```
kangmingtay, line 483:
```suggestion
          token_type: 'bearer',
```

kangmingtay: gotrue returns the `token_type` as bearer in lowercase

PR #446, opened 2022-09-21
Add return type for /challenge and introduces getAAL and getAMR

J0:
- Adds the corresponding return type for the /challenge end point, which previously wasn't included
- Introduces `getAMR` and `getAAL` methods

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -25,6 +25,7 @@ import {
   resolveFetch,
   setItemAsync,
   uuid,
+  decodeJWT,
 } from './lib/helpers'
 import localStorageAdapter from './lib/local-storage'
 import { polyfillGlobalThis } from './lib/polyfills'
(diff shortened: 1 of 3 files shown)


hf (approved):

kangmingtay (approved):

src/GoTrueClient.ts
```diff
@@ -1104,4 +1108,57 @@ export default class GoTrueClient {
...
+   */
+  private async _getAMR(jwt?: string) {
+    try {
+      if (!jwt) {
+        const { data, error } = await this.getSession()
+        if (error) {
+          throw error
+        }
+        // Default to Authorization header if there is no existing session
+        jwt = data.session?.access_token ?? this.headers['Authorization']
```
kangmingtay, line 1124:
```suggestion
        jwt = data.session?.access_token ?? undefined
```

src/GoTrueClient.ts
```diff
@@ -1104,4 +1108,57 @@ export default class GoTrueClient {
...
+   */
+  private async _getAAL(jwt?: string) {
+    try {
+      if (!jwt) {
+        const { data, error } = await this.getSession()
+        if (error) {
+          throw error
+        }
+        // Default to Authorization header if there is no existing session
+        jwt = data.session?.access_token ?? this.headers['Authorization']
```
kangmingtay, line 1151: ditto

PR #461, opened 2022-09-30
fix: refresh token sync across multiple tabs

### LINKED ISSUE #454: Getting 400 on /auth/v1/token?grant_type=refresh_token
# Bug report

## Describe the bug

I'm having an issue where `/auth/v1/token?grant_type=refresh_token` suddenly returns 400. It's unclear what could cause this issue. When the token is approaching expiration supabase-js fires correctly, but it gets to a point where the 400 request occurs. 

This originated after finding out that the supabase token disappears from localStorage on my extension.
The video walks through the situation in more detail: https://share.cleanshot.com/mc6hDzOEoD7lzwk4ZSrk

I've read that supabase v2 fixes "getting logged out” issues, but I'm not so sure this happens from the javascript side. To be precise, I'm initializing the client without the `multiTab` option, so it's not the race condition mentioned throughout the discussions/issues on Github.
Also, the Network logs, are isolated to one process. Meaning only one supabase client. Inside this process, each request uses the previous request's refresh token from the response. Being that it's isolated to only one process, I'm not sure how this could be a library issue. I could be wrong though

This is how I'm creating the supabase client.

```
const browserStorageInterface: SupportedStorage = {
  async getItem(key: string): Promise<string | null> {
    const storage = await browser.storage.local.get(key);
    return storage?.[key];
  },
  setItem(key: string, value: string) {
    browser.storage.local.set({
      [key]: value,
    });
  },
  removeItem(key: string) {
    browser.storage.local.remove(key);
  },
};

const supabase = createClient(supabaseUrl, supabaseKey, {
  autoRefreshToken: true,
  persistSession: true,
  localStorage: browserStorageInterface,
});
```

## To Reproduce

1. Setup supabase client
2. Authenticate using Google oAuth
3. Set JWT expiry to 20 seconds (to reproduce quickly)
4. supabase refreshes token several times
5. Eventually errors with status code 400.
6. Token is removed from localStorage

## Expected behavior

At no point should refreshing the token return status code 400. 

## Screenshots

https://share.cleanshot.com/mc6hDzOEoD7lzwk4ZSrk

## System information

- OS: macOS
- Browser: chrome - browser extension
- Version of supabase-js: v1.35.3
- Version of Node.js: v12.17.0

## Additional context

Help post on Discord: https://discord.com/channels/839993398554656828/1019328795129421944/1019328795129421944

kangmingtay:
* `_startAutoRefreshToken` used to take in the current session which might go stale by the time the `setTImeout` callback is invoked
* Change `_startAutoRefreshToken` to always get the current session from storage / memory instead of relying on the `session` passed in 
* Fixes #454, #442, #444
* Tested with the following scenarios:
  * `JWT_EXPIRY`: 30s, `REUSE_INTERVAL`: 10s, 5 tabs opened (firefox) 
  * `JWT_EXPIRY`: 20s, `REUSE_INTERVAL`: 5s, 5 tabs opened (firefox)

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -875,7 +875,7 @@ export default class GoTrueClient {
       const timeNow = Math.round(Date.now() / 1000)
       const expiresIn = expiresAt - timeNow
       const refreshDurationBeforeExpires = expiresIn > EXPIRY_MARGIN ? EXPIRY_MARGIN : 0.5
-      this._startAutoRefreshToken((expiresIn - refreshDurationBeforeExpires) * 1000, session)
+      this._startAutoRefreshToken((expiresIn - refreshDurationBeforeExpires) * 1000)
     }
 
     if (this.persistSession && session.expires_at) {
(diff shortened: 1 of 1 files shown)


src/GoTrueClient.ts
```diff
@@ -904,22 +904,25 @@ export default class GoTrueClient {
...
+        error: sessionError,
+      } = await this.getSession()
+      if (!sessionError && session) {
+        const { error } = await this._callRefreshToken(session.refresh_token)
+        if (!error) this.networkRetries = 0
+        if (
+          error instanceof AuthRetryableFetchError &&
+          this.networkRetries < NETWORK_FAILURE.MAX_RETRIES
+        )
+          this._startAutoRefreshToken(NETWORK_FAILURE.RETRY_INTERVAL ** this.networkRetries * 100) // exponential backoff
```
kangmingtay, line 924: this seems error prone too - in the event of network failure, the retry interval could possibly fall outside of the reuse interval error

alaister: Yeah, I agree with you that the user should be logged out here. We can always take another look at this if it becomes a bigger issue in the future.

alaister (approved):
lgtm!

src/GoTrueClient.ts
miguelespinoza: @kangmingtay, @alaister. A robust approach could be to listen for a network status change.
https://developer.mozilla.org/en-US/docs/Web/API/Navigator/onLine

This could affect users who are traveling and have their laptops offline for a day. Ideally, you'd want to stay logged in, it's an edge case, but a listener would capture these events. Then you can avoid long polling.

Issue #238, opened 2022-02-28
Transient error missing


theo-m:
# Bug report

## Describe the bug

This code is problematic because it surfaces the wrong error as `getUserError` is never properly used or reported - the error misleads the user into thinking it's a cookie issue when it can just be a regular transient error. In my case it was bad sockets acting up.

https://github.com/supabase/gotrue-js/blob/e440a90099b5c138ac841348e1f1bb5b02f58608/src/GoTrueApi.ts#L682-L686

PR #463, opened 2022-09-30
fix: log error in `getUserByCookie`

hf:
If an error occurs in `GoTrueApi#getUserByCookie` when calling `getUser`, the actual root cause will be masked, which may confuse developers as to what may be going on.

See: #238.


kangmingtay (approved):

hf:
No longer applies.

PR #465, opened 2022-09-30
refactor: use named metadata types on User (next)

hf:
By using named types for `app_metadata` and `user_metadata` developers can use [TypeScript declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) to enforce some form of type checking on these fields.

See: #241.

(Same as #464 but for next branch.)


kangmingtay (approved):
can you change this to go into the `rc` branch instead of `next` please?

PR #467, opened 2022-10-04
feat: update setSession

kangmingtay:
* Fixes `setSession` to make it safe to use in a server-side or SSR context
* `setSession` should take in an `access token` and `refresh token` and only perform a refresh if the `access_token` is expired. We derive the expiry from the `exp` claim in the `access_token`.  
* This is a breaking change because `setSession` now takes in an object.
* References #445 but made some changes to make it compatible for v2

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -535,14 +535,18 @@ export default class GoTrueClient {
    * If the current session does not contain at expires_at field, setSession will use the exp claim defined in the access token.
    * @param currentSession The current session that minimally contains an access token, refresh token and a user.
    */
-  async setSession(refresh_token: string, access_token = ''): Promise<AuthResponse> {
+  async setSession(
+    currentSession: Pick<Session, 'access_token' | 'refresh_token'>
+  ): Promise<AuthResponse> {
     try {
       const timeNow = Date.now() / 1000
       let expiresAt = timeNow
       let hasExpired = true
       let session: Session | null = null
-      if (access_token && access_token.split('.')[1]) {
-        const payload = JSON.parse(Buffer.from(access_token.split('.')[1], 'base64').toString())
+      if (currentSession.access_token && currentSession.access_token.split('.')[1]) {
+        const payload = JSON.parse(
+          Buffer.from(currentSession.access_token.split('.')[1], 'base64').toString()
+        )
         if (payload.exp) {
           expiresAt = payload.exp
           hasExpired = expiresAt <= timeNow
@@ -550,10 +554,10 @@ export default class GoTrueClient {
       }
 
       if (hasExpired) {
-        if (!refresh_token) {
+        if (!currentSession.refresh_token) {
           throw new AuthSessionMissingError()
         }
-        const { data, error } = await this._refreshAccessToken(refresh_token)
+        const { data, error } = await this._refreshAccessToken(currentSession.refresh_token)
         if (error) {
           return { data: { session: null, user: null }, error: error }
         }
@@ -563,13 +567,13 @@ export default class GoTrueClient {
         }
         session = data.session
       } else {
-        const { data, error } = await this.getUser(access_token)
+        const { data, error } = await this.getUser(currentSession.access_token)
         if (error) {
           throw error
         }
         session = {
-          access_token: access_token,
-          refresh_token: refresh_token,
+          access_token: currentSession.access_token,
+          refresh_token: currentSession.refresh_token,
           user: data.user,
           token_type: 'bearer',
           expires_in: expiresAt - timeNow,
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -17,6 +17,7 @@ import {
 } from './lib/errors'
 import { Fetch, _request, _sessionResponse, _userResponse } from './lib/fetch'
 import {
+  decodeBase64URL,
   Deferred,
   getItemAsync,
   getParameterByName,
@@ -544,14 +545,7 @@ export default class GoTrueClient {
       let hasExpired = true
       let session: Session | null = null
       if (currentSession.access_token && currentSession.access_token.split('.')[1]) {
-        const payload = JSON.parse(
-          atob(
-            currentSession.access_token
-              .split('.')[1]
-              .replaceAll(/[-]/g, '+')
-              .replaceAll(/[_]/g, '/')
-          )
-        )
+        const payload = JSON.parse(decodeBase64URL(currentSession.access_token.split('.')[1]))
         if (payload.exp) {
           expiresAt = payload.exp
           hasExpired = expiresAt <= timeNow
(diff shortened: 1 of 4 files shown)


src/GoTrueClient.ts
```diff
@@ -530,28 +530,45 @@ export default class GoTrueClient {
...
   /**
-   * Sets the session data from refresh token and returns current session or an error if the refresh token is invalid.
-   * @param refresh_token A refresh token returned by supabase auth.
+   * Sets the session data from the current session. If the current session is expired, setSession will take care of refreshing it to obtain a new session.
+   * If the refresh token in the current session is invalid and the current session has expired, an error will be thrown.
+   * If the current session does not contain at expires_at field, setSession will use the exp claim defined in the access token.
+   * @param currentSession The current session that minimally contains an access token, refresh token and a user.
    */
-  async setSession(refresh_token: string): Promise<AuthResponse> {
+  async setSession(currentSession: Session): Promise<AuthResponse> {
```
thorwebdev, line 538: This should work with just an `access_token` and a `refresh_token`, no? With it being typed as `Session` typescript would complain, no?

kangmingtay: hmm why would typescript complain? we defined the `Session` type as an interface already and i think it feels slightly more intuitive that `setSession` takes in a session object

thorwebdev: The session interface is 

```js
export interface Session {
  provider_token?: string | null
  /**
   * The access token jwt. It is recommended to set the JWT_EXPIRY to a shorter expiry value.
   */
  access_token: string
  /**
   * A one-time used refresh token that never expires.
   */
  refresh_token: string
  /**
   * The number of seconds until the token expires (since it was issued). Returned when a login is confirmed.
   */
  expires_in: number
  /**
   * A timestamp of when the token will expire. Returned when a login is confirmed.
   */
  expires_at?: number
  token_type: string
  user: User
}
```

Meaning `access_token`, `refresh_token`, `expires_in`, and `user` will be required to be provided. However the code itself doesn't require them. 

So I think we should change it to 

```js
async setSession(currentSession: Pick<Session, 'access_token' | 'refresh_token'> & Partial<Session>): Promise<AuthResponse>
```

This should allow us to provide an object that has only `access_token` & `refresh_token` as well as a full session object. Wdyt?

kangmingtay: Hmm is there any reason why the user wouldn't want to pass in a full session? i feel like this makes things clearer on the server-side because you'll always end up with a complete session. 

In v2, the `Session` type was made tighter to make it clear that one can always expect a `refresh_token` and a `user` unlike in v1. This also makes it easy because the user can just do `getSession` and pass the session returned into `setSession` immediately. 

Also, if we change the type to `Pick<Session, 'access_token' | 'refresh_token'> & Partial<Session>`, then we need to make an extra call inside `setSession` to get the user object using the `access_token` (for the case that the `access_token` hasn't expired) before we can save the session.

src/GoTrueClient.ts
```diff
@@ -530,28 +530,57 @@ export default class GoTrueClient {
...
   /**
-   * Sets the session data from refresh token and returns current session or an error if the refresh token is invalid.
-   * @param refresh_token A refresh token returned by supabase auth.
+   * Sets the session data from the current session. If the current session is expired, setSession will take care of refreshing it to obtain a new session.
+   * If the refresh token in the current session is invalid and the current session has expired, an error will be thrown.
+   * If the current session does not contain at expires_at field, setSession will use the exp claim defined in the access token.
+   * @param currentSession The current session that minimally contains an access token, refresh token and a user.
    */
-  async setSession(refresh_token: string): Promise<AuthResponse> {
+  async setSession(refresh_token: string, access_token = ''): Promise<AuthResponse> {
```
kangmingtay, line 538: @thorwebdev i decided to default the `access_token` here to an empty string so this wouldn't be a breaking change

thorwebdev: @kangmingtay I'd prefer making the breaking change now and changing the argument to an object, which makes it easier to make changes in the future. Wdyt?

hf: Yes I agree with @thorwebdev. An object argument is easier to expand on later in a backward compatible way...

src/GoTrueClient.ts
```diff
@@ -530,28 +530,57 @@ export default class GoTrueClient {
...
-      }
-      const { data, error } = await this._refreshAccessToken(refresh_token)
-      if (error) {
-        return { data: { session: null, user: null }, error: error }
+      const timeNow = Date.now() / 1000
+      let expiresAt = timeNow
+      let hasExpired = true
+      let session: Session | null = null
+      if (access_token && access_token.split('.')[1]) {
+        const payload = JSON.parse(Buffer.from(access_token.split('.')[1], 'base64').toString())
```
hf, line 545: Buffer does not exist on the browser. The only way to parse Base64 on the browser is [btoa](https://developer.mozilla.org/en-US/docs/Web/API/btoa). You can use the code from the v1 PR.

hf (changes requested):

src/GoTrueClient.ts
hf: Also note that `btoa` does not accept Base64-URL, which is what JWT uses, so you need to replace the characters. You don't need to add padding (`=`) though.

src/GoTrueClient.ts
kangmingtay: yup sounds good, will make the change!

thorwebdev (approved):
LGTM

thorwebdev:
Once question, have we tested that this works in V8 runtime which many edge functions (vercel middleware; cloudflare workers) etc use. I remember there were potentially some issues with that, e.g. https://github.com/supabase/auth-helpers/issues/97#issuecomment-1128410368

kangmingtay:
> have we tested that this works in V8 runtime which many edge functions (vercel middleware; cloudflare workers) etc use

we haven't, but the [catch block](https://github.com/supabase/gotrue-js/pull/467/files#diff-26999e6f9a7fac03a622245890c5bb7f2d49a102fc8b4aaed94001b7be8d7fb3R82-R94) should account for those cases if `atob` throws an error

hf (approved):
Looks good!

rphlmr:
Hello there,

If I want to force a refresh, are we ok that passing an "empty string" for the access token is the way?

mikitasolo:
as described in https://github.com/supabase/gotrue-js/issues/476 using `setSession` in react-native fails with `Property 'Buffer' doesn't exist` error, as in RN there's no `atob` and `Buffer`. any thoughts on workaround?

lewisd1996:
I second @mikitasolo's comment, could really do with some guidance with implementing this into react native.

mikitasolo:
@lewisd1996 i've added 
```
import { decode, encode } from 'base-64';

if (!global.btoa) {
  global.btoa = encode;
}

if (!global.atob) {
  global.atob = decode;
}
```
to my index.js as a workaround

lewisd1996:
Amazing! works perfectly, nice workaround. Thank you! 🙏

PR #472, opened 2022-10-05
refactor: add types, routes, missing implementation, methods

hf:


src/GoTrueClient.ts
```diff
@@ -139,6 +146,7 @@ export default class GoTrueClient {
       listFactors: this._listFactors.bind(this),
       getAMR: this._getAMR.bind(this),
       getAAL: this._getAAL.bind(this),
```
J0, line 148: Guess we can remove this since we have `getAuthenticatorAssuranceLevel`

src/lib/types.ts
```diff
@@ -523,9 +523,9 @@ export type GenerateLinkType =
 
 // MFA related types
 export type MFAEnrollParams = {
-  friendlyName?: string
   factorType: 'TOTP'
```
J0, line 526:
```suggestion
  factorType: 'totp'
```

Late update on my end but Server side now takes `totp` as a lowercased values so we can probably do the same in the lib

src/GoTrueClient.ts
```diff
@@ -1017,96 +1008,98 @@ export default class GoTrueClient {
...
-    return { data: { factors: data.user?.factors ?? [] }, error: error }
+  private async _listFactors(): Promise<AuthMFAListFactorsResponse> {
+    const { data, error } = await this.getSession()
+    if (error) {
+      return { data: null, error }
+    }
+
+    const all = data?.session?.user?.factors || []
+    const TOTP = all.filter(
+      (factor) => 'TOTP' === factor.factor_type && 'verified' === factor.status
```
J0, line 1068:
```suggestion
      (factor) => 'totp' === factor.factor_type && 'verified' === factor.status
```

src/lib/types.ts
```diff
@@ -552,18 +552,102 @@ export type DeleteFactorParams = {
...
+  | {
+      data: null
+      error: AuthError
+    }
+
+export type AuthMFAEnrollResponse =
+  | {
+      data: {
+        id: string
+        type: 'TOTP'
```
J0, line 575:
```suggestion
        type: 'totp'
```

J0 (approved):
Looks good to me!

package.json
```diff
@@ -31,7 +31,7 @@
     "test": "run-s test:clean test:infra test:suite test:clean",
     "test:suite": "jest --runInBand",
     "test:infra": "cd infra && docker-compose down && docker-compose pull && docker-compose up -d && sleep 30",
-    "test:clean": "cd infra && docker-compose down --remove-orphans",
+    "test:clean": "cd infra && docker-compose down",
```
hf, line 34: Doesn't work with podman docker.

src/lib/types.ts
```diff
@@ -166,12 +122,25 @@ export interface Session {
...
+ * An entry designates what method was used by the user to verify their
+ * identity and at what time.
+ *
+ * @see {@link GoTrueMFAApi#getAuthenticatorAssuranceLevel}.
+ */
 export interface AMREntry {
-  authentication_method: string
-  timestamp: string
+  /** Authentication method name. */
+  method: string // TODO: explicit method names / types
```
hf, line 137: @J0 Once we finalize the AMR methods, this needs to be changed to include them explicitly.

src/GoTrueClient.ts
```diff
@@ -1026,28 +1052,10 @@ export default class GoTrueClient {
...
-    } catch (error) {
-      if (isAuthError(error)) {
-        return { data: null, error }
-      }
-
-      throw error
-    }
+  private async _unenroll(params: MFAUnenrollParams): Promise<AuthMFAUnenrollResponse> {
+    return await _request(this.fetch, 'DELETE', `${this.url}/factor/${params.factorId}`, {
+      headers: this.headers,
```
kangmingtay, line 1057: don't we still need to pass in the code if we want to unenroll a verified factor?
```suggestion
      headers: this.headers,
      body: { code: params.code },
```

kangmingtay (changes requested):

src/GoTrueClient.ts
hf: Ideally not, see the discussion on Slack

J0: wait, I think as of right now we still need to pass in a code if we want to unenroll a verified factor. Should we be removing that? But let's continue on Slack yes

src/lib/helpers.ts
```diff
@@ -121,17 +121,13 @@ export class Deferred<T = any> {
...
-  const base64 = base64Url.replace(/-/g, '+').replace(/_/g, '/')
-  const jsonPayload = decodeURIComponent(
-    atob(base64)
-      .split('')
-      .map(function (c) {
-        return '%' + ('00' + c.charCodeAt(0).toString(16)).slice(-2)
-      })
-      .join('')
-  )
+export function decodeJWTPayload(token: string) {
```
j4w8n, line 124: Any idea when this will be merged? I'm working on a pull request for something else, and was gonna use a similar decode function. The code I pulled this morning didn't have one, so I was gonna add it. But this PR negates the need.

j4w8n: Also, I was gonna use this function to verify that a passed-in `access_token` was quasi-valid. In other words, folks couldn't just pass an empty string or `hello.there.world` into the method I planned on adding.

So if you wanted to be slightly more paranoid, you could do this:
```js
const base64regex = /^([0-9a-zA-Z+/]{4})*(([0-9a-zA-Z+/]{2}==)|([0-9a-zA-Z+/]{3}=))?$/
const parts = token.split('.')

if (parts.length !== 3) {
  throw new Error('JWT is not valid: not a JWT structure')
}

if (!base64regex.test(parts[1])) {
  throw new Error('JWT is not valid: not a JWT structure')
}
```

src/GoTrueClient.ts
```diff
@@ -1056,148 +1062,121 @@ export default class GoTrueClient {
...
+        jwt: sessionData?.session?.access_token,
+      }
+    )
+    if (error) {
+      return { data: null, error }
     }
+
+    await this._saveSession({
+      expires_at: Math.round(Date.now() / 1000) + data.expires_in,
+      ...data,
```
hf, line 1109: `_saveSession` would not save it unless it has an `expires_at`, this is the best I could come up with. Please double check @kangmingtay @J0.

src/lib/helpers.ts
hf: Thanks! Feel free to pull this out in another PR, and we'll merge it!

hf: Also the regular expression is somewhat wrong as JWTs use Base64-URL. This is a correct regular expression:

```javascript
/^([a-z0-9_-]{4})*($|[a-z0-9_-]{3}=?$|[a-z0-9_-]{2}(==)?$)$/i
```

src/GoTrueClient.ts
```diff
@@ -1056,148 +1062,121 @@ export default class GoTrueClient {
...
    * Displays all devices for a given user
    */
-  private async _listFactors() {
-    const { data, error } = await this.getUser()
-    if (error) throw error
-    return { data: { factors: data.user?.factors ?? [] }, error: error }
+  private async _listFactors(): Promise<AuthMFAListFactorsResponse> {
+    const { data: sessionData, error: sessionError } = await this.getSession()
+    if (sessionError) {
+      return { data: null, error: sessionError }
```
kangmingtay, line 1138:
```suggestion
      return { data: { all: [], totp: [] }, error: sessionError }
```

kangmingtay: slight improvement in DX here because data will always be an object with those keys instead of object | null

PR #473, opened 2022-10-08
fix: Add method setSessionFromRefreshToken

Aerilym:

The setSession method doesn't work with a refresh token, so I made a function using much of a same code that sets the session using a refresh token.

Currently, there is no way to set the session using a refresh token, as was possible in previous versions and documented here: https://supabase.com/docs/reference/javascript/next/auth-setsession

## What is the new behavior?

The session can be set using a refresh token.


Aerilym:
I created this as its own method so I could use it with supabase-js. However, it would make more sense to fix the setSession() method so it can set a session using the refresh token. If you want that done, I can do it.
This is the broken method:
```TypeScript
/**
   * Sets the session data from the current session. If the current session is expired, setSession will take care of refreshing it to obtain a new session.
   * If the refresh token in the current session is invalid and the current session has expired, an error will be thrown.
   * If the current session does not contain at expires_at field, setSession will use the exp claim defined in the access token.
   * @param currentSession The current session that minimally contains an access token, refresh token and a user.
   */
  async setSession(
    currentSession: Pick<Session, 'access_token' | 'refresh_token'>
  ): Promise<AuthResponse> {
    try {
      const timeNow = Date.now() / 1000
      let expiresAt = timeNow
      let hasExpired = true
      let session: Session | null = null
      if (currentSession.access_token && currentSession.access_token.split('.')[1]) {
        const payload = JSON.parse(decodeBase64URL(currentSession.access_token.split('.')[1]))
        if (payload.exp) {
          expiresAt = payload.exp
          hasExpired = expiresAt <= timeNow
        }
      }

      if (hasExpired) {
        if (!currentSession.refresh_token) {
          throw new AuthSessionMissingError()
        }
        const { data, error } = await this._refreshAccessToken(currentSession.refresh_token)
        if (error) {
          return { data: { session: null, user: null }, error: error }
        }

        if (!data.session) {
          return { data: { session: null, user: null }, error: null }
        }
        session = data.session
      } else {
        const { data, error } = await this.getUser(currentSession.access_token)
        if (error) {
          throw error
        }
        session = {
          access_token: currentSession.access_token,
          refresh_token: currentSession.refresh_token,
          user: data.user,
          token_type: 'bearer',
          expires_in: expiresAt - timeNow,
          expires_at: expiresAt,
        }
      }

      await this._saveSession(session)
      this._notifyAllSubscribers('TOKEN_REFRESHED', session)

      return { data: { session, user: session.user }, error: null }
    } catch (error) {
      if (isAuthError(error)) {
        return { data: { session: null, user: null }, error }
      }

      throw error
    }
  }
```
It seems it was made only to take a Session object, and contains a lot of unused code, as 'access_token' and 'refresh_token' can't be used.

hf:
Hey @Aerilym unfortunately you can't use `setSession` with a refresh token only. This was fixed in #445. 

An outline of the reasons:

- A refresh token alone does not represent a user's session, but the combination of access and refresh token do. 
- If you use only the refresh token, your users will experience random logouts because a refresh token is designed to be used only once.
- You will be filling up the database with refresh tokens each time you want to validate it, which can become expensive both from a data storage perspective and from a performance standpoint.

Can I inquire about your use case for using only the refresh token?

Aerilym:
> Hey @Aerilym unfortunately you can't use `setSession` with a refresh token only. This was fixed in #445.
> 
> An outline of the reasons:
> 
> * A refresh token alone does not represent a user's session, but the combination of access and refresh token do.
> * If you use only the refresh token, your users will experience random logouts because a refresh token is designed to be used only once.
> * You will be filling up the database with refresh tokens each time you want to validate it, which can become expensive both from a data storage perspective and from a performance standpoint.
> 
> Can I inquire about your use case for using only the refresh token?

The issue came up when I was trying to sign in with an oauth endpoint using react native and expo. I followed [this guide](https://dev.to/dancurtis/how-to-use-supabase-apple-oauth-in-react-native-4c4h), and it worked on previous versions of supabase-js. Also according to your [v2 documentation about setSession()](https://supabase.com/docs/reference/javascript/auth-setsession), this is still a valid way of using setSession().

Is there another way I should be going about logging in on expo?

Here is specifically how I'm using the new function in the same way as the example and old setSession() usage:
```TypeScript
async function handleLogin(provider: Provider) {
    if (Platform.OS === 'web') {
      await supabase.auth.signInWithOAuth({
        provider,
      });
      return;
    }

    const returnUrl = makeRedirectUri({ useProxy: false });
    const authUrl = `${supabaseUrl}/auth/v1/authorize?provider=${provider}&redirect_to=${returnUrl}`;
    const response = await startAsync({ authUrl, returnUrl });

    if (response.type !== 'success') {
      return;
    }

    const refreshToken = response.params?.refresh_token;
    if (!refreshToken) return;
    await supabase.auth.setSessionFromToken(refreshToken);
  }
```

Aerilym:
Yeah, that input change and doc change fixes it too. My solution for making it work with expo is:
```TypeScript
    const returnUrl = makeRedirectUri({ useProxy: false });
    const authUrl = `${SB_URL}/auth/v1/authorize?provider=${provider}&redirect_to=${returnUrl}`;
    const response = await startAsync({ authUrl, returnUrl });

    if (response.type !== 'success') {
      return;
    }

    const refreshToken = response.params?.refresh_token;
    const accessToken = response.params?.access_token;
    if (!refreshToken) return;
    supabase.auth.setSession({
      refresh_token: refreshToken,
      access_token: accessToken,
    });
```
I'll look into expanding on the doc changes introduced in #511 as the example usage is outdated.

PR #478, opened 2022-10-11
refactor: mfa client updates

kangmingtay:
* merged `rc` branch in and resolved the merge conflicts 
* use the `decodeBase64URL` helper function in the `rc` branch

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -17,6 +17,7 @@ import {
 } from './lib/errors'
 import { Fetch, _request, _sessionResponse, _userResponse } from './lib/fetch'
 import {
+  decodeBase64URL,
   Deferred,
   getItemAsync,
   getParameterByName,
@@ -356,6 +357,7 @@ export default class GoTrueClient {
           headers: this.headers,
           body: {
             email,
+            data: options?.data ?? {},
             create_user: options?.shouldCreateUser ?? true,
             gotrue_meta_security: { captcha_token: options?.captchaToken },
           },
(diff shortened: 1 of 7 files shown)


src/GoTrueClient.ts
```diff
@@ -1114,7 +1153,7 @@ export default class GoTrueClient {
           throw error
         }
         // Default to Authorization header if there is no existing session
-        jwt = data.session?.access_token ?? this.headers['Authorization']
+        jwt = data.session?.access_token ?? this.headers['Authorization'].split(' ')[1]
```
kangmingtay, line 1156: need to split the header because the header also contains `Bearer` in it

src/GoTrueClient.ts
```diff
@@ -1141,7 +1180,7 @@ export default class GoTrueClient {
...
```
kangmingtay, line 1183: ditto here

test/GoTrueClient.test.ts
```diff
@@ -221,19 +238,21 @@ describe('GoTrueClient', () => {
...
-    })
+      const { error, data } = await auth.signUp({
+        email,
+        password,
+      })
 
-    expect(error).toBeNull()
-    expect(data.session).not.toBeNull()
-    expect(data.user).not.toBeNull()
+      expect(error).toBeNull()
```
J0, line 250: is the extra space intended?

J0 (approved):

test/GoTrueClient.test.ts
kangmingtay: extra space here is due to it being wrapped by line 242

PR #482, opened 2022-10-11
fix: respect `EXPIRY_MARGIN` on `getSession`

david-plugge:


thorwebdev:
Thanks @david-plugge. The team mentioned that the margin isn’t necessary because we only want to refresh if we know for sure that it’s expired.

hf:
Hey @david-plugge, @thorwebdev,

We've been seeing a higher incidence of errors connected to this and I think this change will actually fix some of those issues when tokens are used as they are about to expire.

Pinging the rest of the team for a review @supabase/auth.

J0 (approved):

Issue #504, opened 2022-10-18
Add a client method to manually refresh tokens before they expire


j4w8n:
# Feature request

## Is your feature request related to a problem? Please describe.

Currently, there's no proper way to "manually" refresh tokens before they expire. Calling `getSession()` or `setSession()` will refresh them after they've expired. I believe setting the `autoRefreshToken` client option to `true` will attempt a refresh once the time-until-expiration is less than 10 seconds. Some people may want to proactively get new tokens before the current ones expire, without using the `autoRefreshToken` client option.

## Describe the solution you'd like

A new client method which you can call to refresh tokens (aka a Session), no matter the expiry status of the current tokens.

## Describe alternatives you've considered

Currently, there's a bug in `setSession()` where you can pass in an empty string for `access_token` and you can refresh tokens whenever you'd like. This is what I've been doing.

meznaric:
Just dealt with the same issue this morning, this is a temporary solution I came up with: 

`refreshAccessToken` was available in 1.x version.
```
    const {
      data: { session: data },
      error,
    }: // @ts-expect-error this is a private method that used to be public
    AuthResponse = await this.supabaseClient.auth._refreshAccessToken(
      refreshToken
    );
```

j4w8n:
I've submitted PR #505 to get this added.

j4w8n:
@meznaric you can now use the `refreshSession()` client method, as of supabase-js version 2.0.4.

Server-side usage:
`supabaseClient.auth.refreshSession({ refresh_token: 'token' })`

Client-side usage:
`supabaseClient.auth.refreshSession()`

PR #505, opened 2022-10-18
feat: Add refreshSession method

j4w8n:

Adds a new client method, to refresh tokens whenever you'd like - regardless of the token's expiry status.

There is currently no way to do this. See #504

## What is the new behavior?

You'll be able to use the `refreshSession()` method to refresh tokens (aka the Session).

## Additional context


src/GoTrueClient.ts
```diff
@@ -593,6 +595,43 @@ export default class GoTrueClient {
...
+    currentSession: Pick<Session, 'access_token' | 'refresh_token'>
+  ): Promise<AuthResponse> {
+    try {
+      if (!currentSession.access_token || !currentSession.refresh_token) {
+        throw new AuthSessionMissingError()
+      }
+
+      const payload = decodeJWTPayload(currentSession.access_token)
+      if (!payload.exp) {
+        throw new CustomAuthError('Invalid access token!', 'AuthTokenInvalidError', 400)
```
hf, line 613: Can you please rename this error differently. I'm fine with using `AuthInvalidJWTError` or something of that sort. This also does not have a status code, since it does not go out to the network.

src/GoTrueClient.ts
```diff
@@ -593,6 +595,43 @@ export default class GoTrueClient {
...
+   * If the access token or refresh token are invalid, an error will be thrown.
+   * @param currentSession The current session that minimally contains an access token and refresh token.
+   */
+  async refreshSession(
+    currentSession: Pick<Session, 'access_token' | 'refresh_token'>
+  ): Promise<AuthResponse> {
+    try {
+      if (!currentSession.access_token || !currentSession.refresh_token) {
+        throw new AuthSessionMissingError()
+      }
```
hf, line 609: To refresh a session you only need a refresh token, not an access token. I don't see why it's being used here. 

Can you make the refresh token optional; if not provided you should look it up using `getSession()`.

src/lib/helpers.ts
```diff
@@ -120,3 +120,19 @@ export class Deferred<T = any> {
...
+    throw new Error('JWT is not valid: not a JWT structure')
+  }
+
+  if (!base64Regex.test(parts[1])) {
+    throw new Error('JWT is not valid: payload is not base64')
+  }
+
+  const base64Url = parts[1]
+  return JSON.parse(decodeBase64URL(base64Url))
+}
```
hf, line 138: Newlines at ends of files are important for proper `git diff`.

```suggestion
}

```

hf (commented):
This is great! Some comments to get through, and should be good to merge.

src/GoTrueClient.ts
j4w8n: I was basing this on the `setSession()` method. I can remove it.

src/GoTrueClient.ts
j4w8n: Since the `access_token` isn't need to refresh, then I'll remove this code. You want me to go ahead and remove the `decodeJWTPayload()` helper as well, and let your PR take care of it?

src/GoTrueClient.ts
j4w8n: One comment about making the `refresh_token` optional:

A use-case for this method is to use it in a server-side context, which means there wouldn't be a session returned from `getSession()`. So I'm thinking we can make `currentSession` optional, but if it's passed in, then it must contain a `refresh_token`. I assume `currentSession?: Pick<Session, 'refresh_token'>` would do this??

Am I thinking about this correctly?

src/GoTrueClient.ts
j4w8n: I see your PR was already merged. I'll remove the decode function from my branch and sync changes.

src/GoTrueClient.ts
```diff
@@ -593,6 +593,48 @@ export default class GoTrueClient {
     }
   }
 
+  /**
+   * Returns a new session, regardless of expiry status.
+   * Takes in an optional current session. If not passed in, then refreshSession() will attempt to retreive it from getSession().
```
kangmingtay, line 598:
```suggestion
   * Takes in an optional current session. If not passed in, then refreshSession() will attempt to retrieve it from getSession().
```

kangmingtay (commented):
nice work @j4w8n!

j4w8n:
Thanks for catching that @kangmingtay!

david-plugge:
Nice work!

kangmingtay (approved):

PR #509, opened 2022-10-19
feat: add challengeAndVerify

J0:
as per title

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -145,6 +145,7 @@ export default class GoTrueClient {
       unenroll: this._unenroll.bind(this),
       challenge: this._challenge.bind(this),
       listFactors: this._listFactors.bind(this),
+      challengeAndVerify: this._challengeAndVerify.bind(this),
       getAuthenticatorAssuranceLevel: this._getAuthenticatorAssuranceLevel.bind(this),
     }
   }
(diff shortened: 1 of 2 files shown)


src/GoTrueClient.ts
```diff
@@ -1138,6 +1139,22 @@ export default class GoTrueClient {
...
+  private async _challengeAndVerify(params: MFAChallengeAndVerifyParams): Promise<AuthMFAVerifyResponse> {
+    const { data: sessionData, error: sessionError } = await this.getSession()
+    if (sessionError) {
+      return { data: null, error: sessionError }
+    }
+    const {data: challengeData, error: challengeError} = await this.mfa.challenge({factorId: params.factorId})
+    if (challengeError) {
+      return { data: null, error: challengeError}
+    }
+    return await _request(this.fetch, 'POST', `${this.url}/factors/${params.factorId}/verify`, {
```
kangmingtay, line 1151: why don't we call the `this._verify` method?

kangmingtay (changes requested):

src/GoTrueClient.ts
J0: makes sense, just replaced.

src/GoTrueClient.ts
```diff
@@ -1138,6 +1139,17 @@ export default class GoTrueClient {
       jwt: sessionData?.session?.access_token,
     })
   }
+  private async _challengeAndVerify(params: MFAChallengeAndVerifyParams): Promise<AuthMFAVerifyResponse> {
+    const { data: sessionData, error: sessionError } = await this.getSession()
+    if (sessionError) {
+      return { data: null, error: sessionError }
+    }
+    const {data: challengeData, error: challengeError} = await this.mfa.challenge({factorId: params.factorId})
```
kangmingtay, line 1147:
```suggestion
    const { data: challengeData, error: challengeError } = await this.mfa.challenge({factorId: params.factorId})
```

src/lib/types.ts
```diff
@@ -783,6 +793,16 @@ export interface GoTrueMFAApi {
    */
   unenroll(params: MFAUnenrollParams): Promise<AuthMFAUnenrollResponse>
 
+/**
+   * Helper methdo which creates a challenge and immediately uses the given code to verify against it thereafter. The verification code is
```
kangmingtay, line 797:
```suggestion
   * Helper method which creates a challenge and immediately uses the given code to verify against it thereafter. The verification code is
```

PR #512, opened 2022-10-19
fix: add forgotten `data` option for sign ins

hf:
`data` options were forgotten for some sign in methods.

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -300,6 +300,12 @@ export type SignInWithPasswordCredentials =
       /** The user's password. */
       password: string
       options?: {
+        /**
+         * A custom data object to store the user's metadata. This maps to the `auth.users.user_metadata` column.
+         *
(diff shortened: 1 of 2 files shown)


J0 (approved):

kangmingtay (changes requested):
~we also need to add this to `signInWIthOtp` right?~ 

edit: just checked, it's already there

kangmingtay (approved):

src/lib/types.ts
```diff
@@ -300,6 +300,12 @@ export type SignInWithPasswordCredentials =
       /** The user's password. */
       password: string
       options?: {
+        /**
+         * A custom data object to store the user's metadata. This maps to the `auth.users.user_metadata` column.
+         *
+         * The `data` should be a JSON object that includes user-specific info, such as their first and last name.
+         */
+        data?: object
```
kangmingtay, line 308: actually `SignInWithPassword` doesn't accept a data option in gotrue

PR #515, opened 2022-10-21
fix: mfa challenge and verify

kangmingtay:
* Fixes missing type import and invalid params passed into `_verify`


J0 (approved):

J0:
Thanks for this, should have tested. I'll bump the version on supabase-js once the other changes go in

PR #516, opened 2022-10-21
fix: disallow invalid access_token in setSession

### LINKED ISSUE #490: Clarify usage of setSession() in v2.0.0
# Bug report

## Describe the bug

There are a couple of things going on here. One is documentation, the other is a possible bug.

1. Documentation seems to imply you have to pass an `access_token`, `refresh_token`, and a `user`. But I'm able to make this work without passing in a `user`.
```
* @param currentSession The current session that minimally contains an access token, refresh token and a user.
```
2. Based on things I've seen, it seems like the maintainers only want tokens being refreshed if the current `access_token` has expired (or possibly within the `EXPIRY_MARGIN` too, per a [pull comment](https://github.com/supabase/gotrue-js/pull/482#issuecomment-1279362251)). With current code, you can bypass `hasExpired`, and get new tokens, by passing in an empty string for the `access_token` - even if the current `access_token` is still valid.
```
const { data, error } = await supabaseClient.auth.setSession({ access_token: "", refresh_token })
```
I like being able to do this, so that I can get new tokens before the current one has expired. However, as long as there's no downside to user experience, then I'm good with only refreshing if expired. Maybe need to do a check for `access_token`, otherwise throw the `AuthSessionMissingError`.

## Expected behavior

I'm unsure of the expected behavior, but assume we shouldn't be allowed to pass in an empty string for the `access_token`.

## System information

- OS: [macOS]
- Browser (if applies) [Firefox]
- Version of supabase-js: [2.0.0]
- Version of Node.js: [16.15.0]

j4w8n:

Adds checks for the passed-in `access_token`. Also refactors `setSession()`, to align the method with common behaviors of other methods.

Some people, including myself, are passing in an empty string for the access token, to force a session refresh even if the session isn't expired. There is now a `refreshSession()` method to accomplish this, as of supabase-js version 2.0.4

## What is the new behavior?

Checks if the access token is an empty string. Also adds a regex check to the `decodeJWTPayload()` helper, to verify the access token is in base64url format; otherwise we could still pass in something like `hello.there.world` and we'd get the same forced refresh behavior.

As for the refactor, I noticed that this method is directly calling `_refreshAccessToken()` then calling `_saveSession()`. However, `_callRefreshToken()` is used on other methods; and is a better fit, since it seems to handle multiple refresh calls. Plus it calls `_saveSession()` itself, which eliminates the need for that call in `setSession()`.

## Additional context

Closes #490

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -15,7 +15,7 @@ import {
   AuthUnknownError,
   isAuthError,
 } from './lib/errors'
-import { Fetch, _request, _sessionResponse, _userResponse } from './lib/fetch'
+import { Fetch, _request, _sessionResponse, _userResponse, _ssoResponse } from './lib/fetch'
 import {
   Deferred,
   getItemAsync,
@@ -36,11 +36,13 @@ import type {
   GoTrueClientOptions,
   InitializeResult,
   OAuthResponse,
+  SSOResponse,
   Provider,
   Session,
   SignInWithOAuthCredentials,
   SignInWithPasswordCredentials,
   SignInWithPasswordlessCredentials,
+  SignInWithSSO,
   SignUpWithPasswordCredentials,
   Subscription,
   SupportedStorage,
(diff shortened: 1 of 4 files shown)


j4w8n:
I'm not sure why `helpers.ts` is not showing a newline at eof. It's there in VSCode and everything is updated.

```shell
git checkout fix-set-session-loophole
Switched to branch 'fix-set-session-loophole'
git push origin fix-set-session-loophole
Everything up-to-date
```

j4w8n:
> I'm not sure why `helpers.ts` is not showing a newline at eof. It's there in VSCode and everything is updated.
> 
> ```shell
> git checkout fix-set-session-loophole
> Switched to branch 'fix-set-session-loophole'
> git push origin fix-set-session-loophole
> Everything up-to-date
> ```

Actually, when I go to edit the file in the GitHub interface, the eof newline is there. Maybe it's just a bug in GitHub??

PR #525, opened 2022-11-01
fix: use `captcha_token` in `verifyOtp`

hf:
It's sending `captchaToken` to GoTrue which needs to be snake case. Ouch.


kangmingtay (approved):

J0 (commented):
Lgtm

PR #530, opened 2022-11-07
fix: the deleteUser endpoint not return user data

leynier:

Bug fix

Currently, the `deleteUser` method of `AdminApi` returns a response typed as a `UserResponse`, but it is actually an empty object.

## What is the new behavior?

Now only an `AuthError` is returned if it exists.

## Additional context

Code of `GoTrue` server-side

https://github.com/supabase/gotrue/blob/fc6f58d1437f896e893c8c93d122cefa8c3546ac/api/admin.go#L402

### DIFF
--- a/src/GoTrueAdminApi.ts
+++ b/src/GoTrueAdminApi.ts
@@ -226,11 +226,10 @@ export default class GoTrueAdminApi {
    *
    * This function should only be called on a server. Never expose your `service_role` key in the browser.
    */
-  async deleteUser(id: string): Promise<UserResponse> {
+  async deleteUser(id: string): Promise<{ data: null; error: AuthError | null }> {
     try {
       return await _request(this.fetch, 'DELETE', `${this.url}/admin/users/${id}`, {
         headers: this.headers,
-        xform: _userResponse,
       })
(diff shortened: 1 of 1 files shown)


j4w8n:
The [UserResponse](https://github.com/supabase/gotrue-js/blob/master/src/lib/types.ts#L88-L100) seems correct to me. We have to keep in mind that the same response type is returned when creating a user. Why do you believe it's incorrect when deleting a user?

```ts
export type UserResponse =
  | {
      data: {
        user: User
      }
      error: null
    }
  | {
      data: {
        user: null
      }
      error: AuthError
    }
```

When a user is created, I'd expect a non-empty user object:
```js
{ 
  data: {
    user: {
      id: 'something',
      aud: 'authenticated',
      role: 'authenticated',
      ...
    },
  },
  error: null
}
```

When a user is deleted, I'd expect an empty user object:
```js
{ data: { user: {} }, error: null }
```

kangmingtay:
Hey @leynier, this would be a breaking change and @j4w8n is right here. The user object is expected to be empty. Is there a reason why you want to change this behaviour though?

PR #537, opened 2022-11-14
feat: support pagination options for `listUsers()` method

j4w8n:

Allows you to pass in a `params` object to `listUsers()`; to determine paginated results. 

`listUsers()` will only return the first 50 results.

## What is the new behavior?

Introduces an options object to `listUsers()`. The two options are `page` and `per_page`. They can be used together or individually. These option names are consistent with the api code found [here](https://github.com/supabase/gotrue/blob/930f5affdab112db81e17ef799418206bead9092/api/pagination.go#L42-L43).

`page` - which page of paginated results to return
`per_page` - how many results to return per page

Example:
```js
const { data, error } = await client.auth.admin.listUsers({ page: '2', per_page: '100' })
```

## Additional context

`params` option name was based on existing usage like [this](https://github.com/supabase/gotrue-js/blob/master/src/GoTrueAdminApi.ts#L105) code.
`PageParams` type name was based on [this](https://github.com/supabase/gotrue/blob/master/api/admin.go#L103) code.

Closes https://github.com/supabase/gotrue-js/issues/538.

### DIFF
--- a/src/GoTrueAdminApi.ts
+++ b/src/GoTrueAdminApi.ts
@@ -163,7 +163,7 @@ export default class GoTrueAdminApi {
   > {
     
     try {
-     const { data, error } = await _request(this.fetch, 'GET', `${this.url}/admin/users`, {
+      const { data, error } = await _request(this.fetch, 'GET', `${this.url}/admin/users`, {
         headers: this.headers,
         query: params,
       })
--- a/src/GoTrueAdminApi.ts
+++ b/src/GoTrueAdminApi.ts
@@ -157,19 +157,16 @@ export default class GoTrueAdminApi {
    * Get a list of users.
    *
    * This function should only be called on a server. Never expose your `service_role` key in the browser.
-   * @param params An object which supports `page` and `per_page` as numbers, to alter the paginated results.
+   * @param params An object which supports `page` and `per_page` as strings, to alter the paginated results.
    */
   async listUsers(params?: PageParams): Promise<
     { data: { users: User[] }; error: null } | { data: { users: [] }; error: AuthError }
   > {
-    let pageParams = ''
-    if (params) {
-      pageParams = params.page && params.per_page ? `?page=${params.page}&per_page=${params.per_page}` : params.page ? `?page=${params.page}` : params.per_page ? `?per_page=${params.per_page}` : ''
-    }
     
     try {
-     const { data, error } = await _request(this.fetch, 'GET', `${this.url}/admin/users${pageParams}`, {
+     const { data, error } = await _request(this.fetch, 'GET', `${this.url}/admin/users`, {
         headers: this.headers,
+        query: params,
       })
       if (error) throw error
       return { data: { ...data }, error: null }
--- a/src/GoTrueAdminApi.ts
+++ b/src/GoTrueAdminApi.ts
@@ -12,6 +12,7 @@ import {
   AuthMFAAdminDeleteFactorResponse,
   AuthMFAAdminListFactorsParams,
   AuthMFAAdminListFactorsResponse,
+  PageParams,
 } from './lib/types'
 import { AuthError, isAuthError } from './lib/errors'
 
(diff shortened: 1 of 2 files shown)


src/GoTrueAdminApi.ts
```diff
@@ -156,12 +157,18 @@ export default class GoTrueAdminApi {
...
+   * @param params An object which supports `page` and `per_page` as numbers, to alter the paginated results.
    */
-  async listUsers(): Promise<
+  async listUsers(params?: PageParams): Promise<
     { data: { users: User[] }; error: null } | { data: { users: [] }; error: AuthError }
   > {
+    let pageParams = ''
+    if (params) {
+      pageParams = params.page && params.per_page ? `?page=${params.page}&per_page=${params.per_page}` : params.page ? `?page=${params.page}` : params.per_page ? `?per_page=${params.per_page}` : ''
+    }
```
valiafetisov, line 168: Nit: Since the presence of params is already validated by types, you can just directly convert it to a query string, no?
```suggestion
    const pageParams = new URLSearchParams(params).toString();
```

Otherwise, you can generically clean it up beforehand:
```suggestion
    const nonFalsyParams = Object.keys(params).filter(key => params[key]).reduce((object, key) => ({ ...object, [key]: params[key] }), {})
    const pageParams = new URLSearchParams(nonFalsyParams).toString();
```

valiafetisov (approved):

valiafetisov (approved):

src/GoTrueAdminApi.ts
j4w8n: Well that would be badass. When I try that, I get a type error though. I'm not familiar enough with typescript to know if there's a way to resolve it.

```
Type 'PageParams' is not assignable to type 'string | string[][] | Record<string, string> | URLSearchParams | undefined'.
    Type 'PageParams' is not assignable to type 'Record<string, string>'.
      Property 'page' is incompatible with index signature.
        Type 'number' is not assignable to type 'string'.
```

valiafetisov: Oh, interesting, didn't know `URLSearchParams` only accept strings. This should work then:
```suggestion
    const nonFalsyParams = Object.keys(params).filter(key => params[key]).reduce((object, key) => ({ ...object, [key]: params[key].toString() }), {})
    const pageParams = new URLSearchParams(nonFalsyParams).toString();
```
Also, notice that `new URLSearchParams(nonFalsyParams).toString()` wouldn't contain leading `?` symbol, so you have to add it to the url below

j4w8n: Ok, so let me ask you this. In the type definition, `page` and `per_page` are numbers. If I change those types to `string`. Your one-liner code suggestion is fine - no type errors. Of course, people would then need to do:

```js
const { data, error } = await client.auth.admin.listUsers({ page: '2', per_page: '100' })
```

Which would the team prefer? It would seem weird to define them as strings, but.


If I do the two-liner, I get this type error:

```
No overload matches this call.
  Overload 1 of 2, '(o: {}): string[]', gave the following error.
    Argument of type 'PageParams | undefined' is not assignable to parameter of type '{}'.
```

j4w8n:
After some additional digging, I realized that `_request` has a `query` option, so we just need to pass `params` that way.

src/lib/types.ts
```diff
@@ -970,3 +970,8 @@ export type CallRefreshTokenResult =
       session: null
       error: AuthError
     }
+
+export type PageParams = {
+  page?: string,
+  per_page?: string
```
kangmingtay, line 976:
```suggestion
  page?: number,
  perPage?: number
```

src/GoTrueAdminApi.ts
```diff
@@ -155,13 +156,15 @@ export default class GoTrueAdminApi {
...
+   * @param params An object which supports `page` and `per_page` as strings, to alter the paginated results.
    */
-  async listUsers(): Promise<
+  async listUsers(params?: PageParams): Promise<
     { data: { users: User[] }; error: null } | { data: { users: [] }; error: AuthError }
   > {
     try {
       const { data, error } = await _request(this.fetch, 'GET', `${this.url}/admin/users`, {
         headers: this.headers,
+        query: params,
```
kangmingtay, line 167:
```suggestion
        query: { page: params?.page ?? '', per_page: params?.perPage ?? '' },
```

src/lib/types.ts
kangmingtay: nit: preference of camelCase over snake case here

kangmingtay (approved):

kangmingtay:
@j4w8n thanks for taking the time to add this to `listUsers`!

src/lib/types.ts
kangmingtay: using the `number` as the type is better for the DX since it prevents devs from using non-numbers as values

kangmingtay: also, gotrue handles the type conversion for us

j4w8n:
You're welcome @kangmingtay. The only piece I'm concerned about after your changes is the params definition on line 159 of GoTrueClient.ts. It still states using `per_page` and `strings`.

I'm also not familiar with the process of updating the docs since today's launch. I can dig into, but I have zero React experience. Wouldn't need that in place until the gotrue-js version is bumped in supabase-js, but wanted to mention it.

kangmingtay:
@j4w8n ah yes i missed this, will fix in a FLUP PR. I can handle the update on the docs side as well!

Issue #524, opened 2022-10-29
signOut does not clear user jwt from gotrue authorization header allowing gotrue operations and exposing user jwt


GaryAustin1:
# Bug report

## Describe the bug
This is from a user discussion on [DISCORD](https://discord.com/channels/839993398554656828/1035578821853855815 )

Although signOut clears the session data from local storage, it appears it leaves the user jwt in the gotrue headers object and in a second object called changedAccessToken.

If you run getUser() after signOut a request to the server goes out and the user data is returned. 
Although database API calls revert to the anon key, the jwt could also be copied and used until it expires.

A temporary solution is probably to set the client "object" to null, or reinitialize the client if further work needs to be done after signout.

## To Reproduce
```
console.log('login')
        const {data} = await supabase1.auth.signInWithPassword({email: 'email', password:'pass'} )
        console.log('login ',data)
        console.log('signOut ',await supabase1.auth.signOut())
        console.log('getSession ', await supabase1.auth.getSession())   //null
        console.log('supabase ',supabase1)  //see image below
        const result = await supabase1.from('realtest').select()  //authentication header anon
        console.log ('getUser ', await supabase1.auth.getUser())  //auth header has user jwt and returns user data from server
```
Supabase object after signOut()
![signout](https://user-images.githubusercontent.com/54564956/198843903-9c1a2ba5-205d-49bc-b220-86f5e1d6c8ce.JPG)

getUser going out with a user jwt after signOut().
![image](https://user-images.githubusercontent.com/54564956/198844032-df4efafc-c0ff-4060-9dad-795faf89bc07.png)

## Expected behavior

SignOut() should remove any jwt info from memory and certainly no allow gotrue operations to continue.   

It is understood there is a risk of unexpired jwts after signOut, but they should not be left around for someone to find or use after signOut().

## System information

supabase-js 2.0.4

GaryAustin1:
This is a slightly different issue, but related to auth having its own headers...

.getUser(jwt) if called while using service_role sets the authentication header in auth to the User JWT.
The supabase client header still stays service_role, but if you make further admin auth calls after doing a .getUser call they will be rejected with 401.

If this is desired behavior (getUser(jwt) then this should be documented as users are running into this server side.

![getUser](https://user-images.githubusercontent.com/54564956/199859400-7a18244d-52e3-445f-af16-b1aeceaa2a26.JPG)

j4w8n:
[...]

Issue #520, opened 2022-10-23
signOut - improve response if user has been deleted


MatthewPosgate:
At the moment, if you delete a user on the server, and the user then tries to sign out with supabaseClient.auth.signOut(), we get 
AuthApiError: Internal server error.

Can we change this to say something along the lines of 'user does not exist'?
Then if I detect that error I can sign them out anyway in my app

cinan:
I noticed when calling `supabase.auth.signOut()` and server throws an error (500), session won't get cleared. That means I can't log out a user from users' browser.

hf:
Fixed in https://github.com/supabase/gotrue/pull/783. Let us know if you're still facing the same issue.

PR #541, opened 2022-11-19
fix: signout should remove session if user doesn't exist

### LINKED ISSUE #540: No way to clear expired User. Using auth.signOut fails
# Bug report

## Describe the bug

If you call supabase.auth.signOut() with an expired JWT, GoTrue fails with:
AuthApiError: invalid JWT: unable to parse or verify signature, token

The problem is that the stale user session is still stored in the auth module.
There is no way to remove it with the API.

That means that now edge functions will fail because of the invalid JWT (even if an anonymous JWT would have worked).

This puts the browser client in a broken and unfixable state.

## Expected behavior

If GoTrue reports that the JWT is expired, then the user is signed out, and thet session should be cleared from supabase.auth
Alternatively I could catch the error, but would need an API to clear the current session from supabase.auth

## System information

- Browser Chrome
- Version of supabase-js: 2.0.0

kangmingtay:
* Fixes #540, #520

### DIFF
--- a/src/GoTrueAdminApi.ts
+++ b/src/GoTrueAdminApi.ts
@@ -7,7 +7,6 @@ import {
   User,
   UserResponse,
   GoTrueAdminMFAApi,
-  UpdatableFactorAttributes,
   AuthMFAAdminDeleteFactorParams,
   AuthMFAAdminDeleteFactorResponse,
   AuthMFAAdminListFactorsParams,
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -13,6 +13,7 @@ import {
   AuthRetryableFetchError,
   AuthSessionMissingError,
   AuthUnknownError,
+  isAuthApiError,
   isAuthError,
 } from './lib/errors'
 import { Fetch, _request, _sessionResponse, _userResponse, _ssoResponse } from './lib/fetch'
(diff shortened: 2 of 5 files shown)


j4w8n:
Sweet! This seems to resolve enough of 540 to close it, and all of 520 🎉; but not #524, as you stated it would in the description.

I fetched your PR locally and tested. If you add the below code in the `signOut` test (not the test you added, but the one right above it), it still returns user data with `getUser()` after the user is signed out. From the client output, you'll see where the `headers` and `admin.headers` still have the logged-out user's jwt. This is the main concern of 524.

```diff
const res = await authWithSession.signOut()
expect(res).toBeTruthy()
+ console.log(authWithSession)
+ const { data: { user } } = await authWithSession.getUser()
+ expect(user).toBe(null)
```

I was able to "reset" the `GoTrueClient` header by doing `this.headers = { ...DEFAULT_HEADERS }` in `_removeSession()`. But that still leaves the `GoTrueAdminApi` headers; which I wasn't able to access via `this.admin.headers` from the GoTrueClient since it's protected.

Also, when using the supabase-js client in a project, the `this.changedAccessToken` property still contains the jwt. See Gary's output in 524 [here](https://github.com/supabase/gotrue-js/issues/524#issue-1428366104).

I'd say we'll need to make a change in the supabase-js repo as well if we want to resolve 524. See comment [here](https://github.com/supabase/gotrue-js/issues/524#issuecomment-1304857759) for how I was able to do that on the fly. Might not be the way you want to resolve it - I was just experimenting.

thorwebdev:
I think the root of this issue is https://github.com/supabase/supabase/issues/5990

thorwebdev (approved):
LGTM. I think we'll need to fix #524 separately as it will require changes to supabase-js.

kangmingtay:
@j4w8n fixed #524 in [this commit](https://github.com/supabase/gotrue-js/pull/541/commits/0ad574d09d2e44b0ba313eecc47e8d3bb4288f58), turns out the `_request` wrapper method was directly using the headers object passed into it, and since objects in javascript are passed by reference instead of by value, we need to copy it into a new object first

thanks for catching this btw! i've updated the test suite too to prevent regression in the future

PR #544, opened 2022-11-25
feat: return pagination data for the `listUsers()` method

j4w8n:

Returns pagination data, when api requests make it available via the `link` and `x-total-count` response headers.

For example, #537 allows you to pass info to the `listUsers()` method, to get specific paginated results. However, without more information, it's difficult to know what to ask for after the first query.

This PR is effectively useless without #537 being merged. Otherwise there's no way to utilize the additional data returned from this PR. But I wanted to create a separate PR just in case the team decided to implement #537 but not this.

gotrue-js returns whatever data the gotrue api returns.

In the case of `listUsers()`, this is an array of `users` plus `aud` (which always equals `authenticated` from what I can tell)

## What is the new behavior?

gotrue-js will also return these three properties:

`nextPage` - the next pagination page to query for more results; or `null` if you just queried the last set of results.
`lastPage` - the last pagination page for the query results.
`total` - the total number of records for the query (ex, if you asked for 100 results but there were a total of 362, this would be 362)

## Additional context

The below example assumes #537 is merged and a supabase-js request of `await supabase.auth.admin.listUsers({per_page: '10'})`

Relevant headers returned from the api. The new code parses and returns this info.
```js
'link' => '</admin/users?page=2&per_page=10>; rel="next", </admin/users?page=7&per_page=10>; rel="last"'
'x-total-count' => '62'
```

new supabase-js `data` response from above query:
```diff
{
  aud: 'authenticated',
  users: [ <omitted 10 entries> ],
+  nextPage: 2,
+  lastPage: 7,
+  total: 62
}
```


j4w8n:
Hmm... apparently, I wasn't quite using my thinker here.

The newly returned info - specifically `next` and `last` - would still need to be parsed by the user. Having the whole url is kinda worthless in the context of subsequent calls to `listUsers()`. It'd be fine if you were making direct queries to the api.

I suppose an alternative would be to instead return data like:

```js
next_page: '2',
last_page: '7',
per_page: '10',
total: 62
```

However, I'm starting to wonder if it's best to leave these calculations up to the user, for their `listUsers` call, and just return `total` so they can make those calculations.

src/lib/fetch.ts
```diff
@@ -112,10 +113,25 @@ async function _handleRequest(
...
+                const url = link.split(';')[0].replace(/[\<\>\s]/g, '')
+                const rel = JSON.parse(link.split(';')[1].replace(/[\<\>\s]/g, '').split('=')[1])
+                pagination[rel] = url
+            })
+            
+            if (!pagination.next) pagination.next = null
+            const total = result.headers.get('x-total-count') ?? '0'
+            pagination.total = parseInt(total)
+        }
+        return { ...json, ...pagination }
```
kangmingtay, line 134: This logic can be moved to the `listUsers` method instead. There is the option to return the `result` instead of `result.json()` by setting the `noResolveJson` option to true and then you should be able to handle destructuring the appropriate headers in there.

src/GoTrueAdminApi.ts
```diff
@@ -157,7 +157,7 @@ export default class GoTrueAdminApi {
    * This function should only be called on a server. Never expose your `service_role` key in the browser.
    */
   async listUsers(): Promise<
-    { data: { users: User[] }; error: null } | { data: { users: [] }; error: AuthError }
+    { data: { users: User[], aud: string, next?: string | null, last?: string, total?: number }; error: null } | { data: { users: [] }; error: AuthError }
```
kangmingtay, line 160: You can use a type intersection here if you want to reuse the Pagination type defined below
```suggestion
    { data: { users: User[], aud: string } & Pagination; error: null } | { data: { users: [] }; error: AuthError }
```

kangmingtay (changes requested):
hey @j4w8n, yeah this is something we would very much like to add as well since it makes the pagination api more useful in `listUsers`. The idea is right, but i think we can move a couple of things around so as to isolate the logic within `listUsers` rather than have it in the `handleRequest` helper method

src/GoTrueAdminApi.ts
kangmingtay: created an example in the ts playground here: https://www.typescriptlang.org/play?#code/C4TwDgpgBAYg9nKBeKBvAUFKAzAlgJwGdgA5AQwFsIAuKY-XAOwHN0BfddbAV0YGNguOIygAhMvgAUASlqooAEzLAycqN0IQitek2YBtALpQ2UAGSwEJtJij4IwbvhHylKtRq2FaRgDRY8IlJKGigAcgArOAALRjDrDjYgA

src/lib/fetch.ts
j4w8n: Thanks for taking a look at this @kangmingtay!

I can't figure out how to grab the data by the time it gets back to `listUsers()`.

In `_handleRequest()` I can access the headers with `result.headers.get()`.
In `_request()` I have to use something like `data.headers.forEach()` because methods like `get` and `has` have been stripped.
Finally, by the time the data reaches `listUsers()`, `typeof` says `data` is an object, but I'm unable to access things the way I'm used to. Not to mention I'm not sure if I can even access the returned users at this point.

Maybe I've misunderstood your suggestion and gone down the wrong rabbit hole??

j4w8n: In `_handleRequest()` it's simply returning the result
```js
if (options?.noResolveJson) return result
```
In `_request()` it's basically doing the same; spreading it out
```js
return options?.xform ? options?.xform(data) : { data: { ...data }, error: null }
```
And inside of `listUsers()`, my test of `console.log(data.json())` is throwing `data.json is not a function`.
```js
const { data, error } = await _request(this.fetch, 'GET', `${this.url}/admin/users`, {
  headers: this.headers,
  noResolveJson: true
})
if (error) throw error
console.log('list', data.json())
```

j4w8n: @kangmingtay I was thinking about this at 4:45am, and realized what the issue is. We're not returning a response, but more like a copy of the response data.

The only way I can get this to work is to change what is returned in `_request()`. This way we're returning an actual response.

from this
```js
return options?.xform ? options?.xform(data) : { data: { ...data }, error: null }
```
to this
```js
return options?.xform ? options?.xform(data) : 
    options?.noResolveJson ? 
      new Response(data.body, { 
        headers: data.headers, 
        status: data.status, 
        statusText: data.statusText 
      }) 
      : { data: { ...data }, error: null }
```

j4w8n: psh... also looks like I've managed to pull down upstream changes somehow. So there's a lot more diffs that what I've actually done for this.

j4w8n: > psh... also looks like I've managed to pull down upstream changes somehow. So there's a lot more diffs that what I've actually done for this.

Got that rectified.

j4w8n:
@kangmingtay I'm not sure why the test is having issues with the `new Response...` line. Can you give your input? The test:suite passes locally (except the two tests for strict equality on updating a user - which I did not write as part of this PR).

src/lib/fetch.ts
```diff
@@ -162,6 +162,10 @@ export function _generateLinkResponse(data: any): GenerateLinkResponse {
   }
 }
 
+export function _noResolveJsonResponse(data: any): Response {
```
kangmingtay, line 165: we just needed to return the response object if `noResolveJson` is passed it because we want the whole response and not the body

src/GoTrueAdminApi.ts
```diff
@@ -156,21 +163,47 @@ export default class GoTrueAdminApi {
...
+          )
+          pagination[rel] = url
+        })
+
+        if (!pagination.next) pagination.next = null
+        const total = response.headers.get('x-total-count') ?? '0'
+        pagination.total = parseInt(total)
+      }
+      const users = await response.json()
+      return { data: { ...users, ...pagination }, error: null }
```
kangmingtay, line 206: hmm the response returned here is something like the following:

```json
{ 
  users: [ ... ], 
  aud: 'authenticated',
  next: '/admin/users?page=3&per_page=3',
  last: '/admin/users?page=34&per_page=3',
  total: 100
}
```

i'm not too sure how useful the `next` and `last` properties are, maybe it will be more useful to rename this to `nextPage` and `lastPage` and just return the number instead of the url path. Wdyt?

kangmingtay (approved):

src/lib/fetch.ts
kangmingtay: There is an option to pass in a transformer which is just a fancy word here for a callback function. The callback function will be used inside `_request` here: `return options?.xform ? options?.xform(data) : { data: { ...data }, error: null }`

src/GoTrueAdminApi.ts
kangmingtay: i think we don't have to return the `per_page` value, we can just return the page numbers ?

kangmingtay: oops, might have missed that - thanks for being so responsive thus far though!

j4w8n:
Believe this is ready to go now.

j4w8n:
@kangmingtay, thank you for your help and guidance on this!

Issue #539, opened 2022-11-18
getSession with persistSession true(default) fails silently with no localStorage


GaryAustin1:
# Bug report

## Describe the bug

Currently getSession will clear out the memory session and return null if there is no localStorage and persistSession is left to default of true.

This so far has caused 1/2 dozen user issues on Discord where users are moving from v1 to v2 and losing lots of time debugging missing session data after sign in.  Some of these users are moving from the old .user() method to check for a user to the getSession method to check for a user.  Some are on serverside code, another was Electron.  They don't need the persist, and setting the flag to false solves the issue, but that is a change that setting the flag is required.

## To Reproduce

signInWithPassord in an environment without local storage and do getSession().  Result will be null for session AND the in memory session is cleared.

Relevant gotrue-js code here.
https://github.com/supabase/gotrue-js/blob/a441eef4d7291ec20b756cde99ccee87329594d0/src/GoTrueClient.ts#L512

## Expected behavior

Either an error should be generated pointing to persistSession being on and no local storage, or go back to v1 like method where in memory still works,  just nothing will be persisted.

## System information

supabase-js 2.x.x

## Additional context
related: https://github.com/supabase/supabase-js/issues/571

j4w8n:
Hopefully this is getting more visibility soon.

There are similar issues when using `getUser()` server-side, without passing in a jwt.

As well as `setSession()`; but that one is really because of how `_saveSession()` works, when called.

iolyd:
I think something's broken with the storage in general and `getSession()`'s returned session is always null even when providing a server-compatible storage. I'm fairly certain doing something like this worked before one of the latest updates:
```ts
// Fill-in replacement for lack of local storage on server context.
function sessionStorageProvider(): SupportedStorage {
  const s = new Map();
  return {
    getItem: (key: string) => {
      console.log('Get item', s)
      return s.get(key);
    },
    setItem: (key: string, value: string) => {
      s.set(key, value);
      console.log('Set item', s)
    },
    removeItem: (key: string) => {
      s.delete(key);
    },
  };
}

function createServerClient() {
  return createClient<App.DatabaseSchema>(PUBLIC_SUPABASE_URL, PUBLIC_SUPABASE_ANON_KEY, {
    auth: {
      persistSession: true,
      autoRefreshToken: false,
      detectSessionInUrl: false,
      storage: sessionStorageProvider(),
    },
  });
}
```
But when using it recently, the storage is not passed around/kept properly:
```ts
[...]

PR #548, opened 2022-12-02
fix: setSession is in broken state after v2.4.0

j4w8n:

@kangmingtay, bit of a critical here. Completely my fault.

Fixes a critical issue with `setSession()`. With the errant code, as part of #536, if a token hasn't expired, the session is returned but never gets saved. I removed a line when consolidating, but it should've been moved up instead. I completely missed this, and I apologize.

Thanks to someone else for finding this! https://github.com/supabase/gotrue-js/issues/539#issuecomment-1334692479

## Additional Context

Also tweaks the existing setSession test, to ensure the method is saving the session. This would've caught my coding error.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -665,6 +665,7 @@ export default class GoTrueClient {
           expires_in: expiresAt - timeNow,
           expires_at: expiresAt,
         }
+        await this._saveSession(session)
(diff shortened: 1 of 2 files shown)


kangmingtay (approved):
ah i missed this too during the review, thanks for making the fix!

PR #551, opened 2022-12-07
fix: update mfa methods

kangmingtay:
* MFA methods should be wrapped in try-catch blocks so that the error returned from gotrue can be handled gracefully
* This is because of [this line](https://github.com/supabase/gotrue-js/blob/0fcc8f576d255b92e4e8bdc57c450e8b3390b700/src/lib/fetch.ts#L42) in `fetch.ts`, where the promise is rejected if there is a non-5xx error returned from gotrue. 
* made some small updates to the docstrings for some mfa methods


src/GoTrueClient.ts
```diff
@@ -1156,81 +1165,113 @@ export default class GoTrueClient {
...
+      }
+      throw error
+    }
   }
+
+  /**
+   * Creates a challenge and immediately verifies it
+   * @param factorId System assigned identifier for authenticator device as returned by enroll
+   * @param code Code Generated by an authenticator device
+   */
```
hf, line 1274: These comments won't work unfortunately because they are on private methods and are not visible from the outside. Move the comment to the `lib/types.ts` file.

hf (approved):

src/GoTrueClient.ts
kangmingtay: these are just for internal use since they are private methods, the comments won't be exported anyway

PR #555, opened 2022-12-10
feat(gotrue/client): add noResolveJson options in GoTrueClient constructor

canserkanuren:

This is a new feature, implementing the parameter `noResolveJson` that will be used in `_handleRequest` function in https://github.com/supabase/gotrue-js/blob/master/src/lib/fetch.ts#L79

## What is the new behavior?

## Additional context


kangmingtay:
hi @canserkanuren, can you please explain why we need to add this to the constructor?

PR #564, opened 2022-12-20
feat: add new auto refresh token algorithm

hf:
Adds a new algorithm that automatically refreshes the user's session.

Problems with the previous algorithm:
- Complex retry logic 
- Difficult to read and reason about quickly
- Used in-memory state while session state is not in memory -- local storage, cookies, etc.
- Could produce endless retries
- Not multi-tab resilient -- performed refresh from all open tabs at once
- Not clock-skew resilient

New algorithm:
- Runs every 10 seconds (`AUTO_REFRESH_TICK_DURATION`) -- tick
- Each tick checks the session from the proper storage medium (not possible to have unsynced state)
- If the expiration is within 3 ticks (`AUTO_REFRESH_TICK_THRESHOLD`) a refresh is attempted
- If a refresh is not successful, another one will be attempted on the next tick
- A failing refresh due to network errors will be retried exponentially but up to a maximum of 10 seconds
- Auto refresh runs only on visible tabs, not on hidden tabs

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -222,7 +222,7 @@ export default class GoTrueClient {
         error: new AuthUnknownError('Unexpected error during initialization', error),
       }
     } finally {
-      this._handleVisibilityChange()
+      await this._handleVisibilityChange()
     }
   }
 
@@ -1141,7 +1141,9 @@ export default class GoTrueClient {
       }
 
       // session will expire in this many ticks (or has already expired if <= 0)
-      const expiresInTicks = Math.floor((now - session.expires_at) / AUTO_REFRESH_TICK_DURATION)
+      const expiresInTicks = Math.floor(
+        (session.expires_at * 1000 - now) / AUTO_REFRESH_TICK_DURATION
+      )
 
       if (expiresInTicks < AUTO_REFRESH_TICK_THRESHOLD) {
         await this._callRefreshToken(session.refresh_token)
@@ -1156,7 +1158,7 @@ export default class GoTrueClient {
    * algorithms when the browser window/tab are in foreground. On non-browser
    * platforms it assumes always foreground.
    */
-  private _handleVisibilityChange() {
+  private async _handleVisibilityChange() {
     if (!isBrowser() || !window?.addEventListener) {
       if (this.autoRefreshToken) {
         // in non-browser environments the refresh token ticker runs always
@@ -1167,24 +1169,39 @@ export default class GoTrueClient {
     }
 
     try {
-      window?.addEventListener('visibilitychange', async () => {
-        if (document.visibilityState === 'visible') {
-          await this.initializePromise
-          await this._recoverAndRefresh()
+      window?.addEventListener(
+        'visibilitychange',
+        async () => await this._onVisibilityChanged(false)
+      )
 
-          if (this.autoRefreshToken) {
-            // in browser environments the refresh token ticker runs only on focused tabs
-            // which prevents race conditions
-            this.startAutoRefresh()
-          }
-        } else if (document.visibilityState === 'hidden') {
-          if (this.autoRefreshToken) {
-            this.stopAutoRefresh()
-          }
-        }
-      })
+      // now immediately call the visbility changed callback to setup with the
+      // current visbility state
+      await this._onVisibilityChanged(true) // initial call
     } catch (error) {
       console.error('_handleVisibilityChange', error)
+    }
+  }
+
+  /**
+   * Callback registered with `window.addEventListener('visibilitychange')`.
+   */
+  private async _onVisibilityChanged(isInitial: boolean) {
+    if (document.visibilityState === 'visible') {
+      if (!isInitial) {
+        // initial visibility change setup is handled in another flow under #initialize()
+        await this.initializePromise
+        await this._recoverAndRefresh()
+      }
+
+      if (this.autoRefreshToken) {
+        // in browser environments the refresh token ticker runs only on focused tabs
+        // which prevents race conditions
+        this.startAutoRefresh()
+      }
+    } else if (document.visibilityState === 'hidden') {
+      if (this.autoRefreshToken) {
+        this.stopAutoRefresh()
+      }
     }
   }
 
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -27,6 +27,8 @@ import {
   resolveFetch,
   setItemAsync,
   uuid,
+  retryable,
+  sleep,
 } from './lib/helpers'
 import localStorageAdapter from './lib/local-storage'
 import { polyfillGlobalThis } from './lib/polyfills'
@@ -79,6 +81,13 @@ const DEFAULT_OPTIONS: Omit<Required<GoTrueClientOptions>, 'fetch' | 'storage'> 
   headers: DEFAULT_HEADERS,
 }
 
+/** Current session will be checked for refresh will be checked at this interval. */
+const AUTO_REFRESH_TICK_DURATION = 10 * 1000
+
+/**
+ * A token refresh will be attempted this many ticks before the current session expires. */
+const AUTO_REFRESH_TICK_THRESHOLD = 3
+
 export default class GoTrueClient {
   /**
    * Namespace for the GoTrue admin methods.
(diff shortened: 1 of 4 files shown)


src/lib/helpers.ts
```diff
@@ -145,3 +145,45 @@ export function decodeJWTPayload(token: string) {
...
+export function retryable<T>(
+  fn: (attempt: number) => Promise<T>,
+  isRetryable: (attempt: number, error: any | null, result?: T) => boolean
+): Promise<T> {
+  const promise = new Promise<T>((accept, reject) => {
+    // eslint-disable-next-line @typescript-eslint/no-extra-semi
+    ;(async () => {
+      for (let attempt = 0; attempt < Infinity; attempt++) {
+        try {
+          const result = await fn(attempt)
```
kangmingtay, line 172: should we call the `sleep` function here instead of inside the `fn` ? i think it should be pretty standard that any request that needs to be retried should follow an exponential backoff

src/GoTrueClient.ts
```diff
@@ -1060,42 +1064,91 @@ export default class GoTrueClient {
...
+   * algorithms when the browser window/tab are in foreground. On non-browser
+   * platforms it assumes always foreground.
+   */
   private _handleVisibilityChange() {
     if (!isBrowser() || !window?.addEventListener) {
+      if (this.autoRefreshToken) {
+        // in non-browser environments the refresh token ticker runs always
+        this.startAutoRefresh()
+      }
+
```
kangmingtay, line 1151: why is there a need to auto refresh if we're recommending the user to create a client for each api route handler on the server-side and not a globally shared one?

kangmingtay (commented):
> Used in-memory state while session state is not in memory -- local storage, cookies, etc.

iirc, the previous algo didn't do this though, also how does the new approach change this?

> Not clock-skew resilient

just based off reading the code, i don't see how the new approach helps to solve this issue either 

> Could produce endless retries

the new algo seems like it can produce endless retries too if the isRetryable function passed into retryable is incorrect right?

Just to clarify, if the `AUTO_REFRESH_TICK_THRESHOLD = 3` and the  `AUTO_REFRESH_TICK_DURATION = 10s`, then 3 ticks would be equivalent to 30s right? 

If you haven't tested this yet, it might be worth [referencing this PR](https://github.com/supabase/gotrue-js/pull/461) and testing it manually to see if it can handle those scenarios.

src/GoTrueClient.ts
hf: This is actually for Flutter, React Native or other types of non-browser (but also non-server) environments.

hf:
> > Used in-memory state while session state is not in memory -- local storage, cookies, etc.
> 
> iirc, the previous algo didn't do this though, also how does the new approach change this?

The previous algorithm started a timeout when a token was refreshed. The timeout is an object in RAM while the access and refresh tokens are objects in local storage (or equivalent). So if the underlying data changes, the timeout would fire based off of wrong assumptions.

With this algorithm, the source-of-truth (local storage or equivalent) is being checked on each tick which allows other independent processes of the tabs to refresh a token without causing race conditions. For example, if you have two open tabs and one does a refresh of the session (with an MFA step up) the other tab in the background will now automatically synchronize. Previously the timeout would fire regardless of the concurrently changed data.

> > Not clock-skew resilient
> 
> just based off reading the code, i don't see how the new approach helps to solve this issue either

This algorithm is skew resilient because it doesn't rely on having an exactly synchronized clock. Ticks are wall time independent, meaning that the wall clock can change between ticks while they still run on 10 second intervals. Furthermore, the tick threshold defines the maximum clock skew acceptable and can easily be changed without running into other clock timing issues.

> > Could produce endless retries
> 
> the new algo seems like it can produce endless retries too if the `isRetryable` function passed into `retryable` is incorrect right?

Think of `retryable` as a separate helper function. The retry mechanism on refresh token is only invoked for a maximum duration of 10 seconds in very specific cases (where there's a fetch error -- typically a network issue). The previous algorithm was problematic because it was easy to make a bug (due to its complexity and duplication).

> Just to clarify, if the `AUTO_REFRESH_TICK_THRESHOLD = 3` and the `AUTO_REFRESH_TICK_DURATION = 10s`, then 3 ticks would be equivalent to 30s right? I'm not entirely convince how moving to the concept of "ticks" would change how things work because we're just creating a new unit of time (every 10seconds is 1 tick) and this complexity would have to be passed down to the developer initialising the client lib which seems unnecessary.

There's nothing passed to the developer. The API and behavior is equivalent. This PR just adds the _ability_ to non-browser developers to properly initialize auto refresh behavior. For example, if you're building a React Native app there's no window `visbilitychanged` event; so even when your screen is off or the app is in the background a refresh timer would fire. Now we're giving those folks the ability to pause auto refresh when the app is in a background state. 

Browsers remain with the same behavior, and this time it's better because it automatically pauses when the tab goes into the hidden state; preventing all open tabs from refreshing with the same token at once.

> If you haven't tested this yet, it might be worth [referencing this PR](https://github.com/supabase/gotrue-js/pull/461) and testing it manually to see if it can handle those scenarios.

Yes.

src/lib/helpers.ts
hf: I'd rather not call any sleeping because this is a helper function designed to easily build retry logic. Not all retries need to have exponential backoff (for example if there's no internet on the device). By keeping this one small we can be certain that it will always behave properly.

src/GoTrueClient.ts
hf: On server-side you don't ever want a ticker like this to be registered. It likely won't run or will run at the worst possible time. We need to make that explicit in docs.

src/GoTrueClient.ts
```diff
@@ -79,6 +81,13 @@ const DEFAULT_OPTIONS: Omit<Required<GoTrueClientOptions>, 'fetch' | 'storage'>
   headers: DEFAULT_HEADERS,
 }
 
+/** Current session will be checked for refresh will be checked at this interval. */
```
J0, line 84:
```suggestion
/** Current session will be checked for refresh  at this interval. */
```
/nit

src/GoTrueClient.ts
J0: Slightly confused here - would users in a server-side environment have to know that they have to set the `autoRefreshToken` option to false  so that the ticker doesn't run

seems like the `!isBrowser` check is `!(typeof window !== 'undefined')`  which looks like it would  be true for server side environments

J0:
> The timeout is an object in RAM while the access and refresh tokens are objects in local storage (or equivalent). So if the underlying data changes, the timeout would fire based off of wrong assumptions.

Do you mind expanding a little more on how the assumptions would change and lead to the timeout firing based off wrong assumptions? Under the impression that the timeout object would still reference objects in local storage (or equivalent) regardless  of where it's located 


> This algorithm is skew resilient because it doesn't rely on having an exactly synchronized clock. Ticks are wall time independent, meaning that the wall clock can change between ticks while they still run on 10 second intervals. Furthermore, the tick threshold defines the maximum clock skew acceptable and can easily be changed without running into other clock timing issues.

Makes sense - quick sanity check - so does this mean that if we switch to 5 second intervals the tolerance would be 5 seconds? (e.g. if NTP indicates it is 10:10:10 and our clock is 4 seconds ahead at 10:10:14 we'd still be in sync albeit with the auto refresh firing 4s ahead of NTP)

J0 (approved):

hf:
> > The timeout is an object in RAM while the access and refresh tokens are objects in local storage (or equivalent). So if the underlying data changes, the timeout would fire based off of wrong assumptions.
> 
> Do you mind expanding a little more on how the assumptions would change and lead to the timeout firing based off wrong assumptions? Under the impression that the timeout object would still reference objects in local storage (or equivalent) regardless of where it's located

Think about a scenario where you have 2 tabs open. Both tabs have the old process for auto refresh started. The user signs in on tab A and a timeout is scheduled in 1 hour to refresh the token. But then the user switches to tab B (where they're already logged in and because of this, tab B also has a timeout scheduled in 1 hour). Now imagine the user clicks sign out on tab B, which clears its timeout and the global local storage. But tab A's timeout is still scheduled, even though there's nothing to refresh.

Thus we have two ways of solving this problem: explicitly synchronizing the state in all open tabs (which is difficult, error-prone and probably less maintainable), or implicitly by never actually having distributed state i.e. all tabs use local storage for all operations and don't keep any derived state in RAM.

kangmingtay (approved):

thorwebdev (approved):

src/GoTrueClient.ts
j4w8n: @kangmingtay @hf,

My apologies for being late to the party here. I read this as @J0 does - that this would also match a server client. And, because we don't have clear docs that server-side clients should set `autoRefreshToken` to `false`, would firing off the `startAutoRefresh` code cause app issues where a server-side client exists with `autoRefreshToken` not explicitly set to `false`?

Not to cross-contaminate this, but it's a close sibling to the issue Gary created [here](https://github.com/supabase/gotrue-js/issues/539); where we need better docs on some of these auth options for a server-side client. Gary would rather throw an error to handle it, but I'm not sure if that's warranted unless there are negative consequences to an app running in these scenarios.

PR #566, opened 2022-12-22
feat: add multi-tab state change notifications

hf:
Uses the [BroadcastChannel](https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API) API to register a new broadcast channel (using the `storageKey` option as the channel's name) which will broadcast state change events from the originating tab to all other open tabs.

This is only available when a browser environment is detected _and session persistence is enabled._ Environments that are not browsers or browsers without session persistence are not able to use this feature since they have no way of synchronizing the underlying session storage.


alaister (approved):
LGTM! Very cool!

kangmingtay (approved):

kangmingtay:
So this should help resolve the issue where a `PASSWORD_RECOVERY` event emitted in one tab will be sent to all other tabs as well right?

J0 (approved):

PR #567, opened 2022-12-22
feat: adds docs to clarify account existance hiding messages

hf:
Adds docs to sign in methods that clarify ambiguous messages when accounts don't exist, email/phone with password combinations are wrong or accounts can only be accessed via social login.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -362,7 +362,7 @@ export default class GoTrueClient {
    * If the `{{ .Token }}` variable is specified in the email template, an OTP will be sent.
    * If you're using phone sign-ins, only an OTP will be sent. You won't be able to send a magiclink for phone sign-ins.
    *
-   * Be aware that you may get back an error message that will not distingish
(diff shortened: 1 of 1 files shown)


awalias (approved):

src/GoTrueClient.ts
```diff
@@ -348,9 +357,14 @@ export default class GoTrueClient {
 
   /**
    * Log in a user using magiclink or a one-time password (OTP).
+   *
    * If the `{{ .ConfirmationURL }}` variable is specified in the email template, a magiclink will be sent.
    * If the `{{ .Token }}` variable is specified in the email template, an OTP will be sent.
    * If you're using phone sign-ins, only an OTP will be sent. You won't be able to send a magiclink for phone sign-ins.
+   *
+   * Be aware that you may get back an error message that will not distingish
```
J0, line 365:
```suggestion
   * Be aware that you may get back an error message that will not distinguish
```

J0 (approved):

PR #570, opened 2022-12-23
fix: remove comments about calling `getUser()` in listFactors()

dshukertjr:
Since `getUser()` is being called within `listFactors()`, users shouldn't have to call `getUser()` by themselves to get the most up-to-date information. 

Also `listFactors()` does not seem to be documented in the [js docs](https://supabase.com/docs/reference/javascript/introduction), but should we add it?

Credit goes to @Vinzent03 for noticing these while working adding MFA support to the Flutter SDK 🎉

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -794,10 +794,8 @@ export interface GoTrueMFAApi {
   challengeAndVerify(params: MFAChallengeAndVerifyParams): Promise<AuthMFAVerifyResponse>
 
   /**
-   * Returns the list of MFA factors enabled for this user. For most use cases
-   * you should consider using {@link
-   * GoTrueMFAApi#getAuthenticatorAssuranceLevel}. This uses a cached version
-   * of the factors and avoids incurring a network call.
+   * Returns the list of MFA factors enabled for this user. This uses a cached version
+   * of the factors retrieved from the user's access token JWT to avoid incurring a network call.
    *
    * @see {@link GoTrueMFAApi#enroll}
    * @see {@link GoTrueMFAApi#getAuthenticatorAssuranceLevel}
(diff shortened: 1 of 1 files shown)


src/lib/types.ts
```diff
@@ -797,8 +797,7 @@ export interface GoTrueMFAApi {
    * Returns the list of MFA factors enabled for this user. For most use cases
    * you should consider using {@link
    * GoTrueMFAApi#getAuthenticatorAssuranceLevel}. This uses a cached version
-   * of the factors and avoids incurring a network call. If you need to update
-   * this list, call {@link GoTrueClient#getUser} first.
+   * of the factors and avoids incurring a network call.
```
kangmingtay, line 800: hmm just wondering why this change is necessary? @dshukertjr

dshukertjr: @kangmingtay Because with [this](https://github.com/supabase/gotrue-js/pull/510) PR, `getUser()` is always called within `listFactors()`, therefore the developer does not have to call `getUser()` first like it is suggested in the comments to update the list.

j4w8n:
@dshukertjr @kangmingtay I think there's some confusion here, and I can understand why.

I believe the code comment for listFactors is saying, "Yeah, you can use this; but most of the time, it's probably better to call 
getAuthenticatorAssuranceLevel. And by the way: if you use getAuthenticatorAssuranceLevel but need to update the list, call getUser first."

Although I'm unsure how calling getUser will update the list, because getUser returns the user without updating the session.

Vinzent03:
I still think the docs need to be updated, because why should you document behavior of `getAuthenticatorAssuranceLevel` in `listFactors`? And you are right, just calling `getUser` won't do anything, because it's not updating the user.

kangmingtay (approved):

Issue #302, opened 2022-06-09
Removing tokens from the URL's hash fragment should be done via the History.replaceState() API


chipilov:
# Bug report

## Describe the bug

Currently, when a session is recovered from the URL's hash fragment, the token is removed from the hash by setting ```window.location.hash``` explicilty: https://github.com/supabase/gotrue-js/blob/30fb5729a1988dbd7f61d58ec10eb81de63f93c7/src/GoTrueClient.ts#L470

However, if the user uses the Back/Forward buttons they can still get to the tokens in the hash.

In order to avoid that, the tokens should be removed using the ```History.replaceState()``` API.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Use a Supabase provided auth link (i.e. a link which contains an access_token in the hash portion, e.g. a magic link);
2. Notice that as soon as the access_token is consumed by the GoTrue-js library, the browser's URL no longer shows the access_token in the hash fragment of the URL;
3. Click on the browser's Back button and then click on the browser's Forward button;

## Expected behavior

The access_token should NOT be visible to the user, even after using the Back/Forward buttons;

## Actual behavior

The access_token is again visible in the browser's URL and the user MIGHT try to re-use it by refreshing the page at the current URL.

## System information

- OS: Windows
- Browser (if applies): Chrome, v102
- Version of supabase-js: 1.35.3
- Version of gotrue-js: 1.22.15

hf:
Will be fixed with #574.

PR #574, opened 2022-12-30
feat: pop implicit flow URL from back stack

hf:
Uses [`window.location.replace()`]() instead of setting the `window.location.hash = ''` to ensure that the implicit flow URL is popped from the back stack and can't be navigated to.

Does not use the `history.replaceState()` API since this fragment is not a history-style fragment.

Fixes: #302.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -775,8 +775,8 @@ export default class GoTrueClient {
       }
       const redirectType = getParameterByName('type')
 
-      // Remove tokens from URL
-      window.location.hash = ''
+      // Remove tokens from URL and popping the URL from the back stack
+      window.location.replace(window.location.href.split('#')[0])
 
       return { data: { session, redirectType }, error: null }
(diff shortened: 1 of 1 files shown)


J0 (approved):

kangmingtay (approved):

Benjamin-Dobell:
This was a _very_ breaking change. Took me quite some time to track this down. The solution proposed by @chipilov was the correct solution i.e. this _should_ be done with `history.replaceState()`.

Currently supabase-js/gotrue-js is causing a redirect during auth callbacks, if we're doing anything in our callback we're abruptly cut off as the page reloads.

sullivan-sean:
This broke OAuth login using `@supabase/auth-helpers` for me entirely. Even some of the simple [examples](https://github.com/supabase/auth-helpers/tree/main/examples/nextjs) in the `auth-helpers` repo are broken after this change

hf:
Reverting the change.

Issue #417, opened 2022-08-28
output of oAuth signin in v2


bhvngt:
# Improve documentation

## Link 
https://supabase.com/docs/reference/javascript/next/auth-signinwithoauth#examples

## Describe the problem

oAuth signin example in the link above suggest that `signInWithOAuth` returns regular `data` object that contains `session`.For ref: it documents following line in the code snippet.

```javascript
const oAuthToken = data.session.provider_token // use to access provider API
```

In `v1` it used to return `User` and `Session` object.

However while trying to migrate I discovered that [OAuthResponse](https://github.com/supabase/gotrue-js/blob/66cebfad24d77f610a92a88d40f5d9bc025cf7c5/src/lib/types.ts#L66) return provider name and `url` returned by the provider. I am not sure if this is intentional or is it work in progress? Because, it seems inconsistent with other auth methods and `v1` and it is also different from what the `v2` documentation is suggesting.

###Is the documentation missing? Or is it confusing? Why is it confusing?

 It is confusing

## Describe the improvement

Ideally, I would prefer if the `signInWithOAuth` returns `data` object of type [AuthResponse](https://github.com/supabase/gotrue-js/blob/66cebfad24d77f610a92a88d40f5d9bc025cf7c5/src/lib/types.ts#L50) to keep it consistent. If not then the example should describe how one can derive `Session` and `User` object from `OAuthResponse` object.

saschwarz:
I'm trying to migrate to v2 and I also can't figure out how to get the authenticated user when using `signInWithOAuth`

ankushg:
Hi, also curious about this -- are there any examples on how to get a `Session`/`User` object after getting an `OAuthResponse`?

Are we supposed to just call `supabase.auth.getSession()` after we get a non-error `OAuthResponse`?

hf:
If I'm not mistaken [`signInWithOAuth`](https://supabase.github.io/gotrue-js/v2/classes/GoTrueClient.html#signInWithOAuth) will basically never return a `data` object in a browser. (At least not in this form.)

This is because the way OAuth sign-in works is the current window location will be set to the authorization URL for the OAuth provider. If calling from a non-browser environment such as React Native, Flutter or other such environments, since there's no `window.location` there you will receive the URL to open in a browser.

I've added an option that will let you access the data object in #575.

saschwarz:
@hf I looked at #575 but there don't seem to be any examples/documentation on how to use that new option to get the session/user?

[...]

PR #575, opened 2022-12-30
feat: add `skipBrowserRedirect` option to `signInWithOAuth`

hf:
Allows for custom handling of the redirect URL when in browser environments.

Fixes: #417


src/GoTrueClient.ts
```diff
@@ -343,6 +343,7 @@ export default class GoTrueClient {
       redirectTo: credentials.options?.redirectTo,
       scopes: credentials.options?.scopes,
       queryParams: credentials.options?.queryParams,
+      skipBrowserRedirect: credentials.options?.skipBrowserRedirect,
```
kangmingtay, line 346: I think this might be slightly confusing for some because some might interpret the "redirect" here as the redirect after the oauth flow has completed when we're actually referring to the redirect to the oauth provider's page.

kangmingtay (approved):

src/GoTrueClient.ts
hf: Yeah let's merge this next year (hehe) to see if I can come up with a better name.

j4w8n: I assume the objective here is to eventually sign the user in via oauth, but not yet? And that the dev is now responsible for triggering the oauth provider url, which is returned.

Other possibilities:
`delayProviderSignIn`
`enableMiddleware` - if that's why this option is being introduced
`autoAuthorization` - would need a default of `true`

kangmingtay: > I assume the objective here is to eventually sign the user in via oauth, but not yet?

Yup, the purpose is to return the oauth provider's URL so the developer can decide how they want to redirect the user to the site. I'd also prefer for the naming to be more explicit such that it encompasses the meaning of the oauth provider URL being returned.

j4w8n: Thanks for the clarification.

>I'd also prefer for the naming to be more explicit such that it encompasses the meaning of the oauth provider URL being returned.

I'm having trouble coming up with verbiage to express this - especially since the URL is returned no matter what.

hf:
I don't think we came up with any better name, so I'll merge it.

PR #581, opened 2023-01-13
feat: setSession triggers SIGNED_IN event

j4w8n:

Introduces a new event trigger for the `setSession()` method.

Calling `setSession()` with a valid and non-expired session triggers no events.

## What is the new behavior?

Calling `setSession()` will trigger a `SIGNED_IN` event, if the session is valid. This will only happen if the session isn't expired - since an expired session would refresh the tokens and trigger a `TOKEN_REFRESHED` event.

## Additional context

Issues and discussions are coming up around this topic. Some implementations, including auth-helpers, rely on events being triggered in order to properly behave. After being involved in some of these discussions, triggering an event with `setSession()` makes sense to me. This is most relevant when using the method on the server-side, to "sign-in" a user - which, in my mind, is really what `setSession()` is doing. That's why I chose the `SIGNED_IN` event as the one to be triggered.

Expo - https://discord.com/channels/839993398554656828/1062010376155316224
React Native - https://discord.com/channels/839993398554656828/1060543536627453952
React/Next/Capacitor - https://github.com/supabase/supabase/discussions/11548

More validation that people understand this method as a way to sign-in a user. See `Expected behavior` section - https://github.com/supabase/supabase-js/issues/681

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -682,6 +682,7 @@ export default class GoTrueClient {
           expires_at: expiresAt,
         }
         await this._saveSession(session)
+        this._notifyAllSubscribers('SIGNED_IN', session)
       }
 
       return { data: { user: session.user, session }, error: null }


kangmingtay (approved):

src/GoTrueClient.ts
```diff
@@ -682,6 +682,7 @@ export default class GoTrueClient {
           expires_at: expiresAt,
         }
         await this._saveSession(session)
+        this._notifyAllSubscribers('SIGNED_IN', session)
```
kangmingtay, line 685: maybe this should be a `TOKEN_REFRESHED` event instead?
```suggestion
        this._notifyAllSubscribers('TOKEN_REFRESHED', session)
```

j4w8n: Thanks for the feedback!

What are your thoughts on why token refreshed is more suitable than signed in?
Based on what I've seen from discussions, people are using this to "sign in" a user. See links on PR. That's what I'm basing it off of. Glad to hear other perspectives though.

kangmingtay: @j4w8n based on the implementation of `setSession`, it takes in a current session and attempts to refresh it. This is closer to a refresh token behaviour and not a signup behaviour. In fact, we should only be sending the `TOKEN_REFRESHED` event if the session was actually refreshed. 

I'm not too sure why users are using setSession to sign-in. I might not fully understand the issue but if they are using this to sign-in their users, why not just use the signInWithPassword or signInWithOtp method directly?

j4w8n: @kangmingtay gotcha.

> based on the implementation of setSession, it takes in a current session and attempts to refresh it.

I'd push back on this. From what I see, the main point of `setSession` is to take the passed-in `currentSession` and save it, so the supabase client can do work on behalf of the user.

I believe the refresh session part of the method is more of a functional courtesy, if the passed-in access token happens to be expired and the refresh token is still valid. In that case, after refreshing the tokens, the method still saves the session to the supabase client, so it can do work. And a `TOKEN_REFRESHED` event is definitely correct here.

But assuming the current session isn't expired, `setSession` is basically doing what something like `signInWithPassword` would do - sign the user in by figuring out the session value and calling `_saveSession(session)`. See [here](https://github.com/supabase/gotrue-js/blob/master/src/GoTrueClient.ts#L334-L335)

> I'm not too sure why users are using setSession to sign-in. I might not fully understand the issue but if they are using this to sign-in their users, why not just use the signInWithPassword or signInWithOtp method directly?

Great question! Here is the explanation for two of the four links I have in the PR description. I've asked the other two for clarification.
Expo - This person is using Expo's Auth Session for authentication, not Supabase. So they acquire the access and refresh tokens from that, and then need a way to tell the supabase client that the user is logged in.
React Native - This person is using the regular sign in methods you referred to. However, they're having issues with the password recovery workflow. As I'm sure you know, this is done by sending a link. Once the link is clicked, the supabase client grabs the info from the url and takes care of everything. However, per Supabase docs, with the React Native use-case, `detectSessionInUrl` is set to `false`. Therefore, they have to grab that info themselves, and also need a way to tell the supabase client that the user is logged in.

This all leads to my conclusion: that assuming the passed-in session to `setSession` is not expired, we should be firing a `SIGNED_IN` event.

kangmingtay: Ah ok i see, that makes alot more sense, thanks for taking the time to explain it! Also i just realised that the `SIGNED_IN` event is only emitted in the `else {}` block and not in the `if {}` block where the session is being refreshed.

kangmingtay (approved):

PR #586, opened 2023-01-16
fix: decodeBase64URL compatibility

j4w8n:

Better compatibility with mobile applications.

The function only works in browser or node environments. React Native folks can work around this by installing a third-party module; but I've been told these don't work with Expo, although I've not verified.

## What is the new behavior?

The function decodes the JWT payload with native javascript; and no longer relies on atob, buffer, or a third-party module.

## Additional context

99% of the code was created by ChatGPT. I tweaked some variable names and had to add an additional check in a couple of spots.

Has been tested against two JWTs from the interwebs and the test suite.

Closes https://github.com/supabase/supabase/issues/11382
Possibly takes care of https://github.com/supabase/supabase-js/issues/669 as well.

### DIFF
--- a/src/lib/helpers.ts
+++ b/src/lib/helpers.ts
@@ -78,28 +78,32 @@ export const removeItemAsync = async (storage: SupportedStorage, key: string): P
   await storage.removeItem(key)
 }
 
-export const decodeBase64URL = (value: string): string => {
-  try {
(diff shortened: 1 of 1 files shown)


kangmingtay (approved):

kangmingtay:
nice @j4w8n, thanks for putting this together!

PR #591, opened 2023-01-25
feat: add pkce

J0:
The corresponding client side implementation for: https://github.com/supabase/gotrue/pull/891

Points of note:

- We don't support `plain` as a method on the client side for now. Users can make use of `plain` on the server side endpoints if they wish.

TODO:
- [x] Add support for devices without a `window`

After PR:
- Update reference spec with example on how to use PKCE


src/GoTrueClient.ts
```diff
@@ -357,11 +357,13 @@ export default class GoTrueClient {
    */
   async signInWithOAuth(credentials: SignInWithOAuthCredentials): Promise<OAuthResponse> {
     await this._removeSession()
+
     return this._handleProviderSignIn(credentials.provider, {
       redirectTo: credentials.options?.redirectTo,
       scopes: credentials.options?.scopes,
       queryParams: credentials.options?.queryParams,
       skipBrowserRedirect: credentials.options?.skipBrowserRedirect,
+      isPKCE: credentials.options?.isPKCE,
```
hf, line 366: Given that PKCE is an implementation detail, and probably the default flow, I'd maybe change the API here to be something like:

```
flowType: 'pkce' | 'implicit'
```

So that people can choose which flow they want, while `pkce` becomes the default flow.

src/GoTrueClient.ts
```diff
@@ -1229,13 +1242,37 @@ export default class GoTrueClient {
...
+    let code = ''
+    //See PKCE Char Set: https://www.oauth.com/oauth2-servers/pkce/authorization-request/
+    // for details on charset and length of code verifier
+    const minLength = 43
+    const maxLength = 128
+    const characters = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-._~"
+    const codeLength = Math.random() * (maxLength - minLength) + minLength
+    let counter = 0;
+    while (counter < codeLength) {
+      code += characters.charAt(Math.floor(Math.random() * characters.length));
```
hf, line 1268: `Math.random()` is not a CSPRNG, so I'm reluctant to use it like this. Can you explore the use of https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/generateKey instead? We can fall back to `Math.random()` if this API is not available.

src/GoTrueClient.ts
```diff
@@ -1229,13 +1242,37 @@ export default class GoTrueClient {
     if (options?.scopes) {
       urlParams.push(`scopes=${encodeURIComponent(options.scopes)}`)
     }
+    if (options?.isPKCE) {
+      const codeVerifier = this._generatePKCECodeVerifier()
+      setItemAsync(this.storage, this.storageKey, codeVerifier)
```
hf, line 1247: Just realizing I'm not a fan of this... given that `setItemAsync` is cookie storage in auth helpers. :/

src/GoTrueClient.ts
```diff
@@ -1229,13 +1242,37 @@ export default class GoTrueClient {
...
+      urlParams.push(`code_verifier=${encodeURIComponent(codeVerifier)}`)
```
hf, line 1248: Instead of this array-based push logic, can you please use `URLSearchParams` as an object? (Also check if this exists in React Native?)

src/GoTrueClient.ts
J0: Sure, let's go with that for now.

Another option would be to fix a default number  which the user can override

src/GoTrueClient.ts
J0: Not sure if it works natively but should be fine [with a polyfill](https://justinnoel.dev/2020/12/08/react-native-urlsearchparams-error-not-implemented/)

src/GoTrueClient.ts
```diff
@@ -365,14 +368,35 @@ export default class GoTrueClient {
...
     })
   }
 
+  /**
+   * Log in an existing user via a third-party provider.
+   */
+  async exchangeAuthCode(authCode: string): Promise<AuthResponse> {
+    // Fetched local hashed verifier
+    let codeVerifier = localStorage.getItem('pkce')
+    // let codeVerifier = await getItemAsync(this.storage, 'pkce')
```
J0, line 387: Left as a comment until we decide on what's a reasonable storage format. There was some mention of moving away from `localStorage` but I am unsure what the consensus is at the moment

package.json
```diff
@@ -36,9 +36,11 @@
     "docs:json": "typedoc --json docs/v2/spec.json --excludeExternals --excludePrivate --excludeProtected src/index.ts"
   },
   "dependencies": {
-    "cross-fetch": "^3.1.5"
+    "cross-fetch": "^3.1.5",
+    "crypto-js": "^4.1.1"
   },
   "devDependencies": {
+    "@types/crypto-js": "^4.1.1",
```
hf, line 43: Do we really need this new dependency? One of the good things about the library is that it's low on dependencies.

src/GoTrueClient.ts
```diff
@@ -1335,10 +1377,19 @@ export default class GoTrueClient {
     if (options?.scopes) {
       urlParams.push(`scopes=${encodeURIComponent(options.scopes)}`)
     }
+    if (options?.flowType && options.flowType === 'pkce') {
```
hf, line 1380: You can just use 
```suggestion
    if (options?.flowType === 'pkce') {
```

src/GoTrueClient.ts
```diff
@@ -1335,10 +1377,19 @@ export default class GoTrueClient {
...
+      const codeVerifier = await generatePKCEVerifier()
+      await setItemAsync(this.storage, `${this.storageKey}-oauth-code-verifier`, codeVerifier)
+      const codeChallenge = await generatePKCEChallenge(codeVerifier)
+      urlParams.push(`code_challenge=${encodeURIComponent(codeChallenge)}`)
+      urlParams.push(`code_challenge_method=${encodeURIComponent('s256')}`)
+    }
     if (options?.queryParams) {
       const query = new URLSearchParams(options.queryParams)
       urlParams.push(query.toString())
     }
```
hf, line 1391: Why not just use `new URLSearchParams()` to build the params, instead of these string operations?

src/lib/helpers.ts
```diff
@@ -236,3 +236,44 @@ export function retryable<T>(
...
+    }
+  }
+  return Array.from(array, dec2hex).join('')
+}
+
+async function sha256(randomString: string) {
+  const encoder = new TextEncoder()
+  const encodedData = encoder.encode(randomString)
+  if (!isBrowser()) {
+    return sha256CryptoJS(randomString).toString()
```
hf, line 260: Ah OK this is where it's used. OK nvm the comment.

src/lib/helpers.ts
```diff
@@ -236,3 +236,44 @@ export function retryable<T>(
...
+  }
+  const hash = await window.crypto.subtle.digest('SHA-256', encodedData)
+  let hashedString = ''
+  const bytes = new Uint8Array(hash)
+  const len = bytes.byteLength
+  for (let i = 0; i < len; i++) {
+    hashedString += String.fromCharCode(bytes[i])
+  }
```
hf, line 268: This is very inefficient. It's best to convert the `bytes` to a regular JS array and then `.map` over it converting the bytes to chars and then `.join`. It's quite a bit of string concatenations here so it does pay off.

src/lib/helpers.ts
```diff
@@ -236,3 +236,44 @@ export function retryable<T>(
...
+  return hashedString
+}
+
+function base64urlencode(str: string) {
+  return btoa(str).replace(/\+/g, '-').replace(/\//g, '_').replace(/=+$/, '')
+}
```
hf, line 274: I feel like this function is already in helpers somewhere.

src/lib/types.ts
```diff
@@ -443,6 +444,8 @@ export type SignInWithOAuthCredentials = {
     queryParams?: { [key: string]: string }
     /** If set to true does not immediately redirect the current browser context to visit the OAuth authorization page for the provider. */
     skipBrowserRedirect?: boolean
+    /** If Set to true uses the PKCE flow */
```
hf, line 447: Docs seem wrong.

src/lib/helpers.ts
J0: Do you recall where exactly? There's a `decodebase64URL`  and a regex but can't seem to find a function which encodes

src/lib/helpers.ts
J0: Gotcha,  will convert

src/GoTrueClient.ts
```diff
@@ -1004,24 +1037,31 @@ export default class GoTrueClient {
...
       scopes?: string
       queryParams?: { [key: string]: string }
       skipBrowserRedirect?: boolean
+      flowType?: OAuthFlowType
     } = {}
   ) {
-    const url: string = this._getUrlForProvider(provider, {
+    if (!options.flowType) {
+      options.flowType = 'implicit'
+    }
```
kangmingtay, line 1052:
```suggestion
```

kangmingtay: can remove this if you choose to set the default value of `flowType` in the suggestion i made earlier for `signInWithOAuth`

src/lib/types.ts
```diff
@@ -443,6 +444,8 @@ export type SignInWithOAuthCredentials = {
     queryParams?: { [key: string]: string }
     /** If set to true does not immediately redirect the current browser context to visit the OAuth authorization page for the provider. */
     skipBrowserRedirect?: boolean
+    /** If set to 'pkce' PKCE flow. Defaults to the 'implicit' flow otherwise */
+    flowType?: OAuthFlowType
```
kangmingtay, line 448: doesn't seem clear that it defaults to `implicit` because it's only set as a default in `_handleProviderSignIn()` so this might be abit confusing for someone looking through the code for the first time - would be slightly clearer with the suggestion above

src/GoTrueClient.ts
```diff
@@ -365,14 +369,43 @@ export default class GoTrueClient {
...
   async signInWithOAuth(credentials: SignInWithOAuthCredentials): Promise<OAuthResponse> {
     await this._removeSession()
-    return this._handleProviderSignIn(credentials.provider, {
+
+    return await this._handleProviderSignIn(credentials.provider, {
       redirectTo: credentials.options?.redirectTo,
       scopes: credentials.options?.scopes,
       queryParams: credentials.options?.queryParams,
       skipBrowserRedirect: credentials.options?.skipBrowserRedirect,
+      flowType: credentials.options?.flowType,
```
kangmingtay, line 378:
```suggestion
      flowType: credentials.options?.flowType ?? 'implicit',
```

kangmingtay (approved):
overall, lgtm! would be good if @thorwebdev can also drop his thoughts on the DX side of things

thorwebdev (approved):
LGTM 👍

src/lib/helpers.ts
```diff
@@ -236,3 +236,42 @@ export function retryable<T>(
...
+
+// Functions below taken from: https://stackoverflow.com/questions/63309409/creating-a-code-verifier-and-challenge-for-pkce-auth-on-spotify-api-in-reactjs
+export async function generatePKCEVerifier() {
+  const verifierLength = 56
+  const array = new Uint32Array(verifierLength)
+  if (!isBrowser()) {
+    for (let i = 0; i < verifierLength; i++) {
+      array[i] = Math.floor(Math.random() * 256)
+    }
+  }
```
dshukertjr, line 252: @J0 
Trying to implement PKCE in Flutter library as well, and had a question here. 

What is the reason why we need to limit this to only run in browser?

dshukertjr: Are these codes expected to ever run in a server environment?

dshukertjr: Makes sense! Thanks!

PR #598, opened 2023-01-31
feat: keep expired session on initialization

hf:
If a client is constructed when the underlying access token has expired and auto refresh is disabled, the session is getting removed. It should not be removed as the session can be automatically refreshed either with the auto refresh mechanism or whenever `getSession` is called next.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -998,8 +998,6 @@ export default class GoTrueClient {
             console.log(error.message)
             await this._removeSession()
           }
-        } else {
-          await this._removeSession()
         }
       } else {
         if (this.persistSession) {


j4w8n:
> If a client is constructed when the underlying access token has expired and auto refresh is enabled, the session is getting removed. 

I believe this is wrong. The session only gets removed if the token is expired and auto refresh is disabled, aka `false` - not enabled, as you've stated.

However, I can understand why this code should be removed. In a scenario where a session, in storage, is expired and `autoRefreshToken` is set to `false` - common with mobile apps - the dev may still have code that calls `getSession`, `setSession` or `refreshSession`. And assuming the refresh token is still valid, the session will be renewed with one of those calls. So, there doesn't seem to be a pressing need to remove an expired session within the `recoverAndRefresh` function.

j4w8n:
As an aside, while I was reading through the code, trying to wrap my head around it, I realized that line 994 only needs to check for `this.autoRefreshToken`. if `currentSession.refresh_token` isn't true, then it would've been caught on line 983.

Has nothing to do with your PR, but can be tweaked at some point, if desired.

david-plugge:
@j4w8n we ran into this issue with the auth-helpers. Since we are in a server environment we have to set `autoRefreshToken` to false but if a request is made and the session is valid for less than 10 seconds (EXPIRY_MARGIN) the user just gets logged out. To get this working we basically have to delete the cookie ONLY when we explicitly want to logout the user or the refresh token is expired. This can easily be done by setting the cookie lifetime to the same value though.

hf:
> > If a client is constructed when the underlying access token has expired and auto refresh is enabled, the session is getting removed.
> 
> I believe this is wrong. The session only gets removed if the token is expired and auto refresh is disabled, aka `false` - not enabled, as you've stated.

Yup, typo. Edited the description. Thanks for catching it!

hf:
@david-plugge Mind a review?

kangmingtay (approved):

j4w8n:
@kangmingtay @david-plugge I'm starting to wonder if there could be some fallout from this; especially when using auth helpers. I don't know the auth helper code in detail, but I'm thinking this change - although I believe it needed done - could cause issues.

Correct me if I'm wrong:

before this change, the session - aka cookie in the case of auth helpers - would be removed during the initialization of any client if the access token is expired (or close to it) and `autoRefreshToken` is set to `false`. This had the undesired effect of essentially logging out a user, since the cookie is deleted.

after this change, the session is going to get passed into the server-side client; but it'll be expired. So if a dev goes to make a server-side db query, it'll fail because the access token is expired. I couldn't figure out in the auth helper code where a session would be refreshed. Am I just not seeing it?

kangmingtay:
@j4w8n yeah #681 is definitely a breaking change but it seems like a bug that we didn't catch. We'll need to fix the auth-helpers too when we merge it in. The idea is that if you set `autoRefreshToken: false`, you should manage all the session refreshes on your own.

j4w8n:
> @j4w8n yeah #681 is definitely a breaking change but it seems like a bug that we didn't catch. We'll need to fix the auth-helpers too when we merge it in. The idea is that if you set `autoRefreshToken: false`, you should manage all the session refreshes on your own.

Yep, I agree.

PR #436, opened 2022-09-09
fix: fix storage for browsers without local storage

pixtron:
## What kind of change does this PR introduce?

Bug fix

## What is the current behavior?

There are some edge cases where a browser does have `window` defined but not `globalThis.localStorage`. Currently gotrue-js will throw in these environments.
See #433 

## What is the new behavior?
Before accessing `globalThis.localStorage` it is verified that `globalThis.localStorage` is not undefined.

## Additional context

Wasn't really able to test this change so far, therefore opening as draft.

PR #600, opened 2023-02-02
feat: add enhanced `localStorage` support check

hf:
Based on feedback from #436.

### DIFF
--- a/src/lib/helpers.ts
+++ b/src/lib/helpers.ts
@@ -15,6 +15,51 @@ export function uuid() {
 
 export const isBrowser = () => typeof document !== 'undefined'
 
+const localStorageWriteTests = {
+  tested: false,
+  writable: false,
+}
+
+/**
+ * Checks whether localStorage is supported on this browser.
+ */
+export const supportsLocalStorage = () => {
+  if (!isBrowser()) {
+    return false
+  }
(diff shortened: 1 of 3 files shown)


src/lib/helpers.ts
```diff
@@ -15,6 +15,51 @@ export function uuid() {
...
+  try {
+    if (typeof globalThis.localStorage !== 'object') {
+      return false
+    }
+  } catch (e) {
+    // DOM exception when accessing `localStorage`
+    return false
+  }
+
+  if (localStorageWriteTests.tested) {
```
J0, line 40: When will this be true?

J0 (approved):
LGTM, thanks!

tjessessky:
Sorry, I'm going to answer my own comment in case anyone finds themselves with the same problem I had. I wrote a gist here with sample code, but the basic fix is that you can pass a custom Storage class object into the createClient options, and use whatever storage methods you'd like. [See the code here](https://gist.github.com/tjessessky/87ad344ae39206a3063425c79294da6d).

src/lib/helpers.ts
hf: It's set right below when the write test finishes.

PR #611, opened 2023-02-15
feat: allow metadata to be passed in during oauth sign ins as custom …

J0:
Complementary PR for [passing in oauth metadata](https://github.com/supabase/gotrue/pull/943).

[Demo](https://share.cleanshot.com/SnqxPMSn)


J0:
As discussed, we won't be going ahead with this for the implicit flow as it'd involve passing metadata (which might be sensitive) via a query param. This query param can potentially be logged by a server.

We'll consider supporting this in the PKCE flow which should be released soon

PR #616, opened 2023-02-23
feat: add support for whatsapp programmable messaging API

J0:
We wish to add support for the [Whatsapp Programmable Messaging API](https://www.twilio.com/docs/whatsapp/api) for Twilio and possibly for other providers in the near future.

This would allow users to send one time verification codes over Whatsapp instead of SMS which is arguably a more secure channel

Methods affected:
- `supabase.auth.signInWithPassword`
- `supabase.auth.signInWithOTP`

How to test

1. Clone  gotrue-js and checkout `j0/support-whatsapp`
2. Clone gotrue and checkout `j0/support-whatsapp`
3. Go to your app and change dependency to point to point to gotrue-js

```
supabsae.auth.signInWithOtp({ phone: '<your-special-number-with-international-code>', options: {
        channel: 'sms'}}))
```

```
supabase.auth.signInWithPassword({
   phone: '<your-special-number-with-international-code>',
   options: {
       channel: 'whatsapp'
   }
})
```

Note that if you supply an invalid channel or an API which doesn't support whatsapp it will default to the current behaviour which is SMS

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -422,6 +422,7 @@ export type SignInWithPasswordlessCredentials =
         data?: object
         /** Verification token received when the user completes the captcha on the site. */
         captchaToken?: string
+        channel?: string
       }
     }
 
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -387,7 +387,7 @@ export type SignInWithPasswordCredentials =
         /** Verification token received when the user completes the captcha on the site. */
         captchaToken?: string
         /** Messaging channel to use (e.g. whatsapp or sms) */
-        channel?: string
+        channel?: 'sms' | 'whatsapp'
       }
     }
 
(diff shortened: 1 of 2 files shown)


src/lib/types.ts
```diff
@@ -386,6 +386,8 @@ export type SignInWithPasswordCredentials =
         data?: object
         /** Verification token received when the user completes the captcha on the site. */
         captchaToken?: string
+        /** Messaging channel to use (e.g. whatsapp or sms) */
+        channel?: string
```
hf, line 390: Can we use `channel?: 'sms' | 'whatsapp'` instead?

src/lib/types.ts
```diff
@@ -422,6 +424,8 @@ export type SignInWithPasswordlessCredentials =
...
```
hf, line 428: Same.

src/lib/types.ts
hf: I would love for some docs on the method itself saying that if you want to use WhatsApp you need to configure it first.

J0: I will also follow up with a docs PR describing setup on supabase/supabase

kangmingtay (approved):
LGTM, but let's only merge this in once we have deployed the gotrue version with https://github.com/supabase/gotrue/pull/981 to all projects

J0:
Going to merge this PR but not bump Supabase-js version so that we can unblock customer. Customer will install branch of supabase-js with upgraded GoTrue-js

PR #627, opened 2023-03-16
feat: invoke callback on onAuthStateChange

### LINKED ISSUE #313: Call auth.onAuthStateChange at least once when listener is registered.
# Feature request
Call `auth.onAuthStateChange()` at least once when registering the listener.

When getting the current user, I have to first get the user using `auth.user()`, then add a listener using `auth.onAuthStateChange()` and implement the same code there.

I feel like it would be better if `auth.onAuthStateChange()` is called at least once when registering the listener, this removes unnecessary duplicate or spaghetti code. Additionally, this is also how firebase does it: https://firebase.google.com/docs/reference/android/com/google/firebase/auth/FirebaseAuth.AuthStateListener

There is a solution as mentioned above, but doesn't result in the cleanest code.

## Additional context
None.

kangmingtay:
* Run callback when `onAuthStateChange` is first invoked with an `INITIAL_SESSION` event
* Fixes #313

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -941,15 +941,24 @@ export default class GoTrueClient {
 
     this.stateChangeEmitters.set(id, subscription)
 
-    this.getSession()
-      .then(({ data: { session } }) => {
-        this.stateChangeEmitters.get(id)?.callback('INITIAL_SESSION', session)
-      })
-      .catch(() => {
-        this.stateChangeEmitters.get(id)?.callback('INITIAL_SESSION', null)
-      })
+    this.emitInitialSession(id)
 
     return { data: { subscription } }
+  }
+
+  private async emitInitialSession(id: string): Promise<void> {
+    try {
+      const {
+        data: { session },
+        error,
+      } = await this.getSession()
+      if (error) throw error
+
+      this.stateChangeEmitters.get(id)?.callback('INITIAL_SESSION', session)
+    } catch (err) {
+      this.stateChangeEmitters.get(id)?.callback('INITIAL_SESSION', null)
+      console.error(err)
+    }
   }
 
   /**
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -941,6 +941,14 @@ export default class GoTrueClient {
 
     this.stateChangeEmitters.set(id, subscription)
 
+    this.getSession()
+      .then(({ data: { session } }) => {
+        this.stateChangeEmitters.get(id)?.callback('INITIAL_SESSION', session)
+      })
+      .catch(() => {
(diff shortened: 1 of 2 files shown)


src/GoTrueClient.ts
```diff
@@ -941,6 +941,14 @@ export default class GoTrueClient {
 
     this.stateChangeEmitters.set(id, subscription)
 
+    this.getSession()
```
hf, line 944: Can we use async/await and log the error?

src/lib/types.ts
```diff
@@ -23,6 +23,7 @@ export type Provider =
 export type AuthChangeEventMFA = 'MFA_CHALLENGE_VERIFIED'
 
 export type AuthChangeEvent =
+  | 'INITIAL_SESSION'
```
hf, line 26: Add tsdoc somewhere about it

src/GoTrueClient.ts
kangmingtay: don't think we can do that because `onAuthStateChange` is a synchronous method

alaister: async/await is syntax sugar anyway. We can log the error if you like, but tbh the result of an error is a logged-out user, so this may not be the place for a log.

hf: Well that's if the error is not caused by the library. If it's like a `TypeError` this will just fail silently. I'd still insist on the logging so we're not wrecking our brains. Regarding async/await, you can refactor this into a new async method like `emitInitialSession()` which is async but is not being awaited in onAuthStateChange.

kangmingtay: > Regarding async/await, you can refactor this into a new async method like emitInitialSession() which is async but is not being awaited in onAuthStateChange.

@hf hmm can you clarify what you mean by this? `getSession` is already an async method and it has proper error handling in it. Or are you suggesting we create an async version of `onAuthStateChange` instead?

src/lib/types.ts
kangmingtay: will document it in https://supabase.com/docs/reference/javascript/auth-onauthstatechange

src/GoTrueClient.ts
soedirgo: You could also accept an arg on the `.catch()` callback and log that

soedirgo:
```ts
      .catch((error) => {
         console.error(error)
         this.stateChangeEmitters.get(id)?.callback('INITIAL_SESSION', null)
       })
```

soedirgo (approved):

TheThirdRace:
@kangmingtay 

Thanks for this one! 🚀

I was managing it manually, but this is so much more useful when integrated in Supabase directly.

Issue #349, opened 2022-02-10
Password Reset Flow not working as documented


djemrose:
# Bug report

## Describe the bug

As heavily discussed there is a bug that logs the user in when they go to reset their password: https://github.com/supabase/supabase/discussions/3360

The `onAuthStateChange` shows this happening by firing the `SIGNED IN` event before the `PASSWORD_RECOVERY` event. I would imagine this is undesirable for everyone. This also causes the hash-bang url to get cleared in the browser immediately on page load, which prevents some apps from grabbing the documented `access_token` for later use.

## To Reproduce

1. Follow the official guide here whilst listening to the `onAuthStateChange` event: https://supabase.com/docs/reference/javascript/auth-api-resetpasswordforemail
2. Also witness the hash-bang url being cleared on page load.

## Expected behavior

1. The hashbang url to not be cleared
2. Only the PASSWORD_RECOVERY event to fire.
3. No SIGNED_IN event should fire on page load.

## Screenshots

![Screen Shot 2022-02-10 at 8 56 21 pm](https://user-images.githubusercontent.com/363101/153383318-fce08269-bce4-4ee3-aa3e-9cf6fc326aa9.png)

## System information

- OS: macOS
- Browser Chrome
- Version of supabase-js: @supabase/supabase-js 1.1.2
- Version of Node.js: v16.13.1

funwithtriangles:
Having a similar issue, I just get `SIGNED_IN` instead of `PASSWORD_RECOVERY`

funwithtriangles:
FWIW I fixed my issue by avoiding the auth event altogether and just making sure the link redirects the user to the correct page:

```ts
await supabase.auth.resetPasswordForEmail(email, {
    redirectTo: `${window.location.origin}/reset-password`,
  })
```

hf:
Please avoid relying on the sequence of events fired in the `onAuthStateChange` callback as the order is never guaranteed.

You can always use [`getSession()`](https://supabase.github.io/gotrue-js/v2/classes/GoTrueClient.html#getSession) to safely access the latest access token.

laygir:
I'm quite new to Supabase, I've been hearing it a lot and wanted to give it a go in a side project. 
But seeing this issue open for more than a year is a bit surprising considering its impact, but hopefully the fix is on the way I believe 🤞  #629 

[...]

PR #629, opened 2023-03-21
feat: dont fire `SIGNED_IN` event on `PASSWORD_RECOVERY`

kangmingtay:
* Only fire either the `PASSWORD_RECOVERY` event or the `SIGNED_IN` event instead of firing both when a password recovery link is clicked
* Mentioned in: https://github.com/supabase/gotrue-js/issues/349

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -220,9 +220,10 @@ export default class GoTrueClient {
         await this._saveSession(session)
 
         setTimeout(() => {
-          this._notifyAllSubscribers('SIGNED_IN', session)
           if (redirectType === 'recovery') {
             this._notifyAllSubscribers('PASSWORD_RECOVERY', session)
+          } else {
+            this._notifyAllSubscribers('SIGNED_IN', session)
           }
         }, 0)
 


kiwicopple:
could this be a breaking change, if someone is listening to an event which is no longer being fired?

kangmingtay:
Yeah we are firing the `SIGNED_IN` event on every other email link so i don't see how one can use it to differentiate a password recovery link from the rest. But that being said, it is possible that someone might be running some callbacks as long as there is an authenticated user detected. The tradeoff here is that this PR would fix an issue with the password recovery flow for majority of users but potentially break someone who is possibly relying on the `SIGNED_IN` event only for password recovery - which seems weird and not ideal anyway.

silentworks:
@kiwicopple Yes this could be seen as a breaking change, but the current behaviour feels more like a bug so I would consider this fixing a bug rather than a break in its correct usage.

silentworks (approved):

J0 (approved):

s0h311:
Hi guys, I'm still having this issue. No PASSWORD_RECOVERY. 

nuxtjs/supabase 0.3.1
nuxt 3.4.1

kangmingtay:
@s0h311 do you have a reproducible example that we can check out?

s0h311:
@kangmingtay the repo is private, I'm using the code within a middleware:

```JavaScript
export default defineNuxtRouteMiddleware((to, from) => {
  const l = useLocalePath()
  const supabase = useSupabaseClient()
  const isRecovery = ref(false)
  supabase.auth.onAuthStateChange((event, _session) => {
    if (event == 'PASSWORD_RECOVERY') isRecovery.value = true
  })

  if (isRecovery.value) return
  else return navigateTo(l('/'))
})
```

kangmingtay:
Hi @s0h311, seems like the reason you're not receiving the `PASSWORD_RECOVERY` event is because you're always creating a new instance of the supabase client in the middleware. This means that you're always registering a new subscriber (through `onAuthStateChange`) on a new instance of the supabase client.

s0h311:
@kangmingtay thank you, defining a nuxt plugin that returns a supabse instance solved it for me.

*plugins/helpers.ts*
```TypeScript
export default defineNuxtPlugin(() => {
  return {
    provide: {
      supabase: useSupabaseClient(),
    },
  }
})
```

*middleware/password-recovery-js*
```TypeScript
export default defineNuxtRouteMiddleware((to, from) => {
  const { $supabase } = useNuxtApp()
  const isRecovery = ref(false)
  $supabase.auth.onAuthStateChange((event, _session) => {

    if (event == 'PASSWORD_RECOVERY') isRecovery.value = true
  })

  if (isRecovery.value) return
  else return navigateTo('/')
})

```

alexbriannaughton:
Hi y'all, still having trouble with this; getting INITIAL_SESSION instead of the other two events. to repro:
-clone the repo at https://github.com/alexbriannaughton/expo-router-with-supabase-auth
-run npm install in terminal
-create a .env that looks like this:
EXPO_PUBLIC_SUPABASE_URL=(your supabase url here)
EXPO_PUBLIC_SUPABASE_ANON_KEY=(your anon key here)
-run npx expo start in terminal
-type email (that already exists in your database) into email text box and click 'forgot password' button
-click link that gets sent to email
-see that INITIAL_SESSION event is emitted

PR #631, opened 2023-03-22
feat: add resend method

kangmingtay:
* Adds the client lib binding for the `resend` method
* For example, to resend a signup confirmation email, one can now call `supabase.auth.resend({ email, type: "signup"})`.
* To resend password recovery or magiclink emails, you should just call `supabase.auth.resetPasswordForEmail` and `supabase.auth.signInWithOtp` respectively again.
* Resending invites can be done by calling the `supabase.auth.admin.inviteUserByEmail` method


J0 (approved):

J0:
Let's also update reference spec when free

m1dok:
Hi ! see PR #631 , they added `supabase.auth.resend({ email, type: "signup"})` call.

you can check if the response contains an identities array not empty, and session null means the mail has not been verified, and after user input use this call

PR #636, opened 2023-03-30
fix: pkce does not generate truly random values in a browser

J0:
As @dshukertjr pointed out, the current version of `generatePKCEVerifier()` has the random generation deleted. There should be an additional branch which calls `getRandomValues()` in order to truly set random values.

When the user is in a browser, the array defaults to the randomly initialized values as set by RAM which is not ideal.

## What is the new behavior?
Generate random values.


kangmingtay (approved):

dshukertjr (commented):
Not directly related to this PR, but we can remove the `async` keyword on `generatePKCEVerifier` here, right?

dshukertjr:
I think I saw few other places that we can remove async from. Let me create a PR

src/lib/helpers.ts
```diff
@@ -249,6 +249,8 @@ export async function generatePKCEVerifier() {
     for (let i = 0; i < verifierLength; i++) {
       array[i] = Math.floor(Math.random() * 256)
     }
+  } else {
+    window.crypto.getRandomValues(array)
```
dshukertjr, line 253: Do we need to make `array` `var` and assign it here?
```suggestion
    array = window.crypto.getRandomValues(array)
```

dshukertjr: Ah, sorry, never mind. Just verified that the current code works!

J0:
I think some of the `window` methods are async, so you might not be able to remove some of them but if you find any feel free to file a PR.  I'll also take a look later

PR #638, opened 2023-03-31
fix: remove async from `generatePKCEVerifier` and fix parameter of `_handleProviderSignIn`

dshukertjr:

- Removes unnecessary `async` keyword on `generatePKCEVerifier`, and the `await` from where `generatePKCEVerifier` was called.
- Fix the type of the parameter of `_handleProviderSignIn` to remove the following error.
Screenshot 2023-03-31 at 16 17 01


J0 (approved):
Thanks for putting this together!

PR #640, opened 2023-04-01
Replacing `crypto-js` with the native `ohash`

elonsalfati:
This PR solves the bug below when running with `nuxt` and `supabase`. 

```
Uncaught SyntaxError: The requested module '/_nuxt/node_modules/crypto-js/sha256.js?v=a4564608' does not provide an export named 'default'
```

As shown in [this discussion](https://github.com/orgs/supabase/discussions/13340)

The bug completely breaks the use of `supabase` in the latest version with nuxt.

## What is the new behavior?

Everything works :-)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,8 @@
       "version": "0.0.0",
       "license": "MIT",
       "dependencies": {
-        "cross-fetch": "^3.1.5"
+        "cross-fetch": "^3.1.5",
+        "ohash": "^1.0.0"
       },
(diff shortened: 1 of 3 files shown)


elonsalfati:
I've pushed another commit replacing it again with [ohash](https://github.com/unjs/ohash), but let me know if the native one makes more sense - I can revert to the previous commit.

silentworks:
We've fixed this in #641 by removing the external dependency.

PR #641, opened 2023-04-03
fix: remove crypto-js

J0:

Removes crypto-js entirely as users have reported breaking behaviour when used together with Nuxt. Additionally, it seems like subtle crypto is supported on a substantial number of frameworks (e.g. Deno, Chrome extension under https). As such, we'll be removing it and advising users on platforms without native access to the WebCryptoAPI (e.g. React Native) to use a polyfill


silentworks (approved):
LGTM!

src/lib/helpers.ts
```diff
@@ -245,21 +244,18 @@ function dec2hex(dec: number) {
...
   const verifierLength = 56
   const array = new Uint32Array(verifierLength)
-  if (!isBrowser()) {
-    for (let i = 0; i < verifierLength; i++) {
-      array[i] = Math.floor(Math.random() * 256)
-    }
-  } else {
-    window.crypto.getRandomValues(array)
+  if (typeof window.crypto === 'undefined') {
+    throw 'PKCE is not supported on devices without WebCrypto API support or equivalent polyfills'
```
thorwebdev, line 248: wouldn't this need to be `throw new Error('PKCE is not supported on devices without WebCrypto API support or equivalent polyfills')`?

daniel-xyz:
Just found out about this change since my React Native/Expo app started to show the warning. Not a big deal for me, but shouldn't this be a breaking change since it affects security?

PR #646, opened 2023-04-05
refactor: make pkce option a client setting

J0:
Refactor pkce to be a client setting so it doesn't have to be configured on invocation for each method.

- also call `exchangeSessionForCode` in the background if `detectSession` is enabled
- also remove the `flowType` option from signInWithOAuth


src/GoTrueClient.ts
```diff
@@ -920,6 +935,16 @@ export default class GoTrueClient {
         Boolean(getParameterByName('error_description')))
     )
   }
+  /**
+   * Checks if the current URL and backing storage contain parameters given by a PKCE flow
+   */
+  private async _isPKCEFlow(): Promise<boolean> {
+    const currentStorageContent = await getItemAsync(
+      this.storage,
+      `${this.storageKey}-code-verifier`
```
thorwebdev, line 944:
```suggestion
      `${this.storageKey}-oauth-code-verifier`
```

thorwebdev (approved):
One storage naming issue in `_isPKCEFlow` otherwise LGTM, thanks.

src/GoTrueClient.ts
J0: Ah, good catch! Have renamed everything else to `code-verifier` instead so as to keep it generic

J0:
Leaving this with a refactor: prefix so semver doesn't trigger another release as I'd like to get another three PRs in 
1. remove OAuth `flowType` param 
2. add magic link binding
3. add recover bindings
Before making another release

PR #647, opened 2023-04-05
fix: move  channel parameter from sign in to sign up

J0:
The `channel` parameter for WhatsApp should be on the  `Signup` method instead of the `SignIn` method

Relevant Server side code: https://github.com/supabase/gotrue/blob/master/internal/api/signup.go#L22


src/lib/helpers.ts
```diff
@@ -245,7 +245,9 @@ export function generatePKCEVerifier() {
   const verifierLength = 56
   const array = new Uint32Array(verifierLength)
   if (typeof window.crypto === 'undefined') {
-    throw new Error('PKCE is not supported on devices without WebCrypto API support or equivalent polyfills')
+    throw new Error(
+      'PKCE is not supported on devices without WebCrypto API support or equivalent polyfills'
```
hf, line 249:
```suggestion
      'PKCE is not supported on devices without WebCrypto API support, please add polyfills'
```

src/lib/helpers.ts
```diff
@@ -255,7 +257,9 @@ async function sha256(randomString: string) {
   const encoder = new TextEncoder()
   const encodedData = encoder.encode(randomString)
   if (typeof window.crypto === 'undefined') {
-    throw new Error('PKCE is not supported on devices without WebCrypto API support or equivalent polyfills')
+    throw new Error(
+      'PKCE is not supported on devices without WebCrypto API support or equivalent polyfills'
```
hf, line 261:
```suggestion
      'PKCE is not supported on devices without WebCrypto API support, please add polyfills'
```

hf (approved):

dshukertjr:
@J0 Do we need the `channel` parameter support for `resend()` method as well?

J0:
@dshukertjr sorry just seeing this. Don't think we need it at the moment as it is not on the API. Let us know if customers request for it though and we can add it.

PR #650, opened 2023-04-08
feat: add multi-tab lock for refreshing tokens

hf:
Adds a distributed multi-tab spin lock based on `BroadcastChannel`. The protocol is as follows:

1. Any tab wanting to acquire a lock listens and replies with the received message.
2. A tab can assume that the lock is acquired by them if and only if it has not received a reply back after waiting for X amount of time.

This is added because there are some (hard to reproduce) cases where the same refresh token is refreshed in parallel by two different tabs, which can lead to randomly occurring `Invalid Refresh Token` errors due to the token already being reused.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -25,6 +25,7 @@ import {
   sleep,
   generatePKCEVerifier,
   generatePKCEChallenge,
+  multiTabLock,
 } from './lib/helpers'
 import localStorageAdapter from './lib/local-storage'
 import { polyfillGlobalThis } from './lib/polyfills'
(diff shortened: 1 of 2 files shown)


krkd:
Great work. Was wondering if anyone else got app launching multiple refresh requests at once. 

There is already some synchronisation primitive that i am not fully aware [what exactly it does](https://github.com/supabase/gotrue-js/blob/master/src/GoTrueClient.ts#L832). But from the first look it tries to serialise access to shared storage eg do consecutive actions agains shared resource. Which may just lead to refreshToken being called twice not simultaneously but one right after another. i believe same code can be extended/modified in order to implement Event synchronisation primitive which will allow multiple runtimes(tabs) to wait on the result of one call to refreshToken/getSession/getUser etc.

PR #651, opened 2023-04-08
feat: increase auto refresh tick duration to 30s from 10s

hf:
It appears that for unknown reasons the `/token` endpoint in GoTrue may last significantly longer than 10 seconds. This would mean that a repeat request could be sent out while the first one has not finished.

Longer request times than 30 seconds are generally not possible on the modern web, as many CDNs and proxies default time-out responses to 30 seconds.


J0 (approved):
LGTM, this is a temp fix and we'll be looking into issues with the `/token` endpoint after right?

PR #655, opened 2023-04-10
fix: remove oauth flow type

J0:
Replacement for #654 as the commit history there is a little messy

Corresponding server PR: https://github.com/supabase/gotrue/pull/1027

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -392,7 +392,6 @@ export default class GoTrueClient {
       scopes: credentials.options?.scopes,
       queryParams: credentials.options?.queryParams,
       skipBrowserRedirect: credentials.options?.skipBrowserRedirect,
-      flowType: this.flowType ?? 'implicit',
     })
   }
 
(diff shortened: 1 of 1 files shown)


src/GoTrueClient.ts
```diff
@@ -1432,12 +1427,11 @@ export default class GoTrueClient {
     if (options?.scopes) {
       urlParams.push(`scopes=${encodeURIComponent(options.scopes)}`)
     }
-    if (options?.flowType === 'pkce') {
+    if (this.flowType === 'pkce') {
       const codeVerifier = generatePKCEVerifier()
       await setItemAsync(this.storage, `${this.storageKey}-code-verifier`, codeVerifier)
       const codeChallenge = await generatePKCEChallenge(codeVerifier)
       const flowParams = new URLSearchParams({
-        flow_type: `${encodeURIComponent(options.flowType)}`,
```
thorwebdev, line 1440: Do we not still need to send the flowtype to the api?

thorwebdev (approved):
LGTM, just one question there regarding flowtype in the url params.

src/GoTrueClient.ts
kangmingtay: yeah, it's no longer necessary since https://github.com/supabase/gotrue/pull/1027

PR #663, opened 2023-04-17
fix: default to plain code challenge method if crypto API is undefined

kangmingtay:
* This PR improves support for using PKCE in environments where the crypto API is undefined (e.g. Node18 in NextJS server-side) 
* If the crypto API is undefined, gotrue-js will default to using `plain` instead `sha256` for the code challenge method

### DIFF
--- a/src/lib/helpers.ts
+++ b/src/lib/helpers.ts
@@ -245,9 +245,13 @@ export function generatePKCEVerifier() {
   const verifierLength = 56
   const array = new Uint32Array(verifierLength)
   if (typeof crypto === 'undefined') {
-    throw new Error(
-      'PKCE is not supported on devices without WebCrypto API support, please add polyfills'
-    )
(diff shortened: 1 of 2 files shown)


src/lib/helpers.ts
```diff
@@ -245,9 +245,13 @@ export function generatePKCEVerifier() {
   const verifierLength = 56
   const array = new Uint32Array(verifierLength)
   if (typeof crypto === 'undefined') {
-    throw new Error(
-      'PKCE is not supported on devices without WebCrypto API support, please add polyfills'
-    )
+    const charSet = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789'
```
J0, line 248: /nit Consider adding the [four punctuation characters](https://www.oauth.com/oauth2-servers/pkce/authorization-request/)
```suggestion
    const charSet = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-._~'
```

J0 (approved):
LGTM, thanks for putting this together

PR #672, opened 2023-05-04
feat: remove code param

J0:

We remove the `code` parameter  after exchanging  the code for session if user has `detectSessionInURL` enabled

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -888,7 +888,9 @@ export default class GoTrueClient {
         const { data, error } = await this.exchangeCodeForSession(authCode)
         if (error) throw error
         if (!data.session) throw new AuthPKCEGrantCodeExchangeError('No session detected.')
-        window.location.search = ''
+        let url = new URL(window.location.href);
+        url.searchParams.delete('code')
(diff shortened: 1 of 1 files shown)


src/GoTrueClient.ts
```diff
@@ -888,6 +888,7 @@ export default class GoTrueClient {
         const { data, error } = await this.exchangeCodeForSession(authCode)
         if (error) throw error
         if (!data.session) throw new AuthPKCEGrantCodeExchangeError('No session detected.')
+        window.location.search = ''
```
thorwebdev, line 891: I think we might want to use https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/delete instead to make sure we only remove the code param?

thorwebdev: @J0 we also should make sure to not cause a page refresh, do based on some research, we might need something like this:

```js
let url = new URL(window.location.href);
url.searchParams.delete('code')
window.history.replaceState(window.history.state, document.title, url.toString())
```

thorwebdev (changes requested):
I think we need to be a bit more careful when dealing with url params.

thorwebdev (approved):
LGTM

PR #673, opened 2023-05-07
feat: add support for sign in using refresh token

mgitre:

Adds support for signing in with refreshToken, like existed in supabase-js v1

A user can't sign in with a refreshToken, which can make OAuth simpler on react-native

## What is the new behavior?

A user can sign in using a refreshToken by calling signInWithRefreshToken


j4w8n:
Here is a brief explanation of why they haven't gone back to this behavior:

https://github.com/supabase/gotrue-js/pull/473#issuecomment-1274914088

PR #681, opened 2023-05-16
fix: only refresh the session if autoRefreshToken is enabled

### LINKED ISSUE #677: access-token always refresh regardless `autoRefreshToken` setting when token expired
# Bug report

- [X] I confirm this is a bug with Supabase, not with my own application. 
- [X] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

when token is expired and `autoRefreshToken` set to false, it first remove the session in `_recoverAndRefresh`, then it will `_callRefreshToken` in `getSession`. which implicit refresh the access-token.

https://github.com/supabase/gotrue-js/blob/master/src/GoTrueClient.ts#L714-L721

## To Reproduce
1. set jwt_expiry to 10
2. wait jwt expire
3. invoke `getSession`

## Expected behavior

respect autoRefreshToken

## System information

- OS: [e.g. macOS, Windows]
- Browser (if applies) [e.g. chrome, safari]
- Version of supabase-js: [e.g. 6.0.2]
- Version of Node.js: [e.g. 10.10.0]

kangmingtay:
* Partially fixes #677 
* `getSession` should only refresh the session if `autoRefreshToken` is enabled
* Related: Just merged in https://github.com/supabase/gotrue-js/pull/598 which should fix the issue of `_recoverAndRefresh` removing the session from local storage if it's expired, thus preventing the developer from handling the session refresh on their own 

Note: This seems like a possible breaking change for all those who rely on `getSession` to always return a valid session even if the token has expired and regardless of whether `autoRefreshToken` is enabled. Should we return the expired session if `autoRefreshToken` is `false`? 

Edit: I think it will still be useful to return the expired session if `autoRefreshToken` is `false` so that the developer can use the refresh token to refresh the session with `refreshSession()`. I'm also thinking of adding a `isExpired` field to the `Session` object so that it's easier for the developer to check if the session is expired.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -715,12 +715,15 @@ export default class GoTrueClient {
       return { data: { session: currentSession }, error: null }
     }
 
-    const { session, error } = await this._callRefreshToken(currentSession.refresh_token)
-    if (error) {
-      return { data: { session: null }, error }
+    if (this.autoRefreshToken && currentSession.refresh_token) {
+      const { session, error } = await this._callRefreshToken(currentSession.refresh_token)
+      if (error) {
+        return { data: { session: null }, error }
+      }
+      return { data: { session }, error: null }
     }
 
-    return { data: { session }, error: null }
+    return { data: { session: null }, error: null }
   }
 
   /**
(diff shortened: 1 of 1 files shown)


kangmingtay:
closing this PR for now since we decided that `getSession()` should stick to the current behaviour to always return a valid session or no session

Issue #670, opened 2023-04-26
Social Login (OAuth) User Metadata, Invite Shortcomings


shellscape:
# Bug report

I'm opening this as a bug because I view it as a rather large oversight, rather than a feature request.

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Both the social login system and the invite system seem to have a pretty big hole. The invite system can't use the social login for confirmation, and the social login system doesn't appear to provide any mechanism to provide custom user metadata for insertion into raw_user_meta_data when the auth.user record is created.

The latter need exists because the invites system can't use a social login, so we need some method of conveying metadata - like an invite code we hand roll - into the social login process for new users. We've considered multiple redirect, but that's very sloppy and doesn't play well with webhooks or triggers on `auth.users`.

The documentation site is just absolutely woeful, but the types for the package here outline what's possible and what isn't with regard to user metadata during signup processes.

Nearly all of the `signIn`/`signUp` methods include the `data` option, as can be seen in the types: https://github.com/supabase/gotrue-js/blob/eb96536a441d21fae75d25ab6de346808c6e6044/src/lib/types.ts#L389-L395

Even the Invite and Magic Link methods include the `data` options: https://github.com/supabase/gotrue-js/blob/eb96536a441d21fae75d25ab6de346808c6e6044/src/lib/types.ts#L539-L544

Except of course, for `signInWithOAuth`: https://github.com/supabase/gotrue-js/blob/eb96536a441d21fae75d25ab6de346808c6e6044/src/lib/types.ts#L440-L453

Now, granted that allowing the invite system to utilize social/oauth logins would probably be a massive lift, we could in theory get around that by providing metadata to a `data` option to `signInWithOAuth` which could contain an invite code with instructions for any webhooks or triggers on the `auth.users` table.

What about `redirectTo` then? Well, that's certainly less than optimal. If we were able to use `data` the workflow would look like this:

- call `signInWithOAuth` with `data: { inviteCode }`
- social login process completes
- `auth.users` is updated with new user record
- a trigger is executed and `public.profiles` is updated nearly instantaneously with data from the `inviteCode`

Using `redirectTo` the process takes a completely different shape, without the same speed or fidelity:

[...]

PR #682, opened 2023-05-16
fix: `signInWithOAuth` should accept an invite token

kangmingtay:
* Partially address #670 
* Improves the DX for accepting an invite through `signInWithOAuth` as mentioned in [this comment](https://github.com/supabase/gotrue-js/issues/670#issuecomment-1548968808)


J0 (approved):

kangmingtay:
closing for now because this can be done via the query params option rather than having to add another option to the type

gary-lo:
Can you provide an example of how query params can be used for this scenario? @kangmingtay

PR #684, opened 2023-05-18
feat: don't emit `SIGNED_IN` events on visibility change

### LINKED ISSUE #579: onAuthStateChange fires every time we focus the tab

hf:
Each time a tab became visible, the `SIGNED_IN` event was emitted. Although this is normal, it's unnecessary overhead. In some implementations, it could result in unnecessary requests or UI re-renders.

Fixes #579

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -241,7 +241,7 @@ export default class GoTrueClient {
       }
 
       // no login attempt via callback url try to recover session from storage
-      await this._recoverAndRefresh()
+      await this._recoverAndRefresh(true)
       return { error: null }
     } catch (error) {
       if (isAuthError(error)) {
@@ -660,10 +660,6 @@ export default class GoTrueClient {
     }
   }
 
-  /**
-   * Returns the session, refreshing it if necessary.
-   * The session returned can be null if the session is not detected which can happen in the event a user is not signed-in or has logged out.
-   */
   async getSession(): Promise<
     | {
         data: {
@@ -688,6 +684,32 @@ export default class GoTrueClient {
     // save to just await, as long we make sure _initialize() never throws
     await this.initializePromise
 
+    return await this._getSession()
+  }
+
+  /**
+   * Private version of {@link #getSession()} safe to reuse in private methods.
+   */
+  private async _getSession(): Promise<
+    | {
+        data: {
+          session: Session
+        }
+        error: null
+      }
+    | {
+        data: {
+          session: null
+        }
+        error: AuthError
+      }
+    | {
+        data: {
+          session: null
+        }
+        error: null
+      }
+  > {
     let currentSession: Session | null = null
 
     if (this.persistSession) {
@@ -709,7 +731,7 @@ export default class GoTrueClient {
     }
 
     const hasExpired = currentSession.expires_at
-      ? currentSession.expires_at <= Date.now() / 1000
+      ? currentSession.expires_at - EXPIRY_MARGIN <= Date.now() / 1000
       : false
     if (!hasExpired) {
       return { data: { session: currentSession }, error: null }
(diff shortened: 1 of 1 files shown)


src/GoTrueClient.ts
```diff
@@ -1194,32 +1216,17 @@ export default class GoTrueClient {
...
-
-          if (error) {
-            console.log(error.message)
-            await this._removeSession()
-          }
-        }
-      } else {
-        if (this.persistSession) {
-          await this._saveSession(currentSession)
-        }
```
hf, line 1222: No need to have 2 different versions that do refreshing when `getSession` does it all. It uses `_getSession` because `_recoverAndRefresh` is part of the `_initialize` flow and it means that `getSession` would never complete.

src/GoTrueClient.ts
```diff
@@ -1194,32 +1216,17 @@ export default class GoTrueClient {
    * Recovers the session from LocalStorage and refreshes
    * Note: this method is async to accommodate for AsyncStorage e.g. in React native.
    */
-  private async _recoverAndRefresh() {
+  private async _recoverAndRefresh(isInitial: boolean) {
```
kangmingtay, line 1219: what does the `isInitial` param do here? can we document it above?

src/GoTrueClient.ts
```diff
@@ -709,7 +731,7 @@ export default class GoTrueClient {
     }
 
     const hasExpired = currentSession.expires_at
-      ? currentSession.expires_at <= Date.now() / 1000
+      ? currentSession.expires_at - EXPIRY_MARGIN <= Date.now() / 1000
```
kangmingtay, line 734: hmm i rmb we didn't do this before because there were some other places where we used `getSession` and checked the expiry margin too which would result in double counting - but just need to check again to be sure

kangmingtay (approved):

src/GoTrueClient.ts
```diff
@@ -660,10 +660,6 @@ export default class GoTrueClient {
     }
   }
 
-  /**
-   * Returns the session, refreshing it if necessary.
```
J0, line 664: Ah is this section no longer needed/are we making `getSession` internal? Not sure if removing the comments will remove the description from reference spec

src/GoTrueClient.ts
J0: Ah okay, so to expand -  if we called `getSession` instead of `_getSession()`  this would fail as `getSession` would call [intializePromise](https://github.com/supabase/gotrue-js/pull/684/files#diff-3522461172efd6058d6b8da62fc2d30d8b524d2b64894ea2c67218c52f7fdff5R685) which would  call `_initialize()` which would then call `_recoverAndRefresh` leading to an infinite loop?

J0 (approved):

src/GoTrueClient.ts
hf: You're right there was an issue... I'll actually add the `Date.now()` in the param so when called from `_recoverAndRefresh` it will use the current time + margin, and otherwise just the current time.

src/GoTrueClient.ts
hf: Hmm I'll investigate.

hf:
Found something else: in multi-tab this SIGNED_IN event is emitted to all other tabs so it may be a bit too much.

robinjhuang:
It would be fantastic to see this PR merged soon. I'm encountering the same issue in a React application. In React, it's a common practice to store the session object in the component state. Because of that, each time a SIGNED_IN event fires and updates the session, it triggers a re-render of the entire application, impacting performance.

As a workaround, I'm currently comparing the token field of the incoming and existing session objects before updating the state. While this approach mitigates the issue, it feels more like a patch than a proper solution. Addressing this behavior directly within the Supabase library would be a much cleaner and more efficient way to handle it.

J0:
Quick check: is this PR still relevant?

jgabriel98:
> Quick check: is this PR still relevant?

@J0  Yeah, i'm facing this issue right now.

mandarini:
Closing after asking @hf !

PR #685, opened 2023-05-18
feat: add async/await support for `onAuthStateChange` callbacks

### LINKED ISSUE #276: Possibility to sync onAuthStateChange with signIn/Out()
# Feature request

The code is from a nuxt3 app using the @nuxtjs/supabase module.

The onAuthStateChange listener is often used to the set the user on the client side and its a good and central place to do it imo.

But when having a sign in action like this:

```js
const signIn = async() => {
  const { session, error } = await client.auth.signIn({ email: email.value, password: password.value })
  navigateTo('home')
}
```

with a route guard like this

```js
export default defineNuxtRouteMiddleware((to) => {
  const user = useSupabaseUser()
  if (!(user.value))
    return navigateTo({ name: 'login', query: { redirect: to.path } })
})

```

and a onAuthStateChange listener like this

```js
 // Once Nuxt app is mounted
  nuxtApp.hooks.hook('app:mounted', () => {
    // Listen to Supabase auth changes
    client.auth.onAuthStateChange(async (event: AuthChangeEvent, session: Session | null) => {
      await setServerSession(event, session)
      user.value = client.auth.user()
    })
  })

```

The user is still not available when navigating to the 'home' after login. Causing the middleware to redirect the user back to login.

A possibility in the options to wait for the changeListeners to finish before the login/logout promise resolves.
This would clean up client side code and sync the entire auth process. This could default to false, if it introduces
unwanted behaviour for some cases.

Something like this maybe:

```js
const { session, error } = await client.auth.signIn({ email: email.value, password: password.value },{waitForListeners: true})
```

hf:
If a callback passed to `onAuthStateChange` is `async`, the emitter of the event (like `signInXYZ` methods) would wait before returning. All such callbacks will be waited upon to complete, but errors will be reported once all have completed. All such errors are logged, but only the first one is thrown.

Fixes #276

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -181,8 +181,8 @@ export default class GoTrueClient {
         )
       }
 
-      this.broadcastChannel?.addEventListener('message', (event) => {
-        this._notifyAllSubscribers(event.data.event, event.data.session, false) // broadcast = false so we don't get an endless loop of messages
+      this.broadcastChannel?.addEventListener('message', async (event) => {
+        await this._notifyAllSubscribers(event.data.event, event.data.session, false) // broadcast = false so we don't get an endless loop of messages
       })
     }
 
@@ -229,11 +229,11 @@ export default class GoTrueClient {
 
         await this._saveSession(session)
 
-        setTimeout(() => {
+        setTimeout(async () => {
           if (redirectType === 'recovery') {
-            this._notifyAllSubscribers('PASSWORD_RECOVERY', session)
+            await this._notifyAllSubscribers('PASSWORD_RECOVERY', session)
           } else {
-            this._notifyAllSubscribers('SIGNED_IN', session)
+            await this._notifyAllSubscribers('SIGNED_IN', session)
           }
         }, 0)
 
@@ -323,7 +323,7 @@ export default class GoTrueClient {
 
       if (data.session) {
         await this._saveSession(data.session)
-        this._notifyAllSubscribers('SIGNED_IN', session)
+        await this._notifyAllSubscribers('SIGNED_IN', session)
       }
 
       return { data: { user, session }, error: null }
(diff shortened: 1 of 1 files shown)


src/GoTrueClient.ts
```diff
@@ -1268,12 +1270,33 @@ export default class GoTrueClient {
...
     }
 
-    this.stateChangeEmitters.forEach((x) => x.callback(event, session))
+    const errors: any[] = []
+    const promises = Array.from(this.stateChangeEmitters.values()).map(async (x) => {
+      try {
+        await x.callback(event, session)
+      } catch (e: any) {
+        errors.push(e)
+      }
```
hf, line 1288: I guess the logging can be done here.

src/GoTrueClient.ts
```diff
@@ -1069,9 +1071,9 @@ export default class GoTrueClient {
       } = await this.getSession()
       if (error) throw error
 
-      this.stateChangeEmitters.get(id)?.callback('INITIAL_SESSION', session)
+      await this.stateChangeEmitters.get(id)?.callback('INITIAL_SESSION', session)
     } catch (err) {
-      this.stateChangeEmitters.get(id)?.callback('INITIAL_SESSION', null)
+      await this.stateChangeEmitters.get(id)?.callback('INITIAL_SESSION', null)
```
hf, line 1076: I don't think these should be `await`-ed.

J0: bumping this - am keen to know why these statements shouldn't be `await`-ed as well

J0 (approved):

src/GoTrueClient.ts
hf: Trying to remember. 😂

hf: Ah! Nevermind. I thought there was the possibility of recursive await, but doesn't appear so.

PR #687, opened 2023-05-20
feat: update tsdoc

J0:

Companion PR to #13820

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -487,7 +487,7 @@ export default class GoTrueClient {
    * if you are using phone sign in with the 'whatsapp' channel. The whatsapp
    * channel is not supported on other providers
    * at this time.
-   * This method supports PKCE on the magic link flow.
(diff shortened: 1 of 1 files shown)


src/GoTrueClient.ts
```diff
@@ -485,6 +487,7 @@ export default class GoTrueClient {
    * if you are using phone sign in with the 'whatsapp' channel. The whatsapp
    * channel is not supported on other providers
    * at this time.
+   * This method supports PKCE on the magic link flow.
```
J0, line 490: Wasn't sure if users would understand what "the magic link flow" refers to but couldn't think of a better phrasing offhand. Open to suggestions if there are any.

kangmingtay: maybe this?

```suggestion
   * This method supports PKCE when an email is passed.
```

J0: Looks great, thanks!

kangmingtay (approved):

PR #688, opened 2023-05-22
fix: add reauthenticate method

kangmingtay:
* Supercedes #258
* Adds `reauthenticate()` method to allow a user to request for a nonce to subsequently update their password
* Updates the type for `UserAttributes` which is used in `updateUser()` to include the `nonce` 
* If a nonce is not passed into `updateUser()`, an `AuthSessionMissingError()` will be returned

## Details
With this addition, the update password flow will look like the following when "Secure password change" is enabled on the Supabase dashboard. For users that are self-hosting gotrue and using gotrue-js, this requires setting the environmentr variable, `GOTRUE_SECURITY_UPDATE_PASSWORD_REQUIRE_REAUTHENTICATION="true"`
```

// Send reauthentication to the current user. User must be signed-in to perform this action.
await supabase.auth.reauthenticate()

// Update password 
await supabase.auth.updateUser({ password: "new-pasword", "nonce": "123456" }
```

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -620,6 +620,32 @@ export default class GoTrueClient {
   }
 
   /**
+   * Sends a reauthentication OTP to the user's email or phone number.
+   * Requires the user to be signed-in.
+   */
+  async reauthenticate(): Promise<AuthResponse> {
+    try {
+      const {
+        data: { session },
+        error: sessionError,
+      } = await this.getSession()
+      if (sessionError) throw sessionError
+      if (!session) throw new AuthSessionMissingError()
+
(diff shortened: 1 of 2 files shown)


j4w8n:
How does a dev get the nonce?

j4w8n:
@kangmingtay I don't know how to read Go code very well, but that's what I was thinking. So I still don't understand how that code gets from a user's email to a dev's code. Maybe a button click on the site would trigger sending the code, then you show a change password form with a "code" field?

hf (approved):

kangmingtay:
@j4w8n ah i see, yeah if the user wants to update their password, you would have a button that calls the reauthenticate method to send an OTP to the user's email. The update password form field should provide a field to enter the OTP sent to the user. Then, the user can enter the password and the OTP to successfully update their password.

PR #695, opened 2023-05-30
fix: substitute PKCE error for client error message

J0:

Currently, `exchangeCodeForSession` returns the error returned by the API to `gotrue-js`. This can be potentially confusing as the error message might reference a `code_verifier` which could be confusing as `code_verifier` management is currently handled by `supabase-js` and transparent to the developer. 

Once finished, this PR should provide a way for users to determine what error has occurred and some form of remedy action. Might tie in to the work being done for  error codes/error messages.

## Additional context

Attempts to address [this Github issue](https://github.com/supabase/auth-helpers/issues/545)

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -7,6 +7,7 @@ import {
   AuthInvalidCredentialsError,
   AuthRetryableFetchError,
   AuthSessionMissingError,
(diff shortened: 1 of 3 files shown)


J0:
Re-open a fresh PR for this

PR #699, opened 2023-06-03
fix: added package name prefix to all `console.` methods in `src/`

madebyfabian:

After upgrading @supabase/supabase-js, I stumbled on an error message (`No storage option exists to persist the session, which may result in unexpected behavior when using auth.
        If you want to set persistSession to true, please provide a storage option or you may set persistSession to false to disable this warning.`).

It took me a while to figure out from which package this comes and why. Found it only because it was reported in the nuxt/supabase repo https://github.com/nuxt-modules/supabase/issues/188.

I therefore suggest to prefix console.error / console.warning with this package's name, so you can easily check where this random error in your console comes from.

Errors are logged into the server console without me having an idea from which package they come from.

## What is the new behavior?

Now every error/warning has the package name `[@supabase/gotrue-js]` prefixed.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -174,7 +174,7 @@ export default class GoTrueClient {
 
     if (this.persistSession && this.storage === localStorageAdapter && !supportsLocalStorage()) {
       console.warn(
-        `No storage option exists to persist the session, which may result in unexpected behavior when using auth.
+        `[@supabase/gotrue-js] No storage option exists to persist the session, which may result in unexpected behavior when using auth.
         If you want to set persistSession to true, please provide a storage option or you may set persistSession to false to disable this warning.`
       )
     }
@@ -184,7 +184,7 @@ export default class GoTrueClient {
         this.broadcastChannel = new globalThis.BroadcastChannel(this.storageKey)
       } catch (e: any) {
         console.error(
-          'Failed to create a new BroadcastChannel, multi-tab state changes will not be available',
+          '[@supabase/gotrue-js] Failed to create a new BroadcastChannel, multi-tab state changes will not be available',
           e
         )
       }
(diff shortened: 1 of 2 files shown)


sedhha (approved):
LGTM

notjustinshaw:
Can you point me in the direction of a fix for this (or at least some docs for the `persistSession` option mentioned). I seem to have this "unexpected behavior" :)

AcquireAgencyInc:
Would love some info on why this warning is happening and how to get rid of it. It is clogging up our supabase logs rendering them useless.

AdamEisfeld:
Would love some info on why this warning is happening and how to get rid of it, it is clogging up our supabase logs rendering them useless.

madebyfabian:
@AdamEisfeld @notjustinshaw The error appears because some config is not set. Ideally, this would not throw an error, but here we are. The solution in nuxt seems to just set this config value to true/false. But I haven't done any deepdive so it could be something else. https://github.com/nuxt-modules/supabase/issues/188 have you tried that?

nfts2me:
We are also having this issue.
I guess there is no solution yet, right?

j4w8n:
@nfts2me @notjustinshaw @AdamEisfeld this behavior is designed to help devs be aware that they've got a configuration which could cause them issues. There are plenty of documented cases [here](https://github.com/supabase/gotrue-js/issues/539)

Most of the time, you pass the following option to your supabase client - usually just the server-side ones - the error will go away.

```js
const supabase = await createClient(url, key, {
  auth: {
    persistSession: false
  }
}
```

notjustinshaw:
Let’s fix this. Seems easy enough to make persist-state:false the default?

madebyfabian:
Hi there, is there still interest in improving the log messages (which this PR was originally intended to be)? Or can I close this and it's fine due to the core issue being fixed.

AlexIsMaking:
> Hi there, is there still interest in improving the log messages (which this PR was originally intended to be)? Or can I close this and it's fine due to the core issue being fixed.

IMO this change is still worthwhile, it will make it easier to spot the source of the error before making the changes which are mentioned in the thread to avoid it.

Issue #141, opened 2021-09-18
Network error removes session data


dshukertjr:
# Bug report

## Describe the bug

By looking at the code [here](https://github.com/supabase/gotrue-js/blob/master/src/GoTrueClient.ts#L589), it seems like whenever user opens the app while offline, the session is deleted. 
```typescript
const { error } = await this._callRefreshToken(currentSession.refresh_token)
if (error) {
  console.log(error.message)
  await this._removeSession()
}
```

## To Reproduce

1. Create an app using Supabase
2. Sign in with your favorite method
3. Turn on air plane mode
4. Open the app and your are signed out 

## Expected behavior

Session should not be deleted when offline, and refreshing session should be attempted upon device regaining internet connection. 

## Additional context

I guess this issue has two challenges that it needs to overcome:
1. detect network error and in such case, do not remove session (I guess this is pretty easy)
2.  in case the error was a network error, detect when the device becomes online again and retry session recovery

dshukertjr:
I was thinking if removing session only if status code is present, and if not, retry session recovery [here](https://github.com/supabase/gotrue-js/blob/master/src/GoTrueClient.ts#L587).

```typescript
if (error) {
  const status = error?.status
  console.log(error.message)
  if(status) {
    await this._removeSession()
  } else {
    // retry session recovery here
  }
}
```

I feel like this might work, but at the same time, I haven't really thought it though, so I think there are some edge cases where this might be a problem.

ferrannp:
This issue is quite old but this still happening in latest version of `supabase-js v2`. Basically, if you open your app with the Internet off, your session will be killed. This seems extremely critical.

nahtnam:
I've been having this same issue as well! My temporary solution was to extend the session tokens out to be much longer.

If the `initialize` fails the user is immediately logged out: https://github.com/supabase/gotrue-js/blob/eb96536a441d21fae75d25ab6de346808c6e6044/src/GoTrueClient.ts#L1166-L1173 (as you pointed out in the issue). Unfortunately, we don't have the ability to control the `initialize` function as it's immediately run on `createClient`. Even the `autoRefreshToken` prop won't help.

On the other hand, the auto refresher has a nice exponential backoff logic: https://github.com/supabase/gotrue-js/blob/eb96536a441d21fae75d25ab6de346808c6e6044/src/GoTrueClient.ts#L1089

[...]

PR #710, opened 2023-06-26
feat: `_recoverAndRefresh` does not remove session on retryable error

hf:
If the `_callRefreshToken` method failed with a network error, the session would have been removed, which should generally not be the case as the device is offline.


J0 (approved):

kangmingtay:
attempts to fix #141

PR #712, opened 2023-06-27
feat: add debug messages and configuration

hf:
Adds the ability to configure debug messages for the library, which are useful when inspecting what the GoTrueClient is doing in certain circumstances.


src/GoTrueClient.ts
```diff
@@ -1319,6 +1422,8 @@ export default class GoTrueClient {
 
       return result
     } catch (error) {
+      this._debug(debugName, 'erorr', error)
```
J0, line 1425:
```suggestion
      this._debug(debugName, 'error', error)
```

J0 (approved):

PR #713, opened 2023-06-28
feat: add `signOut()` behavior option

hf:
Adds the `behavior` option to the `signOut` method which can be used to control which sessions should be signed out of the account.

See:
- https://github.com/supabase/gotrue/pull/1112


j4w8n:
Excellent.

Was there any discussion about the `behavior` verbiage? I ask because it seems like `scope` would be a good fit here.

J0 (approved):
LGTM, may want to consider the naming suggestion - another minor reason but no to prefer `scope` over behaviour is that there's only one spelling of it between UK and US English. In either case, the functionality seems fine though so will leave the naming up to you

kangmingtay:
lets go with `scope`

PR #714, opened 2023-06-29
fix: `_recoverAndRefresh` efficiency and tidy up

j4w8n:

Removes unneeded checks and other code.

Within `_recoverAndRefresh()`, we're checking for `currentSession.expires_at` as part of a "hey is the session expired" `if` block.

We've also left a `_notifyAllSubscribers()` call for `SIGNED_IN`, after a recent [PR](https://github.com/supabase/gotrue-js/pull/711).

## What is the new behavior?

`expires_at` is already checked for existence within `_isValidSession()`, so it's been removed from the "expired session" check.

We can also collapse a child `if` block and move it's check for `this.autoRefreshToken` to the parent.

We also remove the `SIGNED_IN` emit event that happens towards the end of this function. In all other cases, `_notifyAllSubscribers()` is only called if a `_saveSession()` or `_removeSession()` call preceded it. The one exception is within `constructor()`, where an event listener, to emit events, is setup for multi-tab communication. The recent PR, referenced above, removed the `_saveSession()` call within `_recoverAndRefresh()`, so I'd argue there's no need to emit `SIGNED_IN`.

A possible anti-argument might be something like, "Emitting this event is just expressing that we noticed someone is signed in during recover and refresh". However, in the case of an expired session, we just refresh the token - causing a `TOKEN_REFRESHED` event - and then move on. There's no `SIGNED_IN` event emitted in this case. I'd say we either emit `SIGNED_IN` in both cases or not at all, but I lean towards not at all.

## Additional context


j4w8n:
Hmm... not sure why I have residual commits here. Closing PR and will figure that out.

PR #715, opened 2023-06-29
fix: `_recoverAndRefresh` code efficiency and tidy up

j4w8n:

Removes unneeded checks, other code and makes things slightly more efficient.

Within `_recoverAndRefresh()`, we're checking for `currentSession.expires_at` and `currentSession.refreshToken` as part of a "hey is the session expired?" `if` block.

We've also left a `_notifyAllSubscribers()` call for `SIGNED_IN`, after a recent [PR](https://github.com/supabase/gotrue-js/pull/711).

## What is the new behavior?

`currentSession.expires_at` and `currentSession.refresh_token` are already checked for existence within `_isValidSession()`, so they've been removed from their respective `if` blocks. I can also understand if you'd still rather be explicit, even though that means they're getting checked twice.

We can also collapse a child `if` block and move it's remaining check for `this.autoRefreshToken` to the parent.

We also remove the `SIGNED_IN` emit event that happens towards the end of this function. In all other cases, `_notifyAllSubscribers()` is only called if a `_saveSession()` or `_removeSession()` call preceded it. The one exception is within `constructor()`, where an event listener, to emit events, is setup for multi-tab communication. The recent PR, referenced above, removed the `_saveSession()` call within `_recoverAndRefresh()`, so I'd argue there's no need to emit `SIGNED_IN`.

A possible anti-argument might be something like, "Emitting the `SIGNED_IN` event is just expressing that we noticed someone is signed in during recover and refresh". However, in the case of an expired session, we just refresh the token - causing a `TOKEN_REFRESHED` event - and then move on. There's no `SIGNED_IN` event emitted in this case. I'd say we either emit `SIGNED_IN` in both cases or not at all, but I lean towards not at all.

## Additional context

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -1352,35 +1352,28 @@ export default class GoTrueClient {
       }
 
       const timeNow = Math.round(Date.now() / 1000)
-      const expiresWithMargin = (currentSession.expires_at ?? Infinity) < timeNow + EXPIRY_MARGIN
+      const expiresWithMargin = currentSession.expires_at! < timeNow + EXPIRY_MARGIN
 
       this._debug(
         debugName,
         `session has${expiresWithMargin ? '' : ' not'} expired with margin of ${EXPIRY_MARGIN}s`
       )
 
-      if (expiresWithMargin) {
-        if (this.autoRefreshToken && currentSession.refresh_token) {
-          const { error } = await this._callRefreshToken(currentSession.refresh_token)
+      if (this.autoRefreshToken && expiresWithMargin) {
+        const { error } = await this._callRefreshToken(currentSession.refresh_token)
 
-          if (error) {
-            console.error(error)
+        if (error) {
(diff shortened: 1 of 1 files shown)


hf:
I'm afraid we can't touch the events too much as they are an API contract. The event must remain, however inconvenient or inefficient.

As for the rest of the changes, we'd like to keep the changes in these functions for now within the team at Supabase as we're debugging some weird very rare behaviors with random logouts. I'll take a look at your changes and try to incorporate them in fixes as we go along. Sorry I don't want to discourage you, but this is one of those heisenbugs which we need consistency to figure out where it's coming from.

j4w8n:
Understood. Happy bug-hunting!

> I'm afraid we can't touch the events too much as they are an API contract. The event must remain, however inconvenient or inefficient.

I don't understand this part, however, but that's ok.

j4w8n:
I'm closing this, as it's very old and isn't strictly needed.

PR #717, opened 2023-06-30
fix: don't remove session in resend

Vinzent03:

Bug fix

When you call the `resend` method to e.g. resend an email to change your email, you get logged out.

## What is the new behavior?

On email and phone change you don't get logged out.

## Additional context

I noticed this when implementing this method in gotrue-dart https://github.com/supabase/supabase-flutter/pull/517#discussion_r1246427921

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -723,7 +723,11 @@ export default class GoTrueClient {
    */
   async resend(credentials: ResendParams): Promise<AuthOtpResponse> {
     try {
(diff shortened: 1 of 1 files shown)


dshukertjr:
@kangmingtay would you be able to review this one? @Vinzent03 noticed this, and I think it's a bug, but wanted to be sure.

kangmingtay (approved):
@Vinzent03 thanks for catching this one!

PR #724, opened 2023-07-06
fix: add emailRedirectTo option to resend method

kangmingtay:
* Adds the `emailRedirectTo` option to resend method
* Resolves https://github.com/supabase/gotrue/issues/1142


J0 (approved):
LGTM, nice work!

PR #726, opened 2023-07-09
feat: refactor to `_useSession` semantics

hf:
`getSession()` is problematic. The issue comes because whenever the library itself does anything with the session, it may be very quick < 10ms or very slow >= 10s. For example, if a session needs to be refreshed, when you call `getSession()` the response may be very quick or take very long. 

When such long processing is required, a race condition arises between multiple windows or tabs concurrently doing the same processing. In reality, only one such tab / window (in further text "process") needs to do the processing. Therefore, just using a concept like a method call to `getSession()` is not advisable.

With this change, the internals of the library use a new function called `_useSession()` which expects a callback to be provided. So long as the callback's promise has not resolved, the session is "being used." In follow up PRs, locking semantics will be used to ensure that only one `_useSession` callback is active amongst all processes.

The existing `getSession()` computation is put behind `__loadSession` which should not be used outside of `_useSession()`. A debug-mode stack check is performed to make sure that issues like that are caught.

This PR also adds two new utility functions `stackGuard` and `isInStackGuard`. These can be used to detect recursive calls or inappropriate use of methods. For example, in this PR a stack guard is used at debug time to make sure that `__loadSession` is only called within `_useSession`, as that is the only valid use of the method. In follow up versions this will become a runtime exception.

It works because when using async/await JS engines like browsers and Node will respect stack traces. A stack guard is a special string that shows up in the stack trace of the `Error` object, which if seen (by `isInStackGuard`) can be used to detect whether a function was called from the correct parent-caller or whether there is a recursive call.

It's not possible to use plain functions because of JS minification, which mangles the name of virtually all functions, so a trick has to be used to dynamically generate a function name which is the special stack guard string.


src/GoTrueClient.ts
```diff
@@ -194,6 +195,8 @@ export default class GoTrueClient {
       getAuthenticatorAssuranceLevel: this._getAuthenticatorAssuranceLevel.bind(this),
     }
 
+    this.useSessionStackRegExp = new RegExp(`^\s*at\s+${this._useSession.name}\\(`)
```
hf, line 198: Add comments about minification here.

src/GoTrueClient.ts
```diff
@@ -785,11 +852,15 @@ export default class GoTrueClient {
...
+    if (this.logDebugMessages && !this._isInUseSession()) {
+      throw new Error('Please use #_useSession()')
+    }
+
     // make sure we've read the session from the url if there is one
     // save to just await, as long we make sure _initialize() never throws
     await this.initializePromise
 
-    this._debug('#getSession()', 'begin')
+    this._debug('#_loadSession()', 'begin')
```
hf, line 863: Double underscore.

src/GoTrueClient.ts
```diff
@@ -821,7 +892,7 @@ export default class GoTrueClient {
         : false
 
       this._debug(
-        '#getSession()',
+        '#_loadSession()',
```
hf, line 895: Double underscore.

src/GoTrueClient.ts
```diff
@@ -838,7 +909,7 @@ export default class GoTrueClient {
 
       return { data: { session }, error: null }
     } finally {
-      this._debug('#getSession()', 'end')
+      this._debug('#_loadSession()', 'end')
```
hf, line 912: Double underscore.

src/GoTrueClient.ts
hf:
```suggestion
    this.useSessionStackRegExp = new RegExp(`^\\s*at\\s+${this._useSession.name}\\(`)
```

Actually.

J0 (approved):

src/lib/helpers.ts
```diff
@@ -282,3 +282,112 @@ export async function generatePKCEChallenge(verifier: string) {
...
   return base64urlencode(hashed)
 }
+
+const STACK_GUARD_PREFIX = `__stack_guard__`
+const STACK_GUARD_SUFFIX = `__`
+
+// Firefox and WebKit based browsers encode the stack entry differently, but
+// they all include the function name. So instead of trying to parse the entry,
+// we're only looking for the special string `__stack_guard__${guardName}`. The
+// guard name is terminated when seeing '@', ' ' or '('.
```
hf, line 292: Rewrite this comment.

src/lib/helpers.ts
```diff
@@ -282,3 +282,112 @@ export async function generatePKCEChallenge(verifier: string) {
...
+export function isInStackGuard(name: string): boolean {
+  STACK_GUARD_CHECK_FN()
+
+  let error: Error
+
+  try {
+    throw new Error()
+  } catch (e: any) {
+    error = e
+    console.log(e.stack)
```
hf, line 329: Remove this line.

src/lib/helpers.ts
```diff
@@ -282,3 +282,112 @@ export async function generatePKCEChallenge(verifier: string) {
...
+/**
+ * Checks if the current caller of the function is in a {@link
+ * #stackGuard} of the provided `name`. Works by looking through
+ * the stack trace of an `Error` object for a special function
+ * name (generated by {@link #stackGuard}).
+ *
+ * @param name The name of the stack guard to check for. Must be `[a-zA-Z_-]` only.
+ */
+export function isInStackGuard(name: string): boolean {
+  STACK_GUARD_CHECK_FN()
```
kangmingtay, line 321: hmm what is the point of this?

hf: To warn users that the environment they're running the library in does not support stack guards due to either transpilation to JS that uses generators (<= ES2015) or in a JavaScript engine that does not support async/await stack traces. 

This will very rarely be encountered, as only some fringe JS engines don't support async/await stack traces.

hf: I updated the warning message to be more instructive.

kangmingtay (approved):

PR #730, opened 2023-07-11
ci: use node 18, 20 for tests

hf:
Node 16 is no longer a supported version.


J0 (approved):
Should we bump the lock file as well? If not alr generated using 18 or 20

PR #731, opened 2023-07-12
feat: add experimental lock option with no-op default

hf:
Adds an `@experimental` `lock` option and uses it in `_useSession`. This can be used to test out different lock implementations before committing to a default one within this library.

It includes an implementation of a `broadcastLock` which _can be used as a value to `lock`_ but **is not done by default intentionally.** This is an experimental API that is bound to have changes and potentially serious bugs. Once it has been well tested by the team and others it is likely to be promoted to the default option.

## broadcastLock

Uses a `BroadcastChannel` to serialize access to a named critical section. The lock has three states: Acquiring, Backoff and Acquired.

When in the **Acquiring** state, a message is broadcast `I will acquire the lock!` and a timeout is started. If any message is received in this state, the lock immediately moves to the Backoff state. If no message is received, the lock moves to the Acquired state. It is assumed that if two messages are posted simultaneously at the channel, that both locks will receive the other's message.

When in the **Backoff** state, the lock sleeps for random amount of time before moving back in the Acquiring state. Each time it enters this state, it waits exponentially longer than the last time.

When in the **Acquired** state, the lock broadcasts `I have the lock, please wait!`. If any message is received on the channel, `I have the lock!` is broadcast immediately. The sender of the first `I will acquire the lock!` message received in this state will be sent the `Go` message after the operation is done which gives it priority over all other competing locks. This helps reduce the time needed for the locks to identify who should go next. Once the operation is done, the lock stops replying with `I have the lock, please wait!` messages on the channel.

Illustration of how `broadcastLock` works

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -77,11 +77,12 @@ import type {
   MFAChallengeAndVerifyParams,
   ResendParams,
   AuthFlowType,
+  LockFunc,
 } from './lib/types'
 
 polyfillGlobalThis() // Make "globalThis" available
 
-const DEFAULT_OPTIONS: Omit<Required<GoTrueClientOptions>, 'fetch' | 'storage'> = {
+const DEFAULT_OPTIONS: Omit<Required<GoTrueClientOptions>, 'fetch' | 'storage' | 'lock'> = {
   url: GOTRUE_URL,
   storageKey: STORAGE_KEY,
   autoRefreshToken: true,
@@ -98,6 +99,10 @@ const AUTO_REFRESH_TICK_DURATION = 30 * 1000
 /**
  * A token refresh will be attempted this many ticks before the current session expires. */
 const AUTO_REFRESH_TICK_THRESHOLD = 3
+
+async function lockNoOp<R>(name: string, acquireTimeout: number, fn: () => Promise<R>): Promise<R> {
+  return await fn()
+}
 
 export default class GoTrueClient {
   private static nextInstanceID = 0
@@ -146,6 +151,7 @@ export default class GoTrueClient {
     [key: string]: string
   }
   protected fetch: Fetch
+  protected lock: LockFunc
 
   /**
    * Used to broadcast state change events to other tabs listening.
@@ -174,6 +180,7 @@ export default class GoTrueClient {
     this.autoRefreshToken = settings.autoRefreshToken
     this.persistSession = settings.persistSession
     this.storage = settings.storage || localStorageAdapter
+    this.lock = settings.lock || lockNoOp
     this.admin = new GoTrueAdminApi({
       url: settings.url,
       headers: settings.headers,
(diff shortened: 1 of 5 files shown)


src/GoTrueClient.ts
```diff
@@ -768,7 +775,85 @@ export default class GoTrueClient {
...
+    } else {
+      return await this._acquireLock('_useSession', async () => {
+        return await stackGuard('_useSession', async () => {
+          // the use of __loadSession here is the only correct use of the function!
+          const result = await this.__loadSession()
+
+          return await fn(result)
+        })
+      })
+    }
```
hf, line 848: Review this until #726 gets merged.

src/lib/locks.ts
```diff
@@ -0,0 +1,240 @@
...
+ * wait!`. If any message is received on the channel, `I have the lock!` is
+ * broadcast immediately. The sender of the first `I will acquire the lock!`
+ * message received in this state will be sent the `Go` message after the
+ * operation is done which gives it priority over all other competing locks.
+ * This helps reduce the time needed for the locks to identify who should go
+ * next. Once the operation is done, the lock stops replying with `I have the
+ * lock, please wait!` messages on the channel.
+ *
+ * Lock wait times have a default of 25ms but can be configured with the
+ * `supabase.dashboard.broadcastLock.wait` local storage key.
```
hf, line 116:
```suggestion
 * `supabase.gotrue-js.broadcastLock.lock-wait` local storage key.
```

src/lib/locks.ts
```diff
@@ -0,0 +1,240 @@
...
+          state = 'backoff'
+          sleepOperation.cancel()
+        } else if (state === 'acquired') {
+          // any message was received, reply that the lock is taken
+          bc.postMessage({ msg: 'I have the lock, please wait!', id })
+
+          if (event.data.msg === 'I have the lock, please wait!') {
+            console.error(
+              `broadcastLock(${name})@${id}: multiple tabs have the lock!`,
+              event.data.id
```
hf, line 169:
```suggestion
              event.data
```

src/GoTrueClient.ts
```diff
@@ -1720,9 +1762,16 @@ export default class GoTrueClient {
...
+        }, 0 /* try-lock */)
       } catch (e: any) {
-        console.error('Auto refresh tick failed with error. This is likely a transient error.', e)
+        if (e.isAcquireTimeout) {
+          this._debug(
+            '#_autoRefreshTokenTick()',
+            'lock is already acquired, skipping for next tick'
+          )
+        } else {
+          console.error('Auto refresh tick failed with error. This is likely a transient error.', e)
```
J0, line 1773: What would be the best way to advise a customer to debug this if they come back to us with an error?

J0 (approved):

src/GoTrueClient.ts
hf: It's transient, but logging so it's not silent.

hf:
I will actually re-open with the `broadcastLock` in a new PR and just the locking code here.

PR #735, opened 2023-07-14
feat: parse `expires_at` if present

hf:
Related to: https://github.com/supabase/gotrue/pull/1183

Built on top of: #733 

Parses `expires_at` from the URL or request if present. Emits warnings if clock skew issues are detected.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -1312,6 +1312,7 @@ export default class GoTrueClient {
         access_token,
         refresh_token,
         expires_in,
+        expires_at,
         token_type,
       } = params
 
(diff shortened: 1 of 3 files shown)


src/GoTrueClient.ts
```diff
@@ -1231,7 +1232,35 @@ export default class GoTrueClient {
...
+      if (timeNow - issuedAt >= 120) {
+        console.warn(
+          '@supabase/gotrue-js: Session as retrieved from URL was issued over 120s ago, URL could be stale',
+          issuedAt,
+          expiresAt,
+          timeNow
+        )
+      } else if (timeNow - issuedAt < 0) {
+        console.warn(
+          '@supabase/gotrue-js: Session as retrieved from URL was issued in the future? Check the device clok for skew',
```
J0, line 1258:
```suggestion
          '@supabase/gotrue-js: Session as retrieved from URL was issued in the future? Check the device clock for skew',
```

J0 (approved):

PR #736, opened 2023-07-15
feat: add _acquireLock and navigatorLock

hf:
Built on top of: #734 

Adds an experimental `lock` option. This is used internally within `_acquireLock()` which can be used to acquire a global exclusive lock when using the client. It can prevent issues with concurrent access to local storage.

`navigatorLock` is an experimental lock that can be used to test the library before officially designating it (or an alternative) as the default global exclusive locking mechanism. It will be rolled out on supabase.com/dashboard over time as we grow our confidence in the library.

### DIFF
--- a/src/lib/locks.ts
+++ b/src/lib/locks.ts
@@ -1,0 +1,161 @@
+/**
+ * @experimental
+ */
+export const internals = {
+  /**
+   * @experimental
+   */
+  debug: !!(
+    globalThis &&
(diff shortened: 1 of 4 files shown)


src/lib/locks.ts
```diff
@@ -0,0 +1,161 @@
+/**
```
hf, line 1: Review this file until #732 is merged.

J0 (approved):

PR #738, opened 2023-07-18
Feat: client side error for larger passwords

mansueli:

Throws a client-side error when a password larger than 72 chars is attempted.
Documenting:
https://github.com/supabase/gotrue/issues/1101

Right now the password is truncated without any explicit indication that this happens to the user

## What is the new behavior?

It throws an error, so it is clear that passwords should be limited to 72 chars to avoid mistakes. 

## Additional context

### DIFF
--- a/src/GoTrueAdminApi.ts
+++ b/src/GoTrueAdminApi.ts
@@ -119,9 +119,6 @@ export default class GoTrueAdminApi {
     try {
       const { options, ...rest } = params
       const body: any = { ...rest, ...options }
-      if (options?.password && options.password.length > 72) {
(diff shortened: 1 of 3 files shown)


mansmahamat (approved):

geositta:
@kangmingtay would really appreciate if you could review since the current implementation is not a good experience for users that want to use long passwords generated by password manager apps.

kangmingtay (changes requested):
i noticed that we closed the issue [here](https://github.com/supabase/gotrue/issues/1101) but i think we should be returning the error from gotrue instead of returning a custom error message in the client lib

PR #748, opened 2023-07-24
Improved typing of provider fields

TheOnlyTails:
This changes the type of the `provider` fields on `UserAppMetadata` and `UserIdentity` from `string` to `Provider`, which is the more accurate type.

Right now, the field is typed as `string`, which tells typescript it can be any string, when in fact it can only ever be one of the values of `Provider`.

this is my first PR here, let me know of there's something I missed!

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -231,7 +231,7 @@ export interface UserIdentity {
   identity_data?: {
     [key: string]: any
   }
-  provider: string
+  provider: Provider
   created_at?: string
   last_sign_in_at?: string
   updated_at?: string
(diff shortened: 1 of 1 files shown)


hf:
Hey, actually it can only be a string. This is the information that is coming out from the GoTrue server. If you use an older version of gotrue-js, but a newer version of gotrue (the server) you will not be able to recognize any new providers supported by the server. This is why it has to be a string.

Furthermore, SSO providers are not enumerable, i.e. the provider name is `sso:<uuid>` as they're registered on the server. In that case you also can't use a const-style type.

I'll close this PR, though absolutely feel free to suggest changes whenever you see anything that can be improved. If you want to change something but are not sure, open an issue where you can discuss approaches with the team and community before spending time on it.

TheOnlyTails:
Alright, no problem! it only took me 2 minutes so I didn't waste too much time :sweat_smile:

PR #749, opened 2023-08-01
fix: localStorage not allowed error

naoyayamamoto:

Bug fix

If localStorage is not allowed, e.g. `inside 'data:' URLs.`, an error occurs and it does not work correctly.

## What is the new behavior?

Use `supportsLocalStorage` to check if localStorage is available.

## Additional context

error

スクリーンショット 2023-08-01 20 11 44


naoyayamamoto:
#753 and it's cured, so I'll close it.

PR #754, opened 2023-08-07
feat: add a different synchronization approach

hf:
Previously, the library attempted to synchronize access to the storage across browsers and tabs using a global lock initiated by the `_useSession` function. Because some recursive calls (i.e. `_useSession` called within `_useSession`) were likely, the concept of a stack guard was implemented to detect such calls. This prevented deadlocks with recursion. However, stack guards appear to be quite flaky depending on the environment in which the library runs (not supported in WebKit based platforms) or transpilation/compilation/minifaction toolchains.

This new approach removes the stack guard concept, and attempts to solve synchronization "externally" from the GoTrueClient. This is done like so:

- GoTrueClient no longer deals with acquiring the lock. All methods assume that when they're executing, they have been globally synchronized and there is no other competing client.
- The lock is acquired only when certain methods like `getSession`, `getUser` are called _from outside of the gotrue-js module._ It is implemented by exporting a wrapped `GoTrueClient` with a `Proxy` that detects those methods and acquires the lock before calling into them. These methods are annotated with `@synchronized`.
- Finally, initialization is also handled at the proxy level by acquiring the lock just before the constructed client is returned. The synchronized methods will wait for the initialization to finish before acquiring the lock and calling into the GoTrueClient code.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -106,7 +106,11 @@ async function lockNoOp<R>(name: string, acquireTimeout: number, fn: () => Promi
   return await fn()
 }
 
-export default class GoTrueClient {
+function synchronized(target: any, propertyKey: string, propertyDescriptor: PropertyDescriptor) {
+  target[propertyKey].isSynchronized = true
+}
+
+class GoTrueClientInternal {
   private static nextInstanceID = 0
 
   private instanceID: number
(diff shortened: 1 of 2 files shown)


src/GoTrueClient.ts
```diff
@@ -807,31 +790,15 @@ export default class GoTrueClient {
...
-
-        return await fn()
-      }
-
-      if (isInStackGuard('_acquireLock')) {
-        this._debug('#_acquireLock', 'recursive call')
-        return await fn()
-      }
-
       return await this.lock(`lock:${this.storageKey}`, acquireTimeout, async () => {
```
kangmingtay, line 797: just to clarify, the default implementation of the lock doesn't do anything with the `name` and `acquireTimeout` arguments right?

hf:
Closing in favor of #757.

PR #759, opened 2023-08-14
Update SignInWithIDToken docstring

J0:
Update tsdoc to be on one line for aesthetic purposes  when used in docs- currently split into two lines which looks slightly disjoint

CleanShot 2023-08-14 at 12 09 48@2x


J0:
wait, actually I think it's just the breakpoint nvm

PR #761, opened 2023-08-21
feat: `navigatorLock` check for spec compatibility

hf:
It looks like there are some instances where Chrome returns a `null` lock object from [Navigator LockManager request](https://developer.mozilla.org/en-US/docs/Web/API/LockManager/request) even though this should only be the case if `ifAvailable` is set to `true`. This change logs a warning.


kangmingtay (approved):
approved first but can we get the test to pass before merging please

PR #772, opened 2023-08-30
feat: pkce challenge support in React Native with Segment

MarceloPrado:

This PR fixes an issue where gotrue would throw in React Native due to the lack of the crypto and TextEncoder module.

When [integrating Segment with React Native](https://segment.com/docs/connections/sources/catalog/libraries/mobile/react-native/), they ask you to install [react-native-get-random-values](https://github.com/LinusU/react-native-get-random-values) as a polyfill to crypto's `getRandomValue` method. This causes issues in gotrue, as the code does a simple `crypto !== undefined` check.

Since a lot of RN libraries polyfill only a subset of crypto's method, `crypto !== undefined` is not a bullet-proof way to check for support.

If you have Segment installed, and try to login with `supabase.auth.signInWithOtp` (email), the app throws an error.

## What is the new behavior?

If the required crypto modules are not available, it just fallbacks to plain text as the existing behavior instead of throwing an error.

### DIFF
--- a/src/lib/helpers.ts
+++ b/src/lib/helpers.ts
@@ -289,7 +289,9 @@ function base64urlencode(str: string) {
 }
 
 export async function generatePKCEChallenge(verifier: string) {
-  if (typeof crypto === 'undefined') {
+  const hasCryptoSupport = typeof crypto !== 'undefined' && typeof crypto.subtle !== 'undefined' && typeof TextEncoder !== 'undefined';
+  
+  if (!hasCryptoSupport) {
     console.warn(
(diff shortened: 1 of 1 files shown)


kangmingtay (approved):

kangmingtay:
hey @MarceloPrado, apologies for the delay in reviewing but thanks for putting this together!

MarceloPrado:
Thanks @kangmingtay! Any chance you could also review https://github.com/supabase/supabase-js/pull/878 ?

PR #774, opened 2023-08-31
feat: provide default storage when `persistSession` is false or `localStorage` is not supported

hf:
Provides a default memory-based, per client, storage interface when the `persistSession` is set to false (which previously used a similar approach) or when the `storage` option is not set and the current environment is not a browser or has no `localStorage` registered on `globalThis`.

With this PR, the warning that was often emitted in server-side environments is gone, as those environments now gracefully and safely fall back to using per-client memory storage.


src/GoTrueClient.ts
```diff
@@ -1756,11 +1748,7 @@ export default class GoTrueClient {
   private async _removeSession() {
     this._debug('#_removeSession()')
 
-    if (this.persistSession) {
-      await removeItemAsync(this.storage, this.storageKey)
-    } else {
```
J0, line 1761: Ah okay was slow to figure it out so calling it out explicitly for future ref feel free to correct me if my understanding is off.

Looks like we no longer need to check `persistSession` since there's a default `inMemorySession` when `persistSession` is set to false. Consequently there's no need to check for the `persistSession` before calling `removeItemAsync` since it will now default to the in memory session

J0 (approved):
LGTM, thanks for putting this together!

hf:
Tested this with Supabase Dashboard. Seems to be working!

PR #781, opened 2023-09-07
chore: add response type for secure email change (singleConfirmation)

J0:
see title


J0:
This PR is still relevant but the tests need to be updated. Will  update tomorrow

J0:
Actually let's close it and re-open tmrw

PR #785, opened 2023-09-12
feat: allow customizing the debug log function

hf:
Allows customizing the debug log function (which is `console.log` by default) to support complex logging scenarios such as writing the logs to IndexedDB or sending them to a server or other monitoring software.


J0 (approved):
It looks fine although ideally IMO this would have been a config object rather than bool (e.g.

```js
debug: {
   ...
   handler: ()=> ....
}
)
```
Though I guess we have to leave as is for backward compatibility since it's currently already a bool. Maybe can consider for a release in the far future

PR #788, opened 2023-09-18
feat: acquire locks around mfa methods

hf:
Given the changes introduced in #736, some of the MFA methods were exempt from acquiring the lock. However, this appears to have been a bad idea, especially around `_getAuthenticatedAssuranceLevel` which is used while the user is authenticated.

This PR adds the locks in the correct places.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -2155,70 +2155,74 @@ export default class GoTrueClient {
    * {@see GoTrueMFAApi#verify}
    */
   private async _verify(params: MFAVerifyParams): Promise<AuthMFAVerifyResponse> {
-    try {
-      return await this._useSession(async (result) => {
-        const { data: sessionData, error: sessionError } = result
-        if (sessionError) {
-          return { data: null, error: sessionError }
-        }
-
-        const { data, error } = await _request(
-          this.fetch,
-          'POST',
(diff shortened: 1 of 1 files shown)


J0 (approved):
Consider running prettier to minimize the diff

Otherwise lgtm as it looks like a mostly mechanical change

src/GoTrueClient.ts
```diff
@@ -2257,67 +2265,71 @@ export default class GoTrueClient {
...
    */
   private async _listFactors(): Promise<AuthMFAListFactorsResponse> {
-    const {
-      data: { user },
-      error: userError,
-    } = await this._getUser()
-    if (userError) {
-      return { data: null, error: userError }
-    }
+    return this._acquireLock(-1, async () => {
```
kangmingtay, line 2268: why is it necessary to acquire the lock for listing factors?

kangmingtay (changes requested):

src/GoTrueClient.ts
hf: Good question! It's because of `_getUser` -- I did this somewhat mechanically. But yeah I can re-do it with calling `getUser` and it wouldn't be necessary! Thanks for spotting it!

hf:
Tested on supabase.com.

PR #790, opened 2023-10-02
fix: expose role in admin user type

kangmingtay:
* Exposes the `role` field in the admin user type

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -378,6 +378,15 @@ export interface AdminUserAttributes extends Omit<UserAttributes, 'data'> {
    * Setting the ban duration to 'none' lifts the ban on the user.
    */
   ban_duration?: string | 'none'
+
+  /**
+   * The `role` claim set in the user's access token JWT.
+   *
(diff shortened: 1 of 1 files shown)


src/lib/types.ts
```diff
@@ -378,6 +378,15 @@ export interface AdminUserAttributes extends Omit<UserAttributes, 'data'> {
    * Setting the ban duration to 'none' lifts the ban on the user.
    */
   ban_duration?: string | 'none'
+
+  /**
+   * The `role` claim set in the user's access token JWT.
+   *
+   * When a user signs up, this role is set to `authenticated` by default. You should only modify the `role` if you need to provision several levels of admin access that have different permissions on individual columns in your database.
+   *
+   * Setting this role to `service_role` is not recommended as it grants the user admin privileges.
```
J0, line 387: I guess we can block role from being set to "service_role" at the client library if issues surface since users can default to directly running the SQL query if they are certain that the role change is necessary

J0 (approved):

src/lib/types.ts
kangmingtay: they will still be able to bypass it by hitting the gotrue admin endpoint directly

J0: Yup, but they'd have to go out of their way in order to do it as with writing to the db

PR #803, opened 2023-10-19
feat: Notify when refresh token fails

tungs0ul:
Add event `REFRESH_TOKEN_FAILED` and notify when `_callRefreshToken` failed


hf:
This would be the SIGNED_OUT event, no?

tungs0ul:
> This would be the SIGNED_OUT event, no?

I have changed the event to `SIGNED_OUT`, I want to capture this event. However, it's currently failing and the session becomes null without any notification.

PR #812, opened 2023-11-15
chore: add friendly name to enroll response

J0:

We introduced friendly name in the response of `enroll` a while back: https://github.com/supabase/gotrue/pull/1277

This PR adds the field to the client library response

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -820,7 +820,7 @@ export type AuthMFAEnrollResponse =
            * to use it. Avoid loggin this value to the console. */
           uri: string
         }
-        /** Friendly Name of the Factor **/
+        /** Friendly name of the factor, useful for distinguishing between factors **/
         friendly_name: string
       }
       error: null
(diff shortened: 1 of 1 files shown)


src/lib/types.ts
```diff
@@ -820,6 +820,8 @@ export type AuthMFAEnrollResponse =
            * to use it. Avoid loggin this value to the console. */
           uri: string
         }
+        /** Friendly Name of the Factor **/
```
hf, line 823: Can you use consistent casing?

```suggestion
        /** Friendly name of the factor, useful for distinguishing between factors **/
```

J0:
Ah this probably should be reverted to draft until we deploy

src/lib/types.ts
```diff
@@ -820,6 +820,8 @@ export type AuthMFAEnrollResponse =
            * to use it. Avoid loggin this value to the console. */
           uri: string
         }
+        /** Friendly name of the factor, useful for distinguishing between factors **/
+        friendly_name: string
```
hf, line 824: Shouldn't it be `friendly_name?: string`. It will only be there if it's been added previously, right?

hf (approved):

src/lib/types.ts
J0: Good catch, thanks!

PR #813, opened 2023-11-29
fix: add password recoery flow support for pkce

dshukertjr:

Currently, `PASSWORD_RECOVERY` auth event is not emitted during a password recovery flow using pkce. 

Calling `resetPasswordForEmail()` method 

## What is the new behavior?

Calling `resetPasswordForEmail()` sets a code verifier and the auth flow `PASSWORD_RECOVERY` with a `/` separator on the local storage like this: 
`${codeVerifier}/PASSWORD_RECOVERY`

Within `_exchangeCodeForSession()`, the auth flow and the code verifier is extracted from local storage, and if the auth flow is `PASSWORD_RECOVERY`, then `PASSWORD_RECOVERY` event is fired instead of `SIGNED_IN` event.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -499,11 +499,16 @@ export default class GoTrueClient {
     })
   }
 
-  private async _exchangeCodeForSession(authCode: string): Promise<AuthTokenResponse> {
-    const [codeVerifier, type] = (
+  private async _exchangeCodeForSession(authCode: string): Promise<
+    | {
+        data: { session: Session; user: User; redirectType: string | null }
+        error: null
+      }
+    | { data: { session: null; user: null; redirectType: null }; error: AuthError }
+  > {
+    const [codeVerifier, redirectType] = (
       (await getItemAsync(this.storage, `${this.storageKey}-code-verifier`)) as string
     ).split('/')
-    const isPasswordRecoveryFlow = type === 'PASSWORD_RECOVERY'
     const { data, error } = await _request(
       this.fetch,
       'POST',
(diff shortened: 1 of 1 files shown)


src/GoTrueClient.ts
```diff
@@ -1538,7 +1545,11 @@ export default class GoTrueClient {
     let codeChallengeMethod: string | null = null
     if (this.flowType === 'pkce') {
       const codeVerifier = generatePKCEVerifier()
-      await setItemAsync(this.storage, `${this.storageKey}-code-verifier`, codeVerifier)
+      await setItemAsync(
+        this.storage,
+        `${this.storageKey}-code-verifier`,
+        `${codeVerifier}/PASSWORD_RECOVERY`
```
dshukertjr, line 1551: Using `/` as the separator here, becuase `/` is not used for generating code verifiers. 
https://github.com/supabase/gotrue-js/blob/master/src/lib/helpers.ts#L265

J0 (approved):
Thanks for catching and fixing it @dshukertjr !

src/GoTrueClient.ts
```diff
@@ -499,8 +499,16 @@ export default class GoTrueClient {
...
+  private async _exchangeCodeForSession(authCode: string): Promise<
+    | {
+        data: { session: Session; user: User; redirectType: string | null }
+        error: null
+      }
+    | { data: { session: null; user: null; redirectType: null }; error: AuthError }
+  > {
+    const [codeVerifier, redirectType] = (
+      (await getItemAsync(this.storage, `${this.storageKey}-code-verifier`)) as string
+    ).split('/')
```
ddunkin, line 511: This throws `TypeError: Cannot read properties of null (reading 'split')` when the item is not in storage, e.g. if a magic link is opened in a different browser than used to create it.

dshukertjr: Thanks for notifying about this error. A PR to fix this issue is here:
https://github.com/supabase/gotrue-js/pull/825

PR #816, opened 2023-12-04
feat: base64-encode session value

jebibot:

Feature

The session cookie value is a plain JSON string, which gets encoded in ASCII-only storages, such as cookie store. Especially, this causes a mismatch in the length calculation in the cookie chunker in `@supabase/ssr`; sometimes, cookie values exceed the max length.

## What is the new behavior?

This PR base64-encodes the session value before saving in the storage.

The rationale for using base64:
- it is used in the JWT encoding
- URL encoding encodes `{`, `}`, `"`, `]` to 3 characters tripling in length, whereas base64 encoding increases `x4/3` in bytes, so the overall length would be similar

## Additional context

- Related
  - https://github.com/supabase/auth-helpers/issues/643
  - https://github.com/supabase/auth-helpers/issues/696

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -993,7 +993,7 @@ export default class GoTrueClient {
     try {
       let currentSession: Session | null = null
 
-      const maybeSession = await getItemAsync(this.storage, this.storageKey)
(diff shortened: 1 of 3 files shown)


hf:
I would prefer this be fixed in SSR instead of in this library.

PR #820, opened 2023-12-11
rename: `gotrue` to `auth`

hf:
Renames GoTrue to Auth.

`GoTrueClient` and `GoTrueAdminApi` still remain the main classes, but they're re-exported as `AuthClient` and `AuthAdminApi`.

### DIFF
--- a/infra/docker-compose.yml
+++ b/infra/docker-compose.yml
@@ -2,7 +2,7 @@
 version: '3'
 services:
   gotrue: # Signup enabled, autoconfirm off
-    image: supabase/gotrue:latest
+    image: supabase/auth:v2.129.0
     ports:
       - '9999:9999'
     environment:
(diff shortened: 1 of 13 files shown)


kangmingtay (approved):

J0 (approved):

infra/docker-compose.yml
```diff
@@ -2,7 +2,7 @@
 version: '3'
 services:
   gotrue: # Signup enabled, autoconfirm off
-    image: supabase/gotrue:latest
```
kangmingtay, line 5: why do we pin the version here?

infra/docker-compose.yml
```diff
@@ -2,7 +2,7 @@
...
+    image: supabase/auth:v2.129.0
```
kangmingtay, line 5: why do we not wanna pin against `latest`?

hf: It's not published. We need to see what we do with this https://github.com/supabase/gotrue/blob/master/.github/workflows/publish.yml#L36

PR #824, opened 2023-12-15
feat: return `weakPassword` information after sign-in

hf:
Returns the weak password reasons on sign-in.

### DIFF
--- a/src/lib/fetch.ts
+++ b/src/lib/fetch.ts
@@ -1,6 +1,7 @@
 import { expiresAt, looksLikeFetchResponse } from './helpers'
 import {
   AuthResponse,
+  AuthResponsePassword,
   SSOResponse,
   GenerateLinkProperties,
   GenerateLinkResponse,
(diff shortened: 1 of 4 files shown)


J0 (approved):

src/lib/fetch.ts
```diff
@@ -174,6 +175,25 @@ export function _sessionResponse(data: any): AuthResponse {
   return { data: { session, user }, error: null }
 }
 
+export function _sessionResponsePassword(data: any): AuthResponsePassword {
+  const response = _sessionResponse(data) as AuthResponsePassword
+
+  if (
+    !response.error &&
+    data.weak_password &&
+    typeof data.weak_password === 'object' &&
```
J0, line 184: if possible, consider assigning semantic meaning to parts of this check. It's a little hard to read at first glance.

hf: Yeah will probably remove this validation in a follow up PR, doesn't make much sense.

PR #825, opened 2023-12-17
fix: Type error on code exchange when no item in storage

dshukertjr:

Currently when code exchange for pkce flow is triggered on device that didn't start the auth flow (e.g. signed up via magic link on laptop and clicked on the link on phone), `TypeError: Cannot read properties of null (reading 'split')` error is thrown. This PR fixes that.

Fixes the error pointed out here
https://github.com/supabase/gotrue-js/pull/813#discussion_r1428876014


J0 (approved):
Thanks Tyler!

PR #828, opened 2023-12-20
ci: only publish @supabase/gotrue-js if version is not 0.0.0

hf:
If the previous step with `npx semantic-release` needs to publish a version, it will update the `package.json` file with the correct version. Only then should `@supabase/gotrue-js` be published.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,11 +39,15 @@ jobs:
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
         run: |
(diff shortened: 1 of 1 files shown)


kangmingtay (approved):

kangmingtay:
@hf why do we want to skip the 0.0.0 version?

hf:
> @hf why do we want to skip the 0.0.0 version?

It fails if there's no release because the commits didn't have feat/fix prefixes: https://github.com/supabase/gotrue-js/actions/runs/7276307163/job/19825873842

PR #831, opened 2023-12-28
signup method throws an error instead of returning null user and sess…

dsl400:

minor change to the signup method

for error responses auth.signUp returned `{ data: { user: null, session: null }, error: error } `

## What is the new behavior?

for error responses auth.signUp throws an error.


kangmingtay (commented):
hi @dsl400, what's the motivation behind this change?

PR #832, opened 2023-12-29
fix: provide statusText as error message on retryable network errors

wattroll:

This PR fixes a minor, but annoying lack of error message and adds a test case for an uncovered branch.

Network errors (502,503,504) result into `AuthRetryableFetchError` (correct), with `"{}"` as an `Error#message` (embarassing if displayed to the end user). 

Screenshot of an error in the console

The bug is caused by erroneous usage of `_getErrorMessage(error)` helper in a branch where `looksLikeFetchResponse(error)` assertion have succeeded and `error.status` is `502 | 503 | 504`. I believe `_getErrorMessage` was intended to be called on something that has been thrown or returned as an error, not on a resolved fetch response that was not `ok`. 

The dreaded `"{}"` comes from the final `JSON.stringify` fallback of `_getErrorMessage`.

## What is the new behavior?

I went for the simplest solution of providing `Response.statusText` property as an error message. Status text is the only human-readable piece of information that is reliably/readily available in this context. It is also reasonable to expect it to be clear and accurate.

## Additional context

I briefly considered pulling in the `Response.text()` and including it in the message, but decided against it. Proxy errors rarely include any extra info there, but often include HTML formatting and styling which wouldn't be appropriate here. Also waiting for `Response.text()` might fail with another error.

This fix should be backwards compatible on a patch level. The `Error` constructor coerces the message to a String (or leaves out an empty string in case of `undefined`) therefore even in case of custom fetch implementations/middlewares that don't provide `statusText` our `AuthRetryableFetchError` will still have a `message` field of a string type.

### DIFF
--- a/src/lib/fetch.ts
+++ b/src/lib/fetch.ts
@@ -42,7 +42,7 @@ async function handleError(error: unknown) {
 
   if (NETWORK_ERROR_CODES.includes(error.status)) {
     // status in 500...599 range - server had an error, request might be retryed.
-    throw new AuthRetryableFetchError(_getErrorMessage(error), error.status)
+    throw new AuthRetryableFetchError(error.statusText, error.status)
   }
 
   let data: any
--- a/test/fetch.test.ts
+++ b/test/fetch.test.ts
@@ -105,6 +105,26 @@ describe('fetch', () => {
       await server.start()
     })
 
+    test('should throw AuthRetryableFetchError when response is a network error', async () => {
+      const route = server
+        .get('/')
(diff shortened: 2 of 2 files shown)


mandarini:
Hi @wattroll  ! Thanks for the contribution and your patience.

This repository is deprecated and has moved to the new [Supabase JS monorepo](ttps://github.com/supabase/supabase-js). Since this PR has been inactive for over a year, I’m going to close it to keep the old repo tidy, before archiving.

Note: This old repository will be archived on October 10, 2025. Thank you again for your effort and understanding!

PR #838, opened 2024-01-16
Add explanation for soft deletion to user delete function

encima:

docs update to provide explanation of soft deletion for delete user endpoint

closes https://github.com/supabase/supabase/issues/20057

## What is the new behavior?

Explains what the soft delete flag does so it can be picked up in the Supabase docs

## Additional context


J0 (approved):

J0:
Thanks @encima !

PR #845, opened 2024-01-19
feat: add kakao to sign in with ID token

hf:
Adds the `kakao` string to the supported `signInWithIdToken` types.

Merge after: https://github.com/supabase/gotrue/pull/1381


J0:
Merging, since #1381 has been merged

PR #846, opened 2024-01-20
feat: warn use of `getSession()` when `isServer` on storage

hf:
Warn about using `getSession()` when the storage has `isSever` to true without previously having called `getUser()`.

Warn each time the `user` is accessed as returned by `getSession()`.

Relates to:
- https://github.com/supabase/auth-helpers/pull/722


J0 (approved):
Looks okay to me.

PR #847, opened 2024-02-03
feat: explicit `cache: no-store` in fetch

hf:
Next.js polyfills the `fetch` on the server side to always cache (what's known as their Data Cache) responses. Quote:

> Next.js has a built-in Data Cache that persists the result of data fetches across incoming server requests and deployments. This is possible because Next.js extends the native fetch API to allow each request on the server to set its own persistent caching semantics.
>
>  Good to know: In the browser, the cache option of fetch indicates how a request will interact with the browser's HTTP cache, in Next.js, the cache option indicates how a server-side request will interact with the server's Data Cache.

This is _incredibly dangerous_ for Auth use cases. Thankfully, it appears that using the default client uses `@supabase/node-fetch` in such a circumstance which should not be affected by the issue.

However, if developers do choose to switch back to the native `fetch` in Next.js the library must inform that HTTP client that there should be no caching under any circumstance on Auth responses.

Opting out docs: https://nextjs.org/docs/app/building-your-application/caching#opting-out-1

### DIFF
--- a/src/lib/fetch.ts
+++ b/src/lib/fetch.ts
@@ -118,7 +118,10 @@ export async function _request(
     fetcher,
     method,
     url + queryString,
-    { headers, noResolveJson: options?.noResolveJson },
+    {
+      headers,
+      noResolveJson: options?.noResolveJson,
+    },
     {},
     options?.body
   )
@@ -138,7 +141,12 @@ async function _handleRequest(
   let result: any
 
   try {
-    result = await fetcher(url, requestParams)
+    result = await fetcher(url, {
+      ...requestParams,
+      // UNDER NO CIRCUMSTANCE SHOULD THIS OPTION BE REMOVED, YOU MAY BE OPENING UP A SECURITY HOLE IN NEXT.JS APPS
+      // https://nextjs.org/docs/app/building-your-application/caching#opting-out-1
+      cache: 'no-store',
+    })
   } catch (e) {
     console.error(e)
 


kangmingtay:
agreed that the default behaviour is dangerous for auth in nextjs but they also provide a way to specify this through the [`noStore`](https://nextjs.org/docs/app/api-reference/functions/unstable_noStore) and `cookies`(https://nextjs.org/docs/app/api-reference/functions/cookies) functions, which will force a cache revalidation each time

src/lib/fetch.ts
```diff
@@ -138,7 +141,10 @@ async function _handleRequest(
   let result: any
 
   try {
-    result = await fetcher(url, requestParams)
+    result = await fetcher(url, {
+      ...requestParams,
+      cache: 'no-cache', // UNDER NO CIRCUMSTANCE SHOULD THIS OPTION BE REMOVED, YOU MAY BE OPENING UP A SECURITY HOLE IN NEXT.JS APPS
```
kangmingtay, line 146: can we also include a link to their docs? (https://nextjs.org/docs/app/building-your-application/caching#opting-out-1)

kangmingtay (approved):

elias123tre:
This has introduced a couple of errors regarding Cloudflare Workers, related:
https://github.com/cloudflare/workerd/issues/698
https://github.com/supabase/supabase-js/issues/1001
https://github.com/cloudflare/workers-sdk/issues/2514
https://github.com/nuxt-modules/supabase/issues/340

src/lib/fetch.ts
```diff
@@ -138,7 +141,12 @@ async function _handleRequest(
   let result: any
 
   try {
-    result = await fetcher(url, requestParams)
+    result = await fetcher(url, {
+      ...requestParams,
+      // UNDER NO CIRCUMSTANCE SHOULD THIS OPTION BE REMOVED, YOU MAY BE OPENING UP A SECURITY HOLE IN NEXT.JS APPS
+      // https://nextjs.org/docs/app/building-your-application/caching#opting-out-1
```
ImLunaHey, line 147: "YOU MAY BE OPENING UP A SECURITY HOLE IN NEXT.JS APPS"

there's nothing at all about that in the link provided. 🤔
maybe the comment can be expanded to actually explain what issue this would cause?

hf: It does in the context of NextJS, as the docs state:

 > The Data Cache is persistent across incoming requests and deployments unless you revalidate or opt-out.
 
 So we have to forcefully opt-out of the cache, as it appears that `GET /user` may be cached just because. There's no docs on whether it's cached by the JWT, or not. We reached out to Vercel -- 0 information on that.

ImLunaHey: maybe add that to the comment? most users are likely going to do the same thing i did and look on that page you linked for something mentioning "security" directly.

PR #850, opened 2024-02-06
ci: add conventional commits linter

hf:
Adds a linter for commits and PRs to ensure that correct structure is observed. Certain directives are not compatible with this project:

- Breaking changes are not allowed. Therefore PR titles or commit messages must not use the `!` nor the `BREAKING CHANGE` directive in descriptions or commit messages.
- The `Release-As` directive is not supported as it can break the release process.
- For cleanliness, only a few conventional commit prefixes are allowed.

These are linted using a simple JavaScript script that runs on PRs and pushes.

Similar to:
- https://github.com/supabase/gotrue/pull/1392

### DIFF
--- a/.github/workflows/conventional-commits-lint.js
+++ b/.github/workflows/conventional-commits-lint.js
@@ -1,0 +1,101 @@
+"use strict";
+
+const fs = require("fs");
+
+const TITLE_PATTERN =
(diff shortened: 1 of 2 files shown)


J0 (approved):
Looks fine to me

Issue #853, opened 2024-02-15
Functions that call `_removeSession` internally do not trigger `SIGNED_OUT` event when the function fails


bombillazo:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

When using the client and auth functions, some internally call a `_removeSession` function, presumably to clear the session before attempting any auth-related logic. This effectively clears out the local session data used to keep the user logged in. In many cases the function runs and restores or resets the session data eventually. However, if the function fails, this does not trigger the SIGNED_OUT to notify any even handlers and hooks that the session data is no longer present, causing inconsistent auth data states.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Create a supabase client instance.
2. In your app, use the `auth-helpers-react` lib and use the `useSession` hook to get the session data.
3. Trigger a function like `supabase.auth.resend` with invalid params to trigger an error and session remove.
4. This causes the session data to be erased, but the context data of the `useSession` hook is not updated and retains the old values.

## Expected behavior

Any auth function that fails and deletes the auth session data should trigger a `SIGNED_OUT` event to notify any other serives.

## System information

- Version of supabase-js: 2.39.6

bombillazo:
Any update on this issue?

kangmingtay:
@bombillazo hmm there shouldn't be a session return by `useSession` when the `supabase.auth.resend` method is called because no user is actually signed in yet

bombillazo:
We can call resend to change email or phone. I dont expect users to be signout out for this

kangmingtay:
@bombillazo yeah we handle that [case](https://github.com/supabase/auth-js/blob/4ecfdda65188b71322753e57622be8eafe97ed6b/src/GoTrueClient.ts#L877-L879) already by not removing the session if the type passed indicates an email change or phone change

bombillazo:
This issue is more related to what happens when something fails with the `auth` calls while using the `useSession` react hook that returns the session. Since this library isn't properly triggering the `SIGNED_OUT` event, when an auth call fails and erases the session from the local client, the react hook keeps returning the cached session.

bombillazo:
This is still open and should be fixed with #854

PR #854, opened 2024-02-16
[fix] Call `SIGNED_OUT` event where session is removed

bombillazo:

This adds the SIGNED_OUT event missing in some logic that clears/logs out the session. 

#853


bombillazo:
Any feedback on this @kangmingtay @hf ? We're needing to patch this library whenever a new version is released.

bombillazo:
Any updates here?

src/GoTrueClient.ts
```diff
@@ -471,14 +474,18 @@ export default class GoTrueClient {
...
         await this._notifyAllSubscribers('SIGNED_IN', session)
+      } else if (currentSession) {
+        await this._saveSession(currentSession)
       }
 
       return { data: { user, session }, error: null }
     } catch (error) {
+      if (currentSession) {
+        await this._saveSession(currentSession)
+      }
```
kangmingtay, line 485: i'm not sure why it's necessary to preserve the existing session when one calls the signup method - we always want to remove the session when signup is called to prevent any possibilities of the user logging in as someone else.

kangmingtay: this applies to all the other sign-in methods too - if you're signing in, there won't be any session in the first place, else you'd be logged in already

bombillazo: This truly depends on the UI implementation. We may want to allow users to directly switch accounts, so it is totally possible to call sign in from being already signed in form another account. Right now, if something fails, the user is being kicked out of their current session. This applies to resend events like email change that can also happen while logged in.

I also dont see how this would log a user to someone else's account. Either the signin is successful and the current session switches to the new one, or it fails and you keep the existing one.

j4w8n: If the team wanted to go down this path, you could simplify a lot of these methods by moving the _removeSession call into the `if` blocks which check for `data.session` or `session?.access_token`. There would be a few exceptions to that though.

I've seen more and more discussions about people trying to implement a "switch account" feature; or they already have, but recent changes have broken their implementations.

bombillazo: I agree. I just wasn't sure if some logic was dependent on having no session while executing, so I preferred keeping it and reverting at the end.

Hopefully, we can remove the current limitation, be it by removing `_removeSession` where not necessary or, as my PR suggests, restoring the session on fail.

jczstudios:
Any updates on this, it's breaking our production Electron app, seems that when a request fails, the user just gets signed out entirely.

j4w8n (commented):
I appreciate the review tag, but this one will ultimately need to be looked at by kangmingtay or hf. Last I knew, the team's time was limited; so, who knows if/when they'll consider this PR further.

kangmingtay:
sorry @bombillazo i was on vacation last week - i'll be looking into this further to see if we can just do away with removing the session prematurely before hitting the underlying auth endpoint - not too sure if it will cause any unintended breaking changes here so we need some time to test this out

kangmingtay:
@bombillazo i've made a separate PR (#915) to address this - would be great if you can take a look at it too

bombillazo:
Hey @kangmingtay thanks for the PR, I merged the current repo to my branch and the only thing missing is a few lines of code to emit the `SIGNED_OUT` event during the `_recoverAndRefresh` function if the session is ultimately removed

bombillazo:
Any update here?

bombillazo:
Hey @kangmingtay , any update on this issue?

bombillazo:
Hey, I know 2.65.0 was released but this is missing to truly close #853

bombillazo:
@kangmingtay @j4w8n @hf Could this be rechecked please? We still need to run auth-js patched on our side due to this

j4w8n:
I see you moved the SIGNED_OUT event trigger into _removeSession. I understand wanting to move it there, but I still think there are one or two cases where calling _removeSession shouldn't necessarily trigger the event.

However, this isn't my party, and the maintainers may be fine with it. I'm not sure it would hurt anything.

src/GoTrueClient.ts
```diff
@@ -2085,11 +2081,11 @@ export default class GoTrueClient {
       // finished and tests run endlessly. This can be prevented by calling
       // `unref()` on the returned object.
       ticker.unref()
-      // @ts-ignore
+      // @ts-expect-error TS has no context of Deno
```
j4w8n, line 2084: Is this needed here, since the Deno scenario is covered below? Or maybe the Deno check should happen first?

j4w8n (approved):
This is ultimately up to the maintainers, but I wrote a couple of things.

src/GoTrueClient.ts
```diff
@@ -1891,6 +1891,7 @@ export default class GoTrueClient {
         this._debug(debugName, 'session is not valid')
         if (currentSession !== null) {
           await this._removeSession()
+          await this._notifyAllSubscribers('SIGNED_OUT', null)
```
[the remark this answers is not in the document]
bombillazo: How it is currently implemented: the SIGNED_OUT event is not being raised every moment the session is removed. We have logic that depends on reacting to the session that no longer exists in the client. This happens any time the session is removed, regardless of why it was removed. This is also how it is described in the documentation:

image

src/GoTrueClient.ts
bombillazo: IMO `@ts-expect-error` is a much better way to document why a typescript error needs to be ignored compared to `@ts-ignore`, I simply replaced the existing override with this one.

bombillazo: I don't know of a way to check for Deno without TS complaining

bombillazo:
Thanks @j4w8n for the questions/feedback, hopefully this can be reviewed by the corresponding devs to keep the ball rolling.

bombillazo:
Any update?

kangmingtay (approved):
hey @bombillazo, apologies for the delayed review on this PR - the changes look good, thanks for helping to fix this issue!

PR #855, opened 2024-02-24
feat: add support for error codes

hf:
Adds support for error codes. All `AuthError` descendants will now have a `code` property which will encode (when present and supported by the server) the reason why the error occurred.

To support this, the library will now advertise `X-Supabase-Api-Version: 2024-01-01` which is the first version of a new versioning strategy that supports a different encoding for error responses.

See:
- https://github.com/supabase/gotrue/pull/1377


src/lib/helpers.ts
```diff
@@ -304,3 +306,24 @@ export async function generatePKCEChallenge(verifier: string) {
   const hashed = await sha256(verifier)
   return base64urlencode(hashed)
 }
+
+const API_VERSION_REGEX = /^2[0-9]{3}-(0[1-9]|1[0-2])-(0[1-9]|1[0-9]|2[0-9]|3[0-1])$/i
```
J0, line 310: Consider adding a comment to indicate that this is `2YYY-MM-DD`

J0 (approved):

PR #858, opened 2024-03-04
feat: add method for anonymous sign-in

kangmingtay:
* https://github.com/supabase/gotrue/issues/68
* Adds the `signInAnonymously` method to support https://github.com/supabase/gotrue/pull/1460
* Passing in user metadata / the captcha token in `signInAnonymously` looks like: 
    `signInAnonymously({ options: { data, captchaToken }})` 
which follows the same format as the other sign up / sign in methods

### DIFF
--- a/example/react/package.json
+++ b/example/react/package.json
@@ -12,6 +12,9 @@
     "react-scripts": "5.0.1",
     "tailwindcss": "^1.8.10"
   },
+  "devDependencies": {
+    "@babel/plugin-proposal-private-property-in-object": "^7.21.11"
+  },
   "scripts": {
     "build:tailwind": "tailwindcss build src/App.css -o src/tailwind.output.css",
     "prestart": "npm run build:tailwind",
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -452,6 +452,13 @@ export interface UpdatableFactorAttributes {
   friendlyName: string
 }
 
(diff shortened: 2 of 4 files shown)


example/react/package.json
```diff
@@ -12,6 +12,9 @@
     "react-scripts": "5.0.1",
     "tailwindcss": "^1.8.10"
   },
+  "devDependencies": {
+    "@babel/plugin-proposal-private-property-in-object": "^7.21.11"
+  },
```
hf, line 17: What is this used for?

src/lib/types.ts
```diff
@@ -452,6 +452,13 @@ export interface UpdatableFactorAttributes {
   friendlyName: string
 }
 
+export type SignInAnonymouslyCredentials = {
+  options?: {
+    data?: object
+    captchaToken?: string
```
hf, line 458: Can you add some tsdocs on the fields?

hf (approved):
The world will be a better place!

example/react/package.json
kangmingtay: it's just a dependency that needs to be added to remove a warning in the react example

senadir:
I tried using this and I'm getting a 422 Password is required error, not sure what I'm missing, nothing in the request or the PR above seems to indicate `is_anonymous` is being passed to the server.

PR #861, opened 2024-03-12
feat: add auth code flow

J0:

POC for discussion. Further elaboration in RFC - please use in conjunction with [GoTrue PR](https://github.com/supabase/gotrue/pull/1478)

This only covers the Auth Code flow on existing pkce methods and does not include the addition of auth code flow for methods which are primarily on the implicit flow (e.g. `generateLink` , `resend`)


J0:
As discussed, we'll move to first class support for token hash instead.

PR #865, opened 2024-03-18
feat: add filter support to listUsers

0x62:

There is already support for a `filter=xyz` query param for the listUsers endpoint, but it is not exposed on the client. This PR adds support for the filter, which allows users to filter the respond of listUsers by a partial or full email address. 

This PR also fixes pagination metadata for page values >9. This was originally proposed in #793, but changes were requested and that PR is now stale so I have implemented those changes here too.

It is not possible to filter users via the SDK.

## What is the new behavior?

You can now filter users.

## Additional context

https://github.com/supabase/gotrue/issues/880


0x62:
@kangmingtay any change you could review this?

I came from https://github.com/supabase/auth/pull/1330#issuecomment-1848007008, but discovered there is no way to actually use this endpoint from the SDK

> hi @JKFSOM, thanks for taking the time to contribute but i don't think we'll be adding this endpoint in - the `GET /admin/users` endpoint actually allows you to filter for a user's email address already:
> 
> for example:
> 
> ```shell
> curl -X GET "http://localhost:9999/admin/users?filter=<user_email_address>" -H "Authorization: Bearer ADMIN_JWT"
> ```

mandarini:
Hi @0x62! Thanks for this PR - we really like it, especially the pagination fix for pages >9! Before we can merge, could you please add:

1. **Minimum filter length validation** (at least 3 characters):
```typescript
if (params?.filter && params.filter.trim().length < 3) {
  throw new AuthApiError('Filter must be at least 3 characters', 400);
}
```
2. **TypeDoc comment warning** about performance on the `listUsers` method:
```typescript
/**
 * 
 * @warning The filter parameter is provided for convenience but may have performance implications on large databases.
 * Consider using pagination without filters for projects with many users.
 */
```
Once you add these safeguards, we'll be happy to merge! Thanks again for your contribution to Supabase! :D :D

0x62:
@mandarini Done!

mandarini:
Closing, moved here: https://github.com/supabase/supabase-js/pull/1741

Issue #856, opened 2024-02-28
Memory Leak


nbarrow-inspire-labs:
This line (https://github.com/supabase/gotrue-js/blob/15c7c8258b2d42d3378be4f7738c728a07523579/src/GoTrueClient.ts#L2040) is causing a memory leak. See this PR (albeit in the wrong repo, I figured that out too late): https://github.com/supabase/supabase-js/issues/983

nbarrow-inspire-labs:
I've been playing with something like this, but it won't work in React Native as there is no support for WeakRef nor the FinalizationRegistry:
```ts
export class DemoClient {
  constructor() {
    IntervalTimer.schedule(this, () => {
      console.log("I SURVIVE!!!");
    }, 2000);
  }
}

class IntervalTimer {

  private ref: WeakRef<Object>;
  private interval: NodeJS.Timeout | null = null;
  private cancelled: boolean = false;

  private constructor(obj: Object, callback: () => void, ms?: number) {
    this.ref = new WeakRef(obj);

    const tick = (): NodeJS.Timeout | null => {

      console.log("tick...");

      // execute the callback if not cancelled and still referenced
      if(!this.cancelled && this.ref.deref() !== undefined) {
        callback();

        // schedule the next tick
        this.interval = setInterval(tick, ms);
      } else {
        this.interval = null;
      }
      return this.interval;
    }

    this.interval = tick();
  }

  public static schedule = (obj: Object, callback: () => void, ms?: number): void => {
    new IntervalTimer(obj, callback, ms);
  }

  public cancel() {
    this.cancelled = true;
    if(this.interval) clearInterval(this.interval);
  }
}
```

nbarrow-inspire-labs:
I just wanted to bump this and see if there's a potential work-around while ReactNative works on supporting `WeakRef`, etc? My example is a good starting point but clearly not production ready (until `WeakRef` finds support).

nbarrow-inspire-labs:
A pub/sub method might be more desireable:

The static class can act as a publisher, using `setInterval` **once** at the static level to emit `tick` events on a regular pattern. All instantiated instances could then subscribe to the events that the static class publisher is emitting; when the static class emits a `tick` event, the instances could 'do' the tick function. 

This would solve the garbage collection issue, as the subscribers exist independently of the one main (static) publisher; if an instance is due for garbage collection, so too would its event subscriber, allowing instances to be garbage collected while the static class publisher is just always running in the background. 

[...]

PR #867, opened 2024-03-19
fix: add memory leak warning to `GoTrueClient` for unsupported runtimes

nbarrow-inspire-labs:
Adds a warning regarding memory leaks in some javascript runtimes that do not support `setInterval` dereferencing (allowing users to better/manually clean-up supabase instances, freeing memory and allowing for garbage collection)

Warning/bug fix

Currently, in environments that are neither `Node.js` nor `Deno`, a memory leak occurs due to `setInterval` not being dereferenced. See https://github.com/supabase/gotrue-js/issues/856. The implementor is never warned that such a memory leak exists for their specific runtime (such as in React Native). 

This implements a warning, at least, that fires only once, to notify implementors to better clean-up instances in their environments (such as manually calling `stopAutoRefresh` when ready to dispose of instances).

## What is the new behavior?

Noted above.

## Additional context

https://github.com/supabase/gotrue-js/issues/856

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -2077,6 +2077,7 @@ export default class GoTrueClient {
     const ticker = setInterval(() => this._autoRefreshTokenTick(), AUTO_REFRESH_TICK_DURATION)
     this.autoRefreshTicker = ticker
 
+    let willUnref = false
     if (ticker && typeof ticker === 'object' && typeof ticker.unref === 'function') {
       // ticker is a NodeJS Timeout object that has an `unref` method
       // https://nodejs.org/api/timers.html#timeoutunref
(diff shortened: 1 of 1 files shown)


src/GoTrueClient.ts
```diff
@@ -2085,12 +2087,19 @@ export default class GoTrueClient {
...
       // @ts-ignore
     } else if (typeof Deno !== 'undefined' && typeof Deno.unrefTimer === 'function') {
       // similar like for NodeJS, but with the Deno API
       // https://deno.land/api@latest?unstable&s=Deno.unrefTimer
       // @ts-ignore
       Deno.unrefTimer(ticker)
+      willUnref = true
+    }
+
+    if(!willUnref && !GoTrueClient.memoryLeakWarningShown) { // prevents the warning from being spammed / shown multiple times
```
hf, line 2100: Please use a single global constant about this, not per-client. Some environments, like Next.js for example, create a client on every request and this would pollute logs.

hf: Also please don't show it when `isBrowser()` returns true.

nbarrow-inspire-labs: @hf I believe this is only shown once, because it uses the static/class variable `memoryLeakWarningShown` (**not** a client/instance variable). This is modeled based on the approach used for https://github.com/supabase/auth-js/blob/99821f4a1f6fdb3a222cd0f660210016e6cc823e/src/GoTrueClient.ts#L117

I'll make the `isBrowser` change now.

src/GoTrueClient.ts
```diff
@@ -2097,7 +2097,7 @@ export default class GoTrueClient {
       willUnref = true
     }
 
-    if(!willUnref && !GoTrueClient.memoryLeakWarningShown) { // prevents the warning from being spammed / shown multiple times
+    if(!willUnref && !GoTrueClient.memoryLeakWarningShown && (typeof isBrowser !== "function" || !isBrowser())) { // prevents the warning from being spammed / shown multiple times
       console.warn("[GoTrueClient] dereferencing `setInterval` is not supported in your JavaScript Runtime. This will cause memory leaks unless you call the `stopAutoRefresh` method before losing the last reference to a `GoTrueClient` instance. If you are using `supabase-js`, this can be accomplished by calling `supabase.auth.stopAutoRefresh()`. For more information, see https://github.com/supabase/gotrue-js/issues/856")
       GoTrueClient.memoryLeakWarningShown = true
```
nbarrow-inspire-labs, line 2102: @hf this line is what causes the message to only be shown once, as it sets a static/class flag after the first showing. Also, this revised version uses the changes you suggested about `isBrowser`.

PR #871, opened 2024-03-30
chore: add missing 'is_anonymous' property to User type

dave-irvine:

Types update

User type is missing 'is_anonymous' property.

## What is the new behavior?

User type has the 'is_anonymous' property.

## Additional context

I'm not 100% certain if this property is optional or not, hoping @kangmingtay can offer some insight as they implemented the feature in https://github.com/supabase/auth/issues/68

Fixes https://github.com/supabase/auth-js/issues/872

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -339,6 +339,7 @@ export interface User {
   role?: string
   updated_at?: string
   identities?: UserIdentity[]
+  is_anonymous: boolean
   factors?: Factor[]
 }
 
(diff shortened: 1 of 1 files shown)


src/lib/types.ts
```diff
@@ -339,6 +339,7 @@ export interface User {
   role?: string
   updated_at?: string
   identities?: UserIdentity[]
+  is_anonymous: boolean
```
kangmingtay, line 342:
```suggestion
  is_anonymous?: boolean
```

kangmingtay (approved):
@dave-irvine thanks for catching this, it will always be returned but lets make it optional since we need to account for users on an older version of supabase auth that don't support anonymous sign-ins

PR #874, opened 2024-04-01
fix: check for no error, instead of data, on insecure ssr getSession() warning

hf:
Checks for no error instead of the presence of data when showing the warning about potential insecure use of `getSession()`.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -1174,7 +1174,7 @@ export default class GoTrueClient {
       return await this._getUser()
     })
 
-    if (result.data && this.storage.isServer) {
+    if (!result.error && this.storage.isServer) {
       // no longer emit the insecure warning for getSession() as the access_token is now authenticated
       this.insecureGetSessionWarningShown = true
     }


j4w8n:
I'd like to point out that even though this change is how it should work, it's actually going to cause more console warnings when there's no logged-in user. Because `result.error` is going to be true in that case - returning an `invalid claim: missing sub claim` error and will not trigger a suppression.

Even worse, people will call `getUser()` before `getSession()`, as they've been told to do, but there will be this edge case where the warning will still log. This is going to cause more frustration.

j4w8n:
After doing a lot of testing, I have two suggestions and one observation. Hopefully they're not ridiculous and a waste of your time.

1. Also change the corresponding check in `getSession()` [here](https://github.com/supabase/auth-js/blob/master/src/GoTrueClient.ts#L935), because `result.data` will always be true, even if there is no user logged-in; as it will return `{ session: null }`. But perhaps that's your intention - to run the resulting code no matter what, but if that were the case, you could just do the server storage check.
```diff
- if (result.data && this.storage.isServer) {
+ if (result.data.session && this.storage.isServer) {
```
2. Don't suppress the warning after calling `getSession()` [here](https://github.com/supabase/auth-js/blob/master/src/GoTrueClient.ts#L940) - unless we're just trying to avoid being a logging nuisance. I'm unclear why we're doing this if `getSession()` can't be trusted on the server-side anyway. Doesn't it allow someone to bypass calling `getUser()` first to avoid further warnings?
```diff
- this.insecureGetSessionWarningShown = true;
```

Aside from that, I don't understand the code and comment [here](https://github.com/supabase/auth-js/blob/master/src/GoTrueClient.ts#L1130L1131). It's causing the warning to log three times, even after calling `getUser()` with a logged-in user - at least in my base SvelteKit + SSR implementation.

kangmingtay:
closing since this is outdated

PR #876, opened 2024-04-06
fix: `getUser` returns null if there is no session

hf:
If there is no session (i.e. no JWT/access token), `getUser()` cannot possibly work. It now returns null user.

Previously, it just sent out a request _without the JWT_. When combined with `@supabase/supabase-js` which does some clever tricks with `fetch` by adding a default `Authorization` header using the Supabase `anon` API key, if you called `getUser()` at the wrong time an error such as `missing sub claim` error message would be thrown by Supabase Auth.

(Unfortunately the Supabase `anon` API key is signed with the same JWT secret, so it's hard to disambiguate why this is happening.)

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -1198,6 +1198,11 @@ export default class GoTrueClient {
           throw error
         }
 
+        if (!data.session?.access_token) {
+          // if there's no access token, the user can't be fetched
+          return { data: { user: null }, error: new AuthSessionMissingError() }
+        }
+
         return await _request(this.fetch, 'GET', `${this.url}/user`, {
           headers: this.headers,
           jwt: data.session?.access_token ?? undefined,


src/GoTrueClient.ts
```diff
@@ -1198,6 +1198,11 @@ export default class GoTrueClient {
           throw error
         }
 
+        if (!data.session?.access_token) {
+          // if there's no access token, the user can't be fetched
+          return { data: { user: null }, error: null }
```
kangmingtay, line 1203: hmm should we consider returning the `AuthSessionMissingError` instead of null? i feel like that makes the DX better because the developer can check against that and redirect the user to the login page

kangmingtay (approved):
we should also add this check for all other auth authorized endpoints that require the access token and is wrapped by `useSession` or else they will return the same `invalid sub claim error` if no session is present

src/GoTrueClient.ts
j4w8n: I'd say returning null is best here, because `getUser` is just trying to read information; similar to `getSession`.

Currently, the `AuthSessionMissingError` is only thrown for methods which try to change or set something - update a user, refresh a session, set a session, reauthenticate; none of which will succeed unless there is a session. By contrast, `getUser` has no real success/fail characteristic - the user info is either there or it isn't.

j4w8n: Actually, this can't return an error, or it's going to cause issues with https://github.com/supabase/auth-js/pull/874

hf: I think we'll return error, and deal with #874 differently...

Issue #873, opened 2024-03-31
Can't get rid of getUser() warning


jdgamble555:
# Bug report

I keep getting the `getSession()` error about 500 times on one page.
> Using the user object as returned from supabase.auth.getSession() or from some supabase.auth.onAuthStateChange() events could be insecure! This value comes directly from the storage medium (usually cookies on the server) and many not be authentic. Use supabase.auth.getUser() instead which authenticates the data by contacting the Supabase Auth server.

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [ ] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

This warning should not exist. I do not want to contact supabase on my non-logged in pages to see if a user is logged in. That would be two round trips, which would slow down my app.  There is nothing wrong with checking for a session cookie on my server without doing an extra request. Either way, I am following the tutorials exactly. 

## To Reproduce

Follow any one of the tutorials:

https://supabase.com/docs/guides/auth/server-side/creating-a-client

## Expected behavior

Do not show the warning at all, or allow me to disable warnings. At the very least, do not show it 500 times on one page.

## System information

- supabase - 1.151.1
- @supabase/ssr - 0.1.0
- @supabase/supabase-js - 2.41.1

## Additional context

It is an issue with [this line of code](https://github.com/supabase/auth-js/blob/92fefbd49f25e20793ca74d5b83142a1bb805a18/src/GoTrueClient.ts#L936)

https://github.com/supabase/auth-helpers/issues/755

J

dHofmeister:
I have the same issue after running a rm -rf node_modules and reinstalling for sveltekit SSR. I've followed those guides exactly and even on the home page load this error shows up on first load. 

```
  VITE v5.2.7  ready in 499 ms

  ➜  Local:   http://localhost:5173/
  ➜  Network: http://192.168.50.35:5173/
  ➜  Network: http://10.0.0.1:5173/
  ➜  press h + enter to show help
Using supabase.auth.getSession() is potentially insecure as it loads data directly from the storage medium (typically cookies) which may not be authentic. Prefer using supabase.auth.getUser() instead. To suppress this warning call supabase.auth.getUser() before you call supabase.auth.getSession().
[...]

PR #879, opened 2024-04-18
fix: update session warning

kangmingtay:
* removes the warning from being logged everytime `getSession` is called
* only log the warning if the user property is being accessed from the session
* addresses #873, https://github.com/supabase/supabase-js/issues/1010


hf (approved):
🤞 this reduces the noise.

Anyone reading this! Please be careful how you use the output of `getSession()` to secure your apps!

PR #882, opened 2024-04-20
fix: check for access token in header

### LINKED ISSUE #881: supabase-js@2.42.5 breaks client auth with edge functions
# Bug report

- [ ] I confirm this is a bug with Supabase, not with my own application. 
- [ ] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

When I get the authorization header from the client in an edge function I'm unable to make a supabase client with the newest version of supabase-js (using deno edge functions). The behavior doesn't happen on 2.42.4 and below. I have rolled back but I wanted to bring this to your attention.

## To Reproduce

1. Go to '…'
2. Click on '…'
3. Scroll down to '…'

## Expected behavior

Able to get user auth details. See code snippet at bottom of how I typically do that.

## System information

- OS: [e.g. macOS, Windows]
- Browser (if applies) [e.g. chrome, safari]
- Version of supabase-js: [e.g. 6.0.2]
- Version of Node.js: [e.g. 10.10.0]

## Additional context

`import { corsHeaders } from '../_shared/cors.ts'
import { createClient } from "supabase-js"
import { createSupabaseClient } from '../_shared/supabaseClient.ts'

Deno.serve(async (req) => {
  if (req.method === 'OPTIONS') {
    return new Response('ok', { headers: corsHeaders })
  }

  const authHeader = req.headers.get('Authorization')!
  const supabaseClient = createSupabaseClient(authHeader)

  // Get the session or user object
  const { data } = await supabaseClient.auth.getUser()
  const user = data.user

  if (!user) {
    throw new Error('User is null');
  }

}`

and here is the definition for my client function:
`import { createClient, SupabaseClient } from "supabase-js"

export function createSupabaseClient(authHeader: string): SupabaseClient {
  return createClient(
    Deno.env.get('SUPABASE_URL') ?? '',
    Deno.env.get('SUPABASE_ANON_KEY') ?? '',
    { global: { headers: { Authorization: authHeader } } }
  );
}`

kangmingtay:
* Fixes #881 
* https://github.com/supabase/auth-js/pull/876 introduced a breaking change in `getUser()` which resulted in the authorization header not being used

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -1174,7 +1174,8 @@ export default class GoTrueClient {
           throw error
         }
 
-        if (!data.session?.access_token) {
+        const hasAccessToken = data.session?.access_token || this.headers['Authorization']
+        if (!hasAccessToken) {
           // if there's no access token, the user can't be fetched
           return { data: { user: null }, error: new AuthSessionMissingError() }
         }


soedirgo (approved):

j4w8n (commented):
One thing to keep in mind is that the Authorization header will always exist since the anon key is there by default, right? So `hasAccessToken` will always be true; which makes this whole section of code useless, and you might as well go back to no checks.

One thing you could do when the client is created is check if the dev passed in an Authorization header. If so, set something like `this.hasCustomAuthHeader = true` - which would be intialized as `false` during client creation.

Then you could check for that on the right side of `||` in this getUser code.

hf:
@j4w8n Thanks for catching this! You're a ⭐

PR #890, opened 2024-04-25
fix: re-add `getUser` returns null if there is no session

hf:
Reverts supabase/auth-js#889


kangmingtay:
closing in view of #891

PR #891, opened 2024-04-25
fix: return error if missing session or missing custom auth header

kangmingtay:
* Adds a new property `hasCustomAuthorizationHeader` which will be set by supabase-js 
* `getUser` should return an `AuthSessionMissingError` if there is no session and no custom auth header is set

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -81,7 +81,8 @@ export type GoTrueClientOptions = {
    */
   lock?: LockFunc
   /**
-   * Set to "true" if there is a custom authorization header
+   * Set to "true" if there is a custom authorization header set globally.
+   * @experimental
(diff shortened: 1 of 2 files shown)


src/lib/types.ts
```diff
@@ -80,6 +80,10 @@ export type GoTrueClientOptions = {
    * @experimental
    */
   lock?: LockFunc
+  /**
+   * Set to "true" if there is a custom authorization header
```
hf, line 84:
```suggestion
   * Set to "true" if there is a custom authorization header set globally.
   * @experimental
```

So it's not an API people think it's theirs to use.

hf (approved):

J0 (approved):

PR #906, opened 2024-05-12
ci: add build step

hf:
Build was missing for some reason. 😱


J0 (approved):
It looks like `npm ci`, as a standardized command, primarily installs from `package-lock.json` and does not run build

Tests [do not run the build step either](https://github.com/supabase/auth-js/blob/master/package.json#L32)

Thanks for the catch!

PR #909, opened 2024-05-14
don't use session.user object in _getAuthenticatorAssuranceLevel()

kizivat:
One avoidable source of the `getUser()` warnings

Bug fix

[Can't get rid of getUser() warning](https://github.com/supabase/auth-js/issues/873#issuecomment-2081467385)

## What is the new behavior?

The warning is no longer logged on the line defining `verifiedFactors`

## Additional context

N/A

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -2545,8 +2545,19 @@ export default class GoTrueClient {
 
         let nextLevel: AuthenticatorAssuranceLevels | null = currentLevel
 
+        const {
+          data: { user },
+          error: userError,
+        } = await this._getUser()
+        if (userError) {
+          return { data: null, error: userError }
+        }
+        if (!user) {
+          return { data: { currentLevel, nextLevel, currentAuthenticationMethods: [] }, error: null }
+        }
+        
         const verifiedFactors =
-          session.user.factors?.filter((factor: Factor) => factor.status === 'verified') ?? []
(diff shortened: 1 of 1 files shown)


kizivat:
The force push was to adhere to conventional commits.

sleepdotexe:
If the solution for this issue is making an additional `getUser` check inside the method – which I'm assuming should make the method trustable on the server(?) – should this call be made at the beginning of the `getAuthenticatorAssuranceLevel()` method (ie. before we do checks on the `session`)? 

Otherwise, we might end up using less secure information from `_useSession` since this has the possibility to return early before it reaches the `_getUser` call.

Also to consider: what is the added performance overhead for this function by making an additional user call? Do we need to provide an option for a jwt to be passed as an argument to `getAuthenticatorAssuranceLevel()` since `getUser` can optionally take a jwt?

j4w8n:
You may know this, but making sure: keep in mind this change adds a network call to Supabase.

mandarini:
Hi @kizivat ! Thanks for the contribution and your patience.

This repository is deprecated and has moved to the new [Supabase JS monorepo](ttps://github.com/supabase/supabase-js). Since this PR has been inactive for over a year, I’m going to close it to keep the old repo tidy, before archiving.

Note: This old repository will be archived on October 10, 2025. Thank you again for your effort and understanding!

Issue #904, opened 2024-05-09
Current session lost when auth function call fails


bombillazo:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Currently, the client logic calls `_removeSession` to clear the session when doing an auth call. If the call fails, the existing session is never restored and is lost, effectively disconnecting the user session.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Create a supabase client and log in
2. Call an auth function with invalid parameters
3. Get the client session

## Expected behavior

The existing client session is kept in the event of a fail.

## System information

- Version of supabase-js: 2.64.2

bombillazo:
This was closed with #915

PR #915, opened 2024-06-05
fix: don't call removeSession prematurely 

### LINKED ISSUE #853: Functions that call `_removeSession` internally do not trigger `SIGNED_OUT` event when the function fails

kangmingtay:
* Replaces #854 
* Fixes #853, #904 
* We don't need to remove the existing session prematurely. This causes some issues when users want to implement some sort of switch-account functionality since the existing session will always be removed regardless of whether the signup / sign-in attempt succeeds. 
* It's safe to remove `_removeSession` since calling `_saveSession` multiple times will just replace the existing session

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -369,8 +369,6 @@ export default class GoTrueClient {
    */
   async signInAnonymously(credentials?: SignInAnonymouslyCredentials): Promise<AuthResponse> {
     try {
-      await this._removeSession()
-
       const res = await _request(this.fetch, 'POST', `${this.url}/signup`, {
         headers: this.headers,
         body: {
@@ -414,8 +412,6 @@ export default class GoTrueClient {
    */
   async signUp(credentials: SignUpWithPasswordCredentials): Promise<AuthResponse> {
     try {
-      await this._removeSession()
-
       let res: AuthResponse
       if ('email' in credentials) {
         const { email, password, options } = credentials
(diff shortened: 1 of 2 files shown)


hf (approved):
I think generally fine, though I think we need to emit the `SIGNED_OUT` event in this case (if there was a session previously) and right after `SIGNED_IN`.

But for now, let's just document this here: https://supabase.com/docs/reference/javascript/auth-onauthstatechange

J0 (approved):

PR #917, opened 2024-06-09
Allow code exchange without sign in

arobert93:

Feature

Currently any pkce code exchange call automatically signs the user in.

## What is the new behavior?

This behaviour can be changed by passing the `signIn` argument as `false`.


J0:
Hey @arobert93,

Thanks for the PR - could we trouble you to expand a little more on the use case?  Specifically wondering when you'd want to make a code exchange but not sign the user in.

Let us know

hf:
This is unlikely to be generally useful. The code exchange is done so the user is signed in.

Please see if you can find another workaround for your use case.

Issue #921, opened 2024-06-16
NPM version tagged as latest still 2.62.2


BjoernRave:
On npm the version marked as latest is still 2.62.2 even though according to github there is already a 2.64.2 version.

I want to the use the `signInAnonmymously` fn, but cant. 

For some reason I get `No Access-Control-Allow-origin` when the lib is trying to get the `/user`, no matter if anonymous or with magic link. This is working fine in 2.62.2.

I was not sure if maybe there is a connection between these two.

Also there is `@supabase/auth-js` and `@supabase/gotrue-js` why is that? Is there a difference?

J0:
Hey

Thanks for flagging this - can you try installing: `npm install @supabase/auth-js@2.64.2` for now? We'll look into the latest tag.

> Also there is @supabase/auth-js and @supabase/gotrue-js why is that? Is there a difference?

Supabase renamed GoTrue to Auth so as to clear up the name

PR #922, opened 2024-06-17
fix: patch release workflow

J0:

Fixes: #921  so we add the latest tag on npm.

Currently we [publish the package](https://www.npmjs.com/package/@supabase/auth-js?activeTab=versions) but the tag doesn't seem to be appended correctly


hf:
Nice catch!

hf (approved):

PR #925, opened 2024-06-17
fix: don't refresh session if autoRefreshToken setting is set to false

thomasconner:

Bug fix - Supabase client will ignore `autoRefreshToken` setting when setting a session with an expired access token
Breaking Change - `Session` type now has `refresh_token` as optional

When a Supabase client is initialized with `{ auth: { autoRefreshToken: false } }`

```
const client = createClient(options.supabaseUrl, options.supabaseKey, {
    auth: { autoRefreshToken: false }
});
```
but a session is set with `client.setSession()` and the access token is expired the client will disregard the `autoRefreshToken` setting and attempt to refresh the session.

## What is the new behavior?

The client will not attempt to refresh an expired session when it is set with `client.setSession()` and  `autoRefreshToken` is set to `false`.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -1111,7 +1111,16 @@ export default class GoTrueClient {
         return { data: { session: currentSession }, error: null }
       }
 
+      if (!currentSession.refresh_token) {
+        this._debug(
+          '#__loadSession()',
+          'session has no refresh token to refresh session'
+        )
+        return { data: { session: null }, error: null };
+      }
+
       const { session, error } = await this._callRefreshToken(currentSession.refresh_token)
+
       if (error) {
         return { data: { session: null }, error }
       }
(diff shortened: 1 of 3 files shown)


j4w8n:
What's your use case here, and what issue are you running into?

thomasconner:
I am using the Supabase JavaScript client in a NestJS service. This service retrieves the access token from the incoming request to set the session. This setting of the session should not refresh the token automatically on the server side but instead just return an `UnauthorizedError` by the service to the requesting application. It is up to the requesting application to then properly refresh the access token.

The issue I was running into is that when setting the session with an expired access token, Supabase would go ahead and refresh the session ignoring the `autoRefreshToken` setting. By doing this, the requesting application now has an invalid refresh token since the new refresh token is not communicated back to the application since this was not the expected behavior.

thomasconner:
I removed the change that makes the `refresh_token` optional on a session. This minimizes the changes but still achieves the behavior I  would expect.

j4w8n:
After setting the session, do you use methods like `getSession()` and `getUser()`?

If not, then this PR would be more applicable once it's released: https://github.com/supabase/supabase-js/pull/1004

thomasconner:
In my opinion, this PR and my PR solve/fix two different things. [supabase/supabase-js#1004](https://github.com/supabase/supabase-js/pull/1004) would solves the problem of providing the access token to the Supabase client to make requests. However, my PR I believe still is fixing a bug where the `autoRefreshToken` setting is not obeyed when calling `supabase.setSession()`. 

I would end up using the new API on the server side as described in that PR you linked to. That is a nice solution.

thomasconner:
Do you have anymore feedback regarding this PR?

j4w8n:
> Do you have anymore feedback regarding this PR?

Moving this PR forward isn't up to me, as I don't have approval permissions. I know the auth team can get pretty busy, so I'm not sure if/when they will comment.

I think there could be some more things to think about with this, because the _loadSession method also checks for token expiration.

PR #932, opened 2024-07-21
feat: add bindings for Multi-Factor Authentication (Phone)

J0:

Adds the bindings for MFA (Phone) Client library. In particular, add
- Enroll - now allows for enrollment of Phone Factors
- Challenge for Phone - now takes in a channel
- List Factors - now supports access of Phone factors

### TODOS:
- [x] Remove empty `totp` from response field  on enroll response

### Not addressed yet

- `challengeAndVerify` will currently return an error on the `verify` step when used with a phone factor. We could introduce an alternate behaviour (e.g. terminating at the `challenge` step gracefully) since we have the `type` of the challenge-associated-factor in the response. This is not addressed in this PR

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -2341,25 +2341,38 @@ export default class GoTrueClient {
           return { data: null, error: sessionError }
         }
 
-        const { data, error } = await _request(this.fetch, 'POST', `${this.url}/factors`, {
-          body: {
-            friendly_name: params.friendlyName,
-            factor_type: params.factorType,
-            issuer: params.issuer,
-          },
-          headers: this.headers,
-          jwt: sessionData?.session?.access_token,
-        })
-
-        if (error) {
-          return { data: null, error }
+        if (params.factorType === 'sms') {
+          const { data, error } = await _request(this.fetch, 'POST', `${this.url}/factors`, {
+            body: {
+              friendly_name: params.friendlyName,
+              factor_type: params.factorType,
+              phone_number: params.phoneNumber,
+            },
+            headers: this.headers,
+            jwt: sessionData?.session?.access_token,
+          })
+          if (error) {
+            return { data: null, error }
+          }
+          return { data, error: null }
+        } else {
+          const { data, error } = await _request(this.fetch, 'POST', `${this.url}/factors`, {
+            body: {
+              friendly_name: params.friendlyName,
+              factor_type: params.factorType,
+              issuer: params.issuer,
+            },
+            headers: this.headers,
+            jwt: sessionData?.session?.access_token,
+          })
+          if (error) {
+            return { data: null, error }
+          }
+          if (data?.totp?.qr_code) {
+            data.totp.qr_code = `data:image/svg+xml;utf-8,${data.totp.qr_code}`
+          }
+          return { data, error: null }
         }
-
-        if (data?.totp?.qr_code) {
-          data.totp.qr_code = `data:image/svg+xml;utf-8,${data.totp.qr_code}`
-        }
-
-        return { data, error: null }
       })
     } catch (error) {
       if (isAuthError(error)) {
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -2442,6 +2442,7 @@ export default class GoTrueClient {
             'POST',
             `${this.url}/factors/${params.factorId}/challenge`,
             {
+              body: { channel: params.channel },
               headers: this.headers,
               jwt: sessionData?.session?.access_token,
             }
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -2341,7 +2341,7 @@ export default class GoTrueClient {
           return { data: null, error: sessionError }
         }
 
-        if (params.factorType === 'sms') {
+        if (params.factorType === 'phone') {
           const { data, error } = await _request(this.fetch, 'POST', `${this.url}/factors`, {
             body: {
               friendly_name: params.friendlyName,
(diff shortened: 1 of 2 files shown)


src/lib/types.ts
```diff
@@ -897,6 +910,22 @@ export type AuthMFAEnrollResponse =
...
+        /** ID of the factor that was just enrolled (in an unverified state). */
+        id: string
+
+        /** Type of MFA factor. Only `totp` supported for now. */
+        type: 'sms'
+
+        /** Friendly name of the factor, useful for distinguishing between factors **/
+        friendly_name?: string
+
+        /** Phone number of the MFA factor. Used to send SMS-es  */
```
J0, line 924:
```suggestion
        /** Phone number of the MFA factor. Used to send messages  */
```

src/GoTrueClient.ts
```diff
@@ -2341,25 +2341,38 @@ export default class GoTrueClient {
...
-          jwt: sessionData?.session?.access_token,
-        })
-
-        if (error) {
-          return { data: null, error }
-        }
-
-        if (data?.totp?.qr_code) {
-          data.totp.qr_code = `data:image/svg+xml;utf-8,${data.totp.qr_code}`
+        if (params.factorType === 'sms') {
```
J0, line 2344:
```suggestion
        if (params.factorType === 'phone') {
```

src/lib/types.ts
```diff
@@ -805,18 +805,29 @@ export type GenerateLinkType =
...
+      /** The type of factor being enrolled. */
+      factorType: 'totp'
+      /** Domain which the user is enrolled with. */
+      issuer?: string
+      /** Human readable name assigned to the factor. */
+      friendlyName?: string
+    }
+  | {
+      /** The type of factor being enrolled. */
+      factorType: 'sms'
```
J0, line 819:
```suggestion
      factorType: 'phone'
```

src/lib/types.ts
```diff
@@ -805,18 +805,29 @@ export type GenerateLinkType =
...
+      /** Human readable name assigned to the factor. */
+      friendlyName?: string
+      /** Phone number associated with a factor */
+      phoneNumber: string
+    }
 
 export type MFAUnenrollParams = {
   /** ID of the factor being unenrolled. */
   factorId: string
+  /** Phone number of the SMS Factor being enrolled */
```
J0, line 829:
```suggestion
```
Not needed, factorId should be sufficient

src/lib/types.ts
```diff
@@ -833,6 +844,8 @@ export type MFAVerifyParams = {
 export type MFAChallengeParams = {
   /** ID of the factor to be challenged. Returned in enroll(). */
   factorId: string
+  /** Messaging channel to use (e.g. whatsapp or sms). Only relevant for phone factors */
+  channel?: 'sms' | 'whatsapp'
 }
 
 export type MFAChallengeAndVerifyParams = {
```
J0, line 851: We should block `ChallengeAndVerify` with phone factors since it probably wouldn't make sense.

src/GoTrueClient.ts
```diff
@@ -2341,25 +2341,38 @@ export default class GoTrueClient {
...
+            },
+            headers: this.headers,
+            jwt: sessionData?.session?.access_token,
+          })
+          if (error) {
+            return { data: null, error }
+          }
+          if (data?.totp?.qr_code) {
+            data.totp.qr_code = `data:image/svg+xml;utf-8,${data.totp.qr_code}`
+          }
```
hf, line 2373: Isn't it easier to have one `_request(this.fetch, ...` call with a single body that's conditionally created above?

src/lib/types.ts
```diff
@@ -805,14 +801,23 @@ export type GenerateLinkType =
...
+      issuer?: string
+      /** Human readable name assigned to the factor. */
+      friendlyName?: string
+    }
+  | {
+      /** The type of factor being enrolled. */
+      factorType: 'phone'
+      /** Human readable name assigned to the factor. */
+      friendlyName?: string
+      /** Phone number associated with a factor */
```
hf, line 818: Specify it should be in E.164 in the comment.

hf (approved):

src/GoTrueClient.ts
J0: Yup that was considered but I figured this would better show that the changes are standalone until we lift the feature flag

J0: Updated

src/GoTrueClient.ts
```diff
@@ -2355,7 +2357,12 @@ export default class GoTrueClient {
           return { data: null, error }
         }
 
-        if (data?.totp?.qr_code) {
+        // TODO: Remove once: https://github.com/supabase/auth/pull/1717 is deployed
```
J0, line 2360: Made a mistake here - it still shows up in the response field. With the above change it should no longer show up in the response field for phone responses - till then we can remove the field for phone

It's unlikely that anyone will use (and depend) on the feature via the API given how new the feature is.

src/lib/types.ts
```diff
@@ -897,6 +903,22 @@ export type AuthMFAEnrollResponse =
...
+        id: string
+
+        /** Type of MFA factor. */
+        type: 'phone'
+
+        /** Friendly name of the factor, useful for distinguishing between factors **/
+        friendly_name?: string
+
+        /** Phone number of the MFA factor. Used to send messages  */
+        phone_number: string
```
J0, line 918:
```suggestion
       phone: string
```

src/lib/types.ts
```diff
@@ -897,6 +903,22 @@ export type AuthMFAEnrollResponse =
...
+        /** ID of the factor that was just enrolled (in an unverified state). */
+        id: string
+
+        /** Type of MFA factor. */
+        type: 'phone'
+
+        /** Friendly name of the factor, useful for distinguishing between factors **/
+        friendly_name?: string
+
+        /** Phone number of the MFA factor. Used to send messages  */
```
J0, line 917:
```suggestion
        /** Phone number of the MFA factor in E.164 format. Used to send messages  */
```

hf (approved):
Makes sense, please update the PR description to use human-readable text and clean up based on your TODO comment.

PR #945, opened 2024-08-26
fix: handle error exchange code for session

kangmingtay:
## What kind of change does this PR introduce?
* Fixes https://github.com/supabase/auth-js/issues/782

## What is the current behavior?

Please link any relevant issues here.

## What is the new behavior?

Feel free to include screenshots if it includes visual changes.

## Additional context

Add any other context or screenshots.

PR #946, opened 2024-08-26
fix: don't throw error in exchangeCodeForSession

### LINKED ISSUE #782: `exchangeCodeForSession` throws error instead of returning
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com).

## Describe the bug

When users log in using PKCE and click the email login link in a different browser, you get this `AuthApiError`:
> invalid request: both auth code and code verifier should be non-empty`

I believe that's expected, but the error is thrown instead of returned from `exchangeCodeForSession()`.

## To Reproduce

1. Set up PKCE with an auth handler like this:
	```ts
	const code = event.url.searchParams.get('code')
	if (typeof code === 'string') {
		const { data, error } = await supabase.auth.exchangeCodeForSession(code).catch((e) => {
			console.log('Thrown error', e);
		});
	}
	```
2. 

## Expected behavior

Since it's a expected and fairly common error, this error should be returned so that it can be properly handled

## System information

- OS: macOS
- Browser: Brave
- Version of supabase-js: 2.33.1
- Version of Node.js: 18.17.1

kangmingtay:
* Fixes #782 
* Supercedes #945 

## What is the new behavior?

## Additional context

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -580,33 +580,43 @@ export default class GoTrueClient {
   > {
     const storageItem = await getItemAsync(this.storage, `${this.storageKey}-code-verifier`)
     const [codeVerifier, redirectType] = ((storageItem ?? '') as string).split('/')
-    const { data, error } = await _request(
-      this.fetch,
-      'POST',
-      `${this.url}/token?grant_type=pkce`,
-      {
-        headers: this.headers,
-        body: {
-          auth_code: authCode,
-          code_verifier: codeVerifier,
-        },
-        xform: _sessionResponse,
+
+    try {
+      const { data, error } = await _request(
+        this.fetch,
+        'POST',
+        `${this.url}/token?grant_type=pkce`,
+        {
+          headers: this.headers,
+          body: {
+            auth_code: authCode,
+            code_verifier: codeVerifier,
+          },
+          xform: _sessionResponse,
+        }
+      )
+      await removeItemAsync(this.storage, `${this.storageKey}-code-verifier`)
+      if (error) {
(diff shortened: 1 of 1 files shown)


src/GoTrueClient.ts
```diff
@@ -580,33 +580,43 @@ export default class GoTrueClient {
...
+          body: {
+            auth_code: authCode,
+            code_verifier: codeVerifier,
+          },
+          xform: _sessionResponse,
+        }
+      )
+      await removeItemAsync(this.storage, `${this.storageKey}-code-verifier`)
+      if (error) {
+        throw error
```
J0, line 600: Should we continue to return the error here?  Seems like we return the error in the [original version](https://github.com/supabase/auth-js/pull/946/files#diff-3522461172efd6058d6b8da62fc2d30d8b524d2b64894ea2c67218c52f7fdff5L598) so wanted to confirm it is intended

`{ data: { user: null, session: null, redirectType: null }, error }`

Let me know if there are considerations I've missed though

J0 (approved):
Thanks for picking this up! Should help with and potentially close a few issues

src/GoTrueClient.ts
kangmingtay: @J0 it will get caught in the catch block below and returned

J0: Perfect, thanks for the clarification

PR #956, opened 2024-09-17
fix: improve `mfa.enroll` return types

aloisklink:

Bug fix for the TypeScript types.

- Fixes https://github.com/supabase/auth-js/issues/955

## What is the new behavior?

Use [TypeScript's function overloading](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads) to return

- TOTP data when `mfa.enroll` was passed `factorType: 'totp'`,
- phone data when `mfa.enroll` was passed `factorType: 'phone'`.

This fixes a breaking change in the types in b957c30782065e4cc421a526c62c101d35c443d4.

## Additional context

This PR would also expose the new following types:

- `MFAEnrollTOTPParams`
- `MFAEnrollPhoneParams`
- `AuthMFAEnrollTOTPResponse`
- `AuthMFAEnrollPhoneResponse`
- `AuthMFAEnrollErrorResponse` (it might be better to roll this type in with the other two).

It might be better to keep these types private, since once they're exposed, you can't easily remove them again without creating a breaking change.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -88,6 +88,11 @@ import type {
   LockFunc,
   UserIdentity,
   SignInAnonymouslyCredentials,
+  MFAEnrollTOTPParams,
+  AuthMFAEnrollTOTPResponse,
+  AuthMFAEnrollErrorResponse,
+  MFAEnrollPhoneParams,
+  AuthMFAEnrollPhoneResponse,
 } from './lib/types'
 
 polyfillGlobalThis() // Make "globalThis" available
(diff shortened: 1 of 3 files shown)


J0:
>It might be better to keep these types private, since once they're exposed, you can't easily remove them again without creating a breaking change.

Would agree with this

Thanks for the PR! Looks great so far

J0 (approved):

J0:
Will follow up with the corresponding PR for `challengeParams` and `verifyParams` (if they are at all needed)

aloisklink:
> > It might be better to keep these types private, since once they're exposed, you can't easily remove them again without creating a breaking change.
>
> Would agree with this

If you are still interested in this, I think the easiest way to do this would be to make a new file called something like `src/lib/internal-types.ts` that **isn't** exported outside of the package using `export * from './lib/types'`.

Then, `src/GoTrueClient.ts` and `src/lib/types.ts` could both import types from this file.

(the hardest thing would be picking the right conventional commit prefix :laughing:, since this PR has already been merged, but not released, so it's not yet a breaking change)

> Will follow up with the corresponding PR for `challengeParams` and `verifyParams` (if they are at all needed)
>
> Upd: shouldn't be needed, except in relation to MFA (WebAuthn).

Yep, we use both `supabase.auth.mfa.challenge` and `supabase.auth.mfa.verify` in our code (based of the example in https://github.com/supabase/supabase/blob/10ba4999cd33f796e89c50a0821e46b3d346eb7c/apps/docs/content/guides/auth/auth-mfa/totp.mdx?plain=1#L245-L261), and v2.65.0 works fine in those cases!

And I'm excited to see WebAuthn support!

J0:
@aloisklink, thanks for the update.

 I think an internal-types file would work. Let me do that shortly. Thanks!

PR #960, opened 2024-09-19
fix: add MFA for WebAuthn bindings

J0:

We introduce bindings for MFA (WebAuthn) which consists of primary methods and helper methods. 

### Core Methods

The core methods have single step and multi-step use.  A single step enroll handles the browser interactions for you while the multi-step enroll provides one the freedom to configure options as they wish.

For enrollment single step looks like:

```
await supabase.auth. enroll({factorType: 'webauthn'})
```

While mutli-step enroll contains a webAuthn parameter than can be customized

```
enroll(factorType: 'webauthn, webAuthn{...}`)
```

Verify operates in a similar fashion.

 There are two methods, `challenge` and `challengeAndVerify` which are affected.

### Helper methods

We provide the following helper methods for ease of use:

1. `browserSupportsWebAuthn` - most modern browsers should support the WebAuthn specification.
2. `startRegistration` - this method takes reference from SimpleWebAuthn and wraps `navigator.create`. Use this to customize how a credential is created
3. `startAuthentication` - this methods takes reference from SimpleWebAuthn and wraps `navigator.get`. Use this to customize how a credential is retrieved

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -36,6 +36,10 @@ import {
   supportsLocalStorage,
   parseParametersFromURL,
   getCodeChallengeAndMethod,
+  base64URLStringToBuffer,
+  bufferToBase64URLString,
+  startRegistration,
+  startAuthentication,
 } from './lib/helpers'
 import { localStorageAdapter, memoryLocalStorageAdapter } from './lib/local-storage'
 import { polyfillGlobalThis } from './lib/polyfills'
@@ -71,12 +75,14 @@ import type {
   VerifyOtpParams,
   GoTrueMFAApi,
   MFAEnrollParams,
-  AuthMFAEnrollResponse,
+  MFAVerifyParams,
   MFAChallengeParams,
   AuthMFAChallengeResponse,
   MFAUnenrollParams,
   AuthMFAUnenrollResponse,
-  MFAVerifyParams,
+  MFAVerifyTOTPParams,
+  MFAVerifyPhoneParams,
+  MFAVerifyWebAuthnParams,
   AuthMFAVerifyResponse,
   AuthMFAListFactorsResponse,
   AMREntry,
@@ -89,12 +95,19 @@ import type {
   LockFunc,
   UserIdentity,
   SignInAnonymouslyCredentials,
+  AuthenticatorTransportFuture,
+  PublicKeyCredentialCreationOptionsJSON,
+  RegistrationResponseJSON,
+  MFAVerifySingleStepWebAuthnParams,
+  AuthMFAEnrollResponse,
 } from './lib/types'
 import {
   MFAEnrollTOTPParams,
   MFAEnrollPhoneParams,
+  MFAEnrollWebAuthnParams,
   AuthMFAEnrollTOTPResponse,
   AuthMFAEnrollPhoneResponse,
+  AuthMFAEnrollWebAuthnResponse,
 } from './lib/internal-types'
 
 polyfillGlobalThis() // Make "globalThis" available
@@ -2386,6 +2399,7 @@ export default class GoTrueClient {
    */
   private async _enroll(params: MFAEnrollTOTPParams): Promise<AuthMFAEnrollTOTPResponse>
   private async _enroll(params: MFAEnrollPhoneParams): Promise<AuthMFAEnrollPhoneResponse>
+  private async _enroll(params: MFAEnrollWebAuthnParams): Promise<AuthMFAEnrollWebAuthnResponse>
   private async _enroll(params: MFAEnrollParams): Promise<AuthMFAEnrollResponse> {
     try {
       return await this._useSession(async (result) => {
@@ -2409,9 +2423,35 @@ export default class GoTrueClient {
         if (error) {
           return { data: null, error }
         }
-
         if (params.factorType === 'totp' && data?.totp?.qr_code) {
           data.totp.qr_code = `data:image/svg+xml;utf-8,${data.totp.qr_code}`
+        }
+        if (params.factorType === 'webauthn' && data.type === 'webauthn') {
+          if (params.useMultiStep) {
+            return { data, error: null }
+          }
+          const factorId = data.id
+          const webAuthn = this._getWebAuthnRpDetails()
+          const { data: challengeData, error: challengeError } = await this._challenge({
+            factorId,
+            webAuthn,
+          })
+          if (challengeError) {
+            return { data: null, error: challengeError }
+          }
+
+          if (!(challengeData.type === 'webauthn' && challengeData?.credential_creation_options)) {
+            return { data: null, error: new Error('Invalid challenge data for WebAuthn') }
+          }
+          let challengeOptions = challengeData?.credential_creation_options.publicKey
+          let credential = await startRegistration(challengeOptions)
+          const verifyWebAuthnParams = { ...webAuthn, creationResponse: credential }
+
+          return await this._verify({
+            factorId,
+            challengeId: challengeData.id,
+            webAuthn: verifyWebAuthnParams,
+          })
         }
 
         return { data, error: null }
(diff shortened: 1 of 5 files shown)


src/lib/types.ts
```diff
@@ -918,13 +945,30 @@ export type AuthMFAEnrollPhoneResponse = {
...
   error: null
 }
+
+export type AuthMFAEnrollWebAuthnResponse = {
+  data: {
+    /** ID of the factor that was just enrolled (in an unverified state). */
+    id: string
+
+    /** Type of MFA factor. */
+    type: 'phone'
```
J0, line 955:
```suggestion
    type: 'webauthn'
```

src/GoTrueClient.ts
```diff
@@ -2385,9 +2405,133 @@ export default class GoTrueClient {
...
+                ...challengeOptions.user,
+                id: base64URLStringToBuffer(challengeOptions.user.id),
+              },
+              excludeCredentials: challengeOptions.excludeCredentials?.map(
+                toPublicKeyCredentialDescriptor
+              ),
+            }
+            const options: CredentialCreationOptions = { publicKey }
+
+            const credential = (await navigator.credentials.create(
```
J0, line 2459: wrap and catch for identity error

src/lib/types.ts
```diff
@@ -951,6 +1017,21 @@ export type AuthMFAChallengeResponse =
...
+        /** Timestamp in UNIX seconds when this challenge will no longer be usable. */
+        expires_at: number
+        credential_creation_options: { publicKey: PublicKeyCredentialCreationOptionsJSON }
+        credential_request_options: { publicKey: PublicKeyCredentialRequestOptionsJSON }
+      }
+      error: null
+    }
   | { data: null; error: AuthError }
 
 export type AuthMFAListFactorsResponse =
```
J0, line 1037: Decide whether to include aaguid to image / authenticator type mapping for user: https://github.com/passkeydeveloper/passkey-authenticator-aaguids

or to include it in documentation

src/lib/helpers.ts
```diff
@@ -344,3 +348,105 @@ export function parseResponseAPIVersion(response: Response) {
...
+// Taken from simplewebauthn
+
+/**
+ * Convert from a Base64URL-encoded string to an Array Buffer. Best used when converting a
+ * credential ID from a JSON string to an ArrayBuffer, like in allowCredentials or
+ * excludeCredentials
+ *
+ * Helper method to compliment `bufferToBase64URLString`
+ */
+export function base64URLStringToBuffer(base64URLString: string): ArrayBuffer {
```
J0, line 361: Can switch to base64URL-js as well

src/lib/types.ts
```diff
@@ -1197,3 +1282,275 @@ export type SignOut = {
...
+    code,
+    cause,
+    name,
+  }: {
+    message: string
+    code: WebAuthnErrorCode
+    cause: Error
+    name?: string
+  }) {
+    // @ts-ignore: help Rollup understand that `cause` is okay to set
```
J0, line 1538: The build fails because of this ts-ignore comment in this helper function, which is ported over from simple-webauthn.

src/GoTrueClient.ts
```diff
@@ -2442,6 +2474,150 @@ export default class GoTrueClient {
...
+          },
+          headers: this.headers,
+          jwt: sessionData?.session?.access_token,
+        }
+      )
+
+      if (error) return { data: null, error }
+
+      await this._saveSession({
+        expires_at: Math.round(Date.now() / 1000) + data.expires_in,
```
hf, line 2581: Let's move this in `_saveSession`.

src/lib/helpers.ts
hf: Yes please let's use our library as that will come in handy for asymmetric JWTs too.

src/lib/helpers.ts
```diff
@@ -344,3 +356,551 @@ export function parseResponseAPIVersion(response: Response) {
...
+/**
+ * Visibly warn when we detect an issue related to a passkey provider intercepting WebAuthn API
+ * calls
+ */
+function warnOnBrokenImplementation(methodName: string, cause: Error): void {
+  console.warn(
+    `The browser extension that intercepted this WebAuthn API call incorrectly implemented ${methodName}. You should report this error to them.\n`,
+    cause
+  )
+}
```
hf, line 458: How does this apply to MFA WebAuthn -- we're not using passkeys as we're asking for the `cross-platform` authenticators.

src/lib/helpers.ts
```diff
@@ -344,3 +356,551 @@ export function parseResponseAPIVersion(response: Response) {
...
+    type,
+    clientExtensionResults: credential.getClientExtensionResults(),
+    // authenticatorAttachment: toAuthenticatorAttachment(credential.authenticatorAttachment),
+  }
+}
+
+/**
+ * Attempt to intuit _why_ an error was raised after calling `navigator.credentials.create()`
+ */
+export function identifyRegistrationError({
```
hf, line 605: Are these lifted directly from simplewebauthn? I'd prefer if we didn't do any additional error handling other than what the browser returns. Most of these are not relevant for the MFA use case anyway.

src/lib/helpers.ts
```diff
@@ -344,3 +356,551 @@ export function parseResponseAPIVersion(response: Response) {
...
+    if (options.signal instanceof AbortSignal) {
+      // https://www.w3.org/TR/webauthn-2/#sctn-createCredential (Step 16)
+      return new WebAuthnError({
+        message: 'Registration ceremony was sent an abort signal',
+        code: 'ERROR_CEREMONY_ABORTED',
+        cause: error,
+      })
+    }
+  } else if (error.name === 'ConstraintError') {
+    if (publicKey.authenticatorSelection?.requireResidentKey === true) {
```
hf, line 628: This does not apply for MFA.

src/lib/helpers.ts
```diff
@@ -344,3 +356,551 @@ export function parseResponseAPIVersion(response: Response) {
...
+      // https://www.w3.org/TR/webauthn-2/#sctn-op-make-cred (Step 4)
+      return new WebAuthnError({
+        message:
+          'Discoverable credentials were required but no available authenticator supported it',
+        code: 'ERROR_AUTHENTICATOR_MISSING_DISCOVERABLE_CREDENTIAL_SUPPORT',
+        cause: error,
+      })
+    } else if (publicKey.authenticatorSelection?.userVerification === 'required') {
+      // https://www.w3.org/TR/webauthn-2/#sctn-op-make-cred (Step 5)
```
hf, line 637: This error can be simply included in the enroll step.

src/lib/helpers.ts
```diff
@@ -344,3 +356,551 @@ export function parseResponseAPIVersion(response: Response) {
...
+      return new WebAuthnError({
+        message: 'User verification was required but no available authenticator supported it',
+        code: 'ERROR_AUTHENTICATOR_MISSING_USER_VERIFICATION_SUPPORT',
+        cause: error,
+      })
+    }
+  } else if (error.name === 'InvalidStateError') {
+    // https://www.w3.org/TR/webauthn-2/#sctn-createCredential (Step 20)
+    // https://www.w3.org/TR/webauthn-2/#sctn-op-make-cred (Step 3)
+    return new WebAuthnError({
```
hf, line 647: I also don't think this applies for MFA.

src/lib/helpers.ts
```diff
@@ -344,3 +356,551 @@ export function parseResponseAPIVersion(response: Response) {
...
+     * Pass the error directly through. Platforms are overloading this error beyond what the spec
+     * defines and we don't want to overwrite potentially useful error messages.
+     */
+    return new WebAuthnError({
+      message: error.message,
+      code: 'ERROR_PASSTHROUGH_SEE_CAUSE_PROPERTY',
+      cause: error,
+    })
+  } else if (error.name === 'NotSupportedError') {
+    const validPubKeyCredParams = publicKey.pubKeyCredParams.filter(
```
hf, line 663: This is unlikely to be ever thrown as all authenticators support P-256 and/or RSA.

src/lib/helpers.ts
```diff
@@ -344,3 +356,551 @@ export function parseResponseAPIVersion(response: Response) {
...
+    // https://www.w3.org/TR/webauthn-2/#sctn-op-make-cred (Step 2)
+    return new WebAuthnError({
+      message:
+        'No available authenticator supported any of the specified pubKeyCredParams algorithms',
+      code: 'ERROR_AUTHENTICATOR_NO_SUPPORTED_PUBKEYCREDPARAMS_ALG',
+      cause: error,
+    })
+  } else if (error.name === 'SecurityError') {
+    const effectiveDomain = window.location.hostname
+    if (!isValidDomain(effectiveDomain)) {
```
hf, line 685: Not possible to be thrown as the library chooses the origin for the developer.

src/lib/helpers.ts
```diff
@@ -344,3 +356,551 @@ export function parseResponseAPIVersion(response: Response) {
...
+      // https://www.w3.org/TR/webauthn-2/#sctn-createCredential (Step 7)
+      return new WebAuthnError({
+        message: `${window.location.hostname} is an invalid domain`,
+        code: 'ERROR_INVALID_DOMAIN',
+        cause: error,
+      })
+    } else if (publicKey.rp.id !== effectiveDomain) {
+      // https://www.w3.org/TR/webauthn-2/#sctn-createCredential (Step 8)
```
hf, line 693: Likewise...

src/lib/helpers.ts
```diff
@@ -344,3 +356,551 @@ export function parseResponseAPIVersion(response: Response) {
...
+      return new WebAuthnError({
+        message: `The RP ID "${publicKey.rp.id}" is invalid for this domain`,
+        code: 'ERROR_INVALID_RP_ID',
+        cause: error,
+      })
+    }
+  } else if (error.name === 'TypeError') {
+    if (publicKey.user.id.byteLength < 1 || publicKey.user.id.byteLength > 64) {
```
hf, line 701: Also impossible to be thrown given the server should always correctly set up a user ID.

src/lib/helpers.ts
J0: Yes directly lifted - we can remove yup

src/lib/helpers.ts
J0: Yup it doesn't

src/lib/helpers.ts
J0: Should implement the warning message. Think the checks might be still relevant though (e.g. in MFA Authenticators we might still need `getPublicKey` on L3)

Let me double check and udpate

src/GoTrueClient.ts
J0: I think we have a slightly different computation in the `_saveSession` calls elsewhere. Will try to re-concile though.

src/lib/types.ts
hf: Docs is better, we can add that in a follow up based on feedback.

src/lib/helpers.ts
hf: Let's please switch to this: https://github.com/supabase-community/base64url-js/blob/main/src/base64url.ts

The README has an example of using `Uint8Array` to convert to Base64-URL and v.v.

src/lib/helpers.ts
```diff
@@ -344,3 +355,338 @@ export function parseResponseAPIVersion(response: Response) {
...
+/**
+ * Begin authenticator "registration" via WebAuthn attestation
+ *
+ * @param optionsJSON Output from **@simplewebauthn/server**'s `generateRegistrationOptions()`
+ */
+export async function startRegistration(
+  optionsJSON: PublicKeyCredentialCreationOptionsJSON
+): Promise<RegistrationResponseJSON> {
+  if (!browserSupportsWebAuthn()) {
+    throw new Error('WebAuthn is not supported in this browser')
```
hf, line 513: Add additional context here... something like:

```suggestion
    throw new Error('WebAuthn is not supported in this browser. If using in SSR please use this in client components or pages.')
```

src/lib/types.ts
```diff
@@ -889,6 +931,21 @@ export type AuthMFAChallengeResponse =
...
+        /** ID of the newly created challenge. */
+        id: string
+
+        /** Factor Type which generated the challenge */
+        type: 'webauthn'
+
+        /** Timestamp in UNIX seconds when this challenge will no longer be usable. */
+        expires_at: number
+        credential_creation_options: { publicKey: PublicKeyCredentialCreationOptionsJSON }
+        credential_request_options: { publicKey: PublicKeyCredentialRequestOptionsJSON }
```
hf, line 945: These both should be optional, no?

src/lib/types.ts
```diff
@@ -1131,3 +1191,224 @@ export type SignOut = {
...
+  parseCreationOptionsFromJSON?(
+    options: PublicKeyCredentialCreationOptionsJSON
+  ): PublicKeyCredentialCreationOptions
+  // See https://w3c.github.io/webauthn/#sctn-parseRequestOptionsFromJSON
+  parseRequestOptionsFromJSON?(
+    options: PublicKeyCredentialRequestOptionsJSON
+  ): PublicKeyCredentialRequestOptions
+  // See https://w3c.github.io/webauthn/#dom-publickeycredential-tojson
+  toJSON?(): PublicKeyCredentialJSON
+}
```
hf, line 1414: There's just so much going on here, and doesn't feel very used.

Can we just simplify this... the TypeScript compiler already includes all of the types necessary. We don't need to redefine anything.

src/GoTrueClient.ts
J0: Yeah a bit harder to compute in `_saveSession` as there are places where we pass in the session with the `expires_at` field unedited [like here](https://github.com/supabase/auth-js/blob/master/src/GoTrueClient.ts#L1257) so we can't do the adjustment without affecting existing calls.

Let me know if there's a workaround missed though.

src/lib/helpers.ts
J0: The checks are still relevant, as some authenticators (probably a small amount) may not have `getPublicKeyAlgorithm()` , `getPublicKey()` etc. Changed `passkeys` to `keys` in the comment to make it more general

src/lib/types.ts
J0: The entire section is done as sugar around the TypeScript compiler fields. In the TypeScript version, certain fields are [BufferSource](https://github.com/MasterKale/SimpleWebAuthn/blob/5e3e5718f6b97ee3df09468a4400d3c7770a3f31/packages/types/src/dom.ts#L95) while in the sugar version suffixed with `-JSON` these fields are cast to `base64URL / string` which is slightly easier to work with.

I would also like to remove this if possible, will see if there's a clean workaround. In the meantime open to suggestions as well

src/lib/types.ts
```diff
@@ -824,11 +819,55 @@ export type MFAVerifyParams = {
...
+export type MFAVerifyWebAuthnParams =
+  | {
+      /** The type of factor being enrolled. */
+      factorType: 'webauthn'
+    }
+  | MFAVerifySingleStepWebAuthnParams
+
+export type MFAVerifySingleStepWebAuthnParams = {
+  factorId: string
+  challengeId: string
```
J0, line 834:
```suggestion
  challengeId?: string
```

PR #963, opened 2024-09-26
fix: remove phone mfa deletion, match on error codes

J0:

Addresses a few existing TODOs

- Remove the deletion of  TOTP object for backward compatibility
- Matches on identity linking error codes instead of messages

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -313,11 +313,8 @@ export default class GoTrueClient {
         if (error) {
           this._debug('#_initialize()', 'error detecting session from URL', error)
 
-          // hacky workaround to keep the existing session if there's an error returned from identity linking
-          // TODO: once error codes are ready, we should match against it instead of the message
           if (
-            error?.message === 'Identity is already linked' ||
-            error?.message === 'Identity is already linked to another user'
+            error?.code === identity_already_exists
           ) {
             return { error }
           }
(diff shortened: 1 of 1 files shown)


src/GoTrueClient.ts
```diff
@@ -313,11 +313,8 @@ export default class GoTrueClient {
         if (error) {
           this._debug('#_initialize()', 'error detecting session from URL', error)
 
-          // hacky workaround to keep the existing session if there's an error returned from identity linking
-          // TODO: once error codes are ready, we should match against it instead of the message
           if (
-            error?.message === 'Identity is already linked' ||
-            error?.message === 'Identity is already linked to another user'
+            error?.code === identity_already_exists
```
J0, line 317: https://github.com/supabase/auth/blob/ba00f75c28d6708ddf8ee151ce18f2d6193689ef/internal/api/identity.go#L118

src/GoTrueClient.ts
```diff
@@ -2388,11 +2385,6 @@ export default class GoTrueClient {
           return { data: null, error }
         }
 
-        // TODO: Remove once: https://github.com/supabase/auth/pull/1717 is deployed
-        if (params.factorType === 'phone') {
```
J0, line 2392: TOTP object should now be null since v2.158.1 is deployed  (we are now on v2.160.0)

To verify this go toSettings > Infrastructure

src/GoTrueClient.ts
J0, line 317:
```suggestion
            error?.code === 'identity_already_exists'
```

kangmingtay (approved):

PR #964, opened 2024-09-27
fix: move MFA sub types to internal file

J0:

Follow up to #956, hides the MFA sub-types (e.g. MFA Phone, MFA TOTP) needed to fix type errors before we release to public so that we can remove them as and when we like. The sub-types are now stored in an `internal-types` file which, unlike the `types` file is not exposed to public.

The types mentioned here have not been released yet so fine to hide them

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -88,12 +88,13 @@ import type {
   LockFunc,
   UserIdentity,
   SignInAnonymouslyCredentials,
-  MFAEnrollTOTPParams,
-  AuthMFAEnrollTOTPResponse,
-  AuthMFAEnrollErrorResponse,
-  MFAEnrollPhoneParams,
-  AuthMFAEnrollPhoneResponse,
 } from './lib/types'
+import {
+  MFAEnrollTOTPParams,
+  MFAEnrollPhoneParams,
+  AuthMFAEnrollTOTPResponse,
+  AuthMFAEnrollPhoneResponse,
+} from './lib/internal-types'
 
 polyfillGlobalThis() // Make "globalThis" available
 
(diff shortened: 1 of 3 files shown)


src/lib/internal-types.ts
```diff
@@ -0,0 +1,75 @@
...
+export type AuthMFAEnrollPhoneResponse =
+  | {
+      data: {
+        /** ID of the factor that was just enrolled (in an unverified state). */
+        id: string
+
+        /** Type of MFA factor. */
+        type: 'phone'
+
+        /** Friendly name of the factor, useful for distinguishing between factors **/
```
J0, line 64: We only export [the types file](https://github.com/supabase/auth-js/blob/master/src/index.ts#L6) so this shouldn't get exposed

aloisklink (approved):
I've tested this and it seems to work fine!

This is exactly what I was thinking of in https://github.com/supabase/auth-js/pull/956#issuecomment-2357652590 :rocket: 

Depending on your release/coding style, consider making this a `refactor:` commit/PR instead. It looks like you're automatically generating release notes from the commits, and so you might not want this to show up in the release notes, since it only "fixes" unreleased APIs!

(this is probably my main issue with conventional commits and automated changelogs, it's a bit difficult to mark a commit that changes unreleased behavior).

src/GoTrueClient.ts
```diff
@@ -313,9 +314,7 @@ export default class GoTrueClient {
         if (error) {
           this._debug('#_initialize()', 'error detecting session from URL', error)
 
-          if (
-            error?.code === 'identity_already_exists'
-          ) {
+          if (error?.code === 'identity_already_exists') {
```
J0, line 317: prettier

hf (approved):

PR #977, opened 2024-10-29
feat: add process lock for optional use in non-browser environments (React Native)

hf:
When using the library in non-browser environments like React Native or other JavaScript-based runtimes (Electron's main process for example) certain race conditions could still occur.

We've received some signal from customers using React Native that at scale these become more visible. This is why I'm introducing a `processLock` that developers can import like so:

```typescript
import { processLock } from '@supabase/auth-js/lib/locks'
```

And add to their apps by specifying the lock option with the process lock.


kangmingtay (approved):
we should update our react native guides to use this when creating the supabase client

PR #982, opened 2024-11-19
fix: `isBrowser()` to include check on `window`

salmoro:
There are several checks in `GoTrueClient.ts` that check `window?.[prop]` after checking `isBrowser()` which throws an error in node environments that have `document` defined but not `window`.

Fixes:

- https://github.com/supabase/supabase-js/issues/786

### DIFF
--- a/src/lib/helpers.ts
+++ b/src/lib/helpers.ts
@@ -14,7 +14,8 @@ export function uuid() {
   })
 }
 
-export const isBrowser = () => typeof document !== 'undefined'
+export const isBrowser = () =>
+  typeof window !== 'undefined' && typeof document !== 'undefined'
 
 const localStorageWriteTests = {
   tested: false,


hf:
What environment is this throwing in?

salmoro:
> What environment is this throwing in?

Sveltekit.
It's so bad that the only work around we found was to create a pnpm patch on auth-js package and patches are major pains to maintain.

j4w8n:
Just keep React Native, and possible other similar things, in mind here.

https://github.com/supabase/supabase-js/issues/786#issuecomment-1582717743

salmoro:
> Just keep React Native, and possible other similar things, in mind here.
> 
> [supabase/supabase-js#786 (comment)](https://github.com/supabase/supabase-js/issues/786#issuecomment-1582717743)

@j4w8n Yes. My changes here only add a requirement for `isBrowser()` to also check that `typeof window !== 'undefined'`. The previous check for `typeof document !== 'undefined'` is still there and will return `false` for React Native.

Issue #830, opened 2023-12-23
Angular application started polluting logs with exception.


glani:
# Bug report

## Describe the bug

The simple angular application started polluting logs. 

## To Reproduce

app.module.ts

in AppModule

```
export const initializer1 =
  (
  ): (() => Promise<boolean>) =>
    async () => {
      const supabase = createClient(
        'url',
        'key',
        {
          auth: {
            debug: true,
          }
        }
      );
      supabase.auth.onAuthStateChange(() => {});

      return true;
    };


{
      provide: APP_INITIALIZER,
      useFactory: initializer1,
      multi: true,
    },
```

## Expected behavior

No errors in Zone <root> or opportunity to customize showing exceptions.

## System information

- OS: macOS, centos
- Browser chrome
- Version of supabase-js: [^2.39.12]
- Version of Node.js: [e.g. 20.10.0]

## Additional context
Debug output in the browser console from GoTrue

GoTrueClient@0 (2.62.0) 2023-12-23T09:37:44.619Z #_acquireLock begin -1
render-document-frontend-impl.ts:80 GoTrueClient@0 (2.62.0) 2023-12-23T09:37:44.619Z #onAuthStateChange() registered callback with id 83bf91bd-ddb2-4add-803e-0e3b2cf2b78d
render-document-frontend-impl.ts:89 GoTrueClient@0 (2.62.0) 2023-12-23T09:37:44.619Z #onAuthStateChange() registered callback with id a347217b-4993-4ee7-af13-921531e6c12e
core.mjs:26656 Angular is running in development mode.
GoTrueClient.js:131 GoTrueClient@0 (2.62.0) 2023-12-23T09:37:44.643Z #_acquireLock lock acquired for storage key sb-guksenxnyizhahqwwusz-auth-token
GoTrueClient.js:131 GoTrueClient@0 (2.62.0) 2023-12-23T09:37:44.644Z #_initialize() begin is PKCE flow false
GoTrueClient.js:131 GoTrueClient@0 (2.62.0) 2023-12-23T09:37:44.644Z #_recoverAndRefresh() begin
GoTrueClient.js:131 GoTrueClient@0 (2.62.0) 2023-12-23T09:37:44.644Z #_recoverAndRefresh() session from storage null
GoTrueClient.js:131 GoTrueClient@0 (2.62.0) 2023-12-23T09:37:44.644Z #_recoverAndRefresh() session is not valid
[...]

PR #989, opened 2024-12-06
feat: wrap navigator.locks.request with plain promise to help zone.js

hf:
Wraps the `navigator.locks.request()` function with a plain Promise as libraries such as zone.js patch this object to track execution context. It appears that this browser API uses a native promise that's not patched, causing the tracking context to be lost.

It is believed that wrapping this non-zone.js Promise returned by the browser with a promise that's patched by zone.js can help the situation.

Related:
- https://github.com/supabase/supabase-js/issues/936
- #830


J0 (approved):
Thanks! Hopefully wrapping with a promise and adding additional logs will alleviate the issue

PR #992, opened 2024-12-12
fix: return error early for redirects

kangmingtay:
* When using the PKCE flow, if there's a redirect error after the callback and a code is not returned, `getSessionFromUrl` ends up returning `Not a valid PKCE flow url.` instead of the actual error


J0 (approved):

ChuckJonas:
@kangmingtay just wanted to make sure you are aware this change causes the JWT to be nuked from local storage every time the client is loaded (when there is no session in the URL).

[#1334](https://github.com/supabase/supabase-js/issues/1334)

PR #998, opened 2024-12-13
fix: revert reduce parsing query params unnecessarily

hf:
Reverts supabase/auth-js#993

Reports of a problem with the session found by users.

https://github.com/supabase/auth-js/issues/995


hf:
Closing in favor of #999

PR #1007, opened 2024-12-19
ci: publish with provenance

hf:
Will publish `auth-js` and `gotrue-js` with [provenance](https://docs.npmjs.com/generating-provenance-statements).

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   "main": "dist/main/index.js",
   "module": "dist/module/index.js",
   "types": "dist/module/index.d.ts",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@supabase/auth-js",
+  "name": "@supabase/gotrue-js",
```
hf, line 2: Ah not this change. This was the test. :D

staaldraad (approved):
🚀

J0 (approved):

PR #1022, opened 2025-01-14
Support throwing errors instead of returning them

### LINKED ISSUE #1020: Support throwing errors instead of returning them
# Feature request

## Describe the feature

Add a `throwOnError` client option to have errors be thrown instead of returning it. As an example, right now we do:

```ts
const auth = new AuthClient({ url: AUTH_URL })
const { data, error } = await auth.signUp({ email: 'myemail@example.com', password: 'foobar' })
```

But with `throwOnError` we do:

```ts
const auth = new AuthClient({ url: AUTH_URL, throwOnError: true })
const { data } = await auth.signUp({ email: 'myemail@example.com', password: 'foobar' })
```

## Additional context

Follows https://github.com/supabase/postgrest-js/pull/188 & https://github.com/supabase/postgrest-js/pull/248

kiwicopple:
Fixes #1020

Add support for `throwOnError` option in `AuthClient` to throw errors instead of returning them.

* **GoTrueClient Changes:**
  - Add `throwOnError` option to `GoTrueClient` constructor.
  - Modify methods to throw errors when `throwOnError` is true.
  - Update methods like `signUp`, `signInWithPassword`, `updateUser`, and others to throw errors if `throwOnError` is true.

* **AuthClient Changes:**
  - Update `AuthClient` to pass `throwOnError` option to `GoTrueClient`.

* **Tests:**
  - Add tests in `test/GoTrueClient.test.ts` to verify `throwOnError` functionality.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/supabase/auth-js/issues/1020?shareId=XXXX-XXXX-XXXX-XXXX).

### DIFF
--- a/src/AuthClient.ts
+++ b/src/AuthClient.ts
@@ -1,5 +1,10 @@
 import GoTrueClient from './GoTrueClient'
 
-const AuthClient = GoTrueClient
+const AuthClient = (options) => {
+  return new GoTrueClient({
+    ...options,
+    throwOnError: options.throwOnError || false,
+  })
+}
 
 export default AuthClient
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -107,6 +107,7 @@ const DEFAULT_OPTIONS: Omit<Required<GoTrueClientOptions>, 'fetch' | 'storage' |
   flowType: 'implicit',
   debug: false,
   hasCustomAuthorizationHeader: false,
+  throwOnError: false,
 }
 
 /** Current session will be checked for refresh at this interval. */
@@ -167,6 +168,7 @@ export default class GoTrueClient {
   protected lock: LockFunc
   protected lockAcquired = false
   protected pendingInLock: Promise<any>[] = []
+  protected throwOnError: boolean
 
   /**
    * Used to broadcast state change events to other tabs listening.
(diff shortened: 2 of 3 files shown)


src/AuthClient.ts
```diff
@@ -1,5 +1,10 @@
 import GoTrueClient from './GoTrueClient'
 
-const AuthClient = GoTrueClient
+const AuthClient = (options) => {
```
j4w8n, line 3: Pardon me for butting in here, but can you explain why these changes are being made for AuthClient? I'm not understanding.

kiwicopple: i'm trying to solve [this issue](https://github.com/supabase/auth-js/issues/1020) with GitHub Copilot

kiwicopple (commented):
remove silly AI suggestion

hf (commented):
Maybe a better way to deal with this is to have a function like so:

```typescript
private function _returnResult<D, E>(result: { data: D, error: E }): { data: D, error: E } {
  if (this.throwOnError && result.error) {
    throw result.error
  }

  return result
}
```

Then each time we return the object we call:

```typescript
return this._returnResult({ data: null, error })
```

It will be easier to spot and we could also potentially write a lint rule for it so it's not forgotten.

hf:
Don't worry about the check-conventional-commits errors, I'll fix those before merging.

mandarini:
moved: https://github.com/supabase/supabase-js/pull/1766

PR #1023, opened 2025-01-16
feat: introduce experimental split user and session storage

hf:
A common complaint when using Supabase in SSR is that the cookie size is huge. Some server configurations are not able to use such large cookies.

A major contributor to cookie size is that the user object is stored alongside the access and refresh tokens. This object _should not be used on the server_ but nevertheless has to exist to make this library happy.

This change introduces the ability for this library to store the user object in a separate storage location. For now it's experimental mode to be proofed before being widely adopted.

**How does it work?**

You can initialize the client by passing in a new option `userStorage` in addition to the already existing and optional `storage` option. By default `userStorage` is not set and a single storage is used for all elements of the session (including `user` property).

If `userStorage` is set, **all future changes to the session** will write the user there, and the rest of the session object to `storage`.

**Unsolvable Problems**

Say you set up the client like so:

```typescript
new GoTrueClient(URL, {
  // ...
  storage: cookieStorage,
  userStorage: window.localStorage,
})
```

On the server, the cookies -- obviously -- will not contain the `user` object. Because the `Session` type defines `user: User` as non-nullable, attempting to access a property on this object will throw an exception. Instead you should always call `getUser()` to fetch a trusted and fresh user object. This problem will be solved in v3 of this library.

**Testing**

[This PR](https://github.com/supabase/supabase/pull/32833) can be used to test this PR before merging. Merging should be safe as this is opt-in behavior for now.


src/GoTrueClient.ts
```diff
@@ -1952,7 +2017,13 @@ export default class GoTrueClient {
         // no need to persist currentSession again, as we just loaded it from
         // local storage; persisting it again may overwrite a value saved by
         // another client with access to the same local storage
-        await this._notifyAllSubscribers('SIGNED_IN', currentSession)
+
+        // Prevent sending a proxy user object as it can't be structured cloned
+        let finalPayloadSession =
+          currentSession.user && (currentSession.user as any).__isUserNotAvailableProxy
+            ? null
+            : currentSession
```
cemalkilic, line 2025: maybe we can `getUser()` to freshen the `user` if we detect the proxy?

PR #1024, opened 2025-01-17
feat: improved insecure `getSession()` warning

hf:
Previously when attempting to use `getSession()` from insecure storage such as cookies in SSR on the server, a disruptive warning was logged. 

It's been a long while and most developers should already be conditioned that _general_ use of `getSession()` for authorization on the server should not be used. There are, however, some valid use cases for `getSession()` that used to log a warning but now won't:

**Does (any) session exist or not?**

Most applications want to redirect the user to a login page if there's no session and probably are not interested in validating that the session is actually OK. This can be achieved with the following sample code:

```typescript
const supabase = createServerClient(/* ... */)

const { data: session, error } = await supabase.auth.getSession()

if (!session) {
  // no warning logged
  return NextResponse.redirect('/login')
}
```

**Simple boolean user check?**

Note that this does not answer *what user* is there.

```typescript
const supabase = createServerClient(/* ... */)

const { data: session, error } = await supabase.auth.getSession()

if (!session?.user) {
  // no warning logged
  return NextResponse.redirect('/login')
}

// but if you use
if (session?.user?.id === '284e7abb-785c-453c-bca4-a93da0825776') {
  // this will never be true because session.user.id === '@@ not-available @@'
  // and a warning will be logged
}

// same for app_metadata or user_metadata
```

**Just access an up-to-date access token to send to a private API?**

```typescript
const supabase = createServerClient(/* ... */)

const { data: session, error } = await supabase.auth.getSession()

await fetch(URL, {
  headers: {
    Authorization: `Bearer ${session.access_token}`
  },
})
```


src/GoTrueClient.ts
```diff
@@ -1120,21 +1121,10 @@ export default class GoTrueClient {
...
-                suppressWarning = true // keeps this proxy instance from logging additional warnings
-                this.suppressGetSessionWarning = true // keeps this client's future proxy instances from warning
-              }
-              return Reflect.get(target, prop, receiver)
-            },
-          })
-          currentSession = proxySession
+          currentSession.user = userNotAvailableProxy(
+            currentSession.user,
+            'User object comes from insecure storage and may not be authentic. Call getUser() instead to prevent security issues.'
```
kangmingtay, line 1125:
```suggestion
            'User object comes from insecure storage and cannot be trusted. Call getUser() instead to obtain a trusted user object.'
```

kangmingtay: reworded it abit to make the warning message less ambiguous

src/lib/helpers.ts
```diff
@@ -1,12 +1,26 @@
...
+    'crypto' in globalThis &&
+    typeof globalThis.crypto === 'object' &&
+    'randomUUID' in globalThis.crypto &&
+    typeof globalThis.crypto.randomUUID === 'function'
+  ) {
+    try {
+      return globalThis.crypto.randomUUID()
+    } catch (e: any) {
+      // ignore and fall back to below implementation, as crypto.randomUUID()
+      // only works in secure contexts
```
kangmingtay, line 20:
```suggestion
      // ignore and fall back to below implementation, as crypto.randomUUID()
      // only works in secure contexts (https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts)
```

kangmingtay: just to confirm, this change isn't related to the getSession stuff right? what is the benefit to try to use `crypto.randomUUID` over our implementation below?

kangmingtay (approved):

src/lib/helpers.ts
hf: Not related, just saw it. It's faster + does not use `Math.random()` which isn't really random.

mandarini:
Closing, as discussed.

PR #1030, opened 2025-01-29
feat: introduce getClaims method to verify asymmetric JWTs

kangmingtay:
* `getClaims` supports verifying JWTs (both asymmetric and symmetric) and returns the entire set of claims in the JWT payload

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -20,6 +20,8 @@ import {
   isAuthRetryableFetchError,
   isAuthSessionMissingError,
   isAuthImplicitGrantRedirectError,
+  CustomAuthError,
+  AuthInvalidJwtError,
 } from './lib/errors'
 import {
   Fetch,
@@ -43,7 +45,10 @@ import {
   supportsLocalStorage,
   parseParametersFromURL,
   getCodeChallengeAndMethod,
+  getAlgorithm,
+  validateExp,
 } from './lib/helpers'
+import { base64url } from './lib/rfc4648'
 import { localStorageAdapter, memoryLocalStorageAdapter } from './lib/local-storage'
 import { polyfillGlobalThis } from './lib/polyfills'
 import { version } from './lib/version'
@@ -2598,5 +2603,118 @@ export default class GoTrueClient {
         return { data: { currentLevel, nextLevel, currentAuthenticationMethods }, error: null }
       })
     })
+  }
+
+  /**
+   * Gets the claims from a JWT. If the JWT is symmetric JWTs, it will call getUser() to verify against the server.
+   * If the JWT is asymmetric, it will be verified against the JWKS using the WebCrypto API.
+   */
+  async getClaims(jwt?: string): Promise<
+    | {
+        data: { claims: { [key: string]: any } }
+        error: null
+      }
+    | { data: null; error: AuthError }
+    | { data: null; error: null }
+  > {
+    try {
+      let token = jwt
+      if (!token) {
+        const { data, error } = await this.getSession()
+        if (error || !data.session) {
+          return { data: null, error }
+        }
+        token = data.session.access_token
+      }
+
+      const parts = token.split('.')
+      if (parts.length !== 3) {
+        throw new AuthInvalidJwtError('Invalid JWT structure')
+      }
+
+      const [rawHeader, rawPayload, rawSignature] = parts
+      const decoder = new TextDecoder()
+      const header = JSON.parse(decoder.decode(base64url.parse(rawHeader, { loose: true })))
+      const payload = JSON.parse(decoder.decode(base64url.parse(rawPayload, { loose: true })))
+
+      // Reject expired JWTs
+      validateExp(payload.exp)
+
+      // If symmetric algorithm, fallback to getUser()
+      if (header.alg === 'HS256') {
+        const { error } = await this.getUser(token)
+        if (error) {
+          throw error
+        }
+        // getUser succeeds so the claims in the JWT can be trusted
+        return {
+          data: {
+            claims: this._decodeJWT(token),
+          },
+          error: null,
+        }
+      }
+
+      const { kid, alg } = header
+      if (!kid) {
+        throw new AuthInvalidJwtError('Missing kid claim')
+      }
+
+      const algorithm = getAlgorithm(alg)
+
+      // Fetch JWKS
+      const {
+        data: { keys },
+        error,
+      } = await _request(this.fetch, 'GET', `${this.url}/.well-known/jwks.json`, {
+        headers: this.headers,
+      })
+      if (error) {
+        throw error
+      }
+      if (!keys || keys.length === 0) {
+        throw new AuthInvalidJwtError('JWKS is empty')
+      }
+
+      // Find the signing key
+      const signingKey = keys.find((key: any) => key.kid === header.kid)
+      if (!signingKey) {
+        throw new AuthInvalidJwtError('No matching signing key found in JWKS')
+      }
+
+      // Convert JWK to CryptoKey
+      const publicKey = await crypto.subtle.importKey('jwk', signingKey, algorithm, true, [
+        'verify',
+      ])
+      // Convert rawSignature from base64url to binary array
+      const signature = base64url.parse(rawSignature, { loose: true })
+      // Verify the signature
+      const isValid = await crypto.subtle.verify(
+        algorithm,
+        publicKey,
+        signature,
+        new TextEncoder().encode(`${rawHeader}.${rawPayload}`)
+      )
+
+      if (!isValid) {
+        throw new AuthInvalidJwtError('Invalid JWT signature')
+      }
+
+      // If verification succeeds, decode and return claims
+      return {
+        data: {
+          claims: payload,
+        },
+        error: null,
+      }
+    } catch (error) {
+      if (isAuthError(error)) {
+        return { data: null, error }
+      }
+      return {
+        data: null,
+        error: new AuthUnknownError('Unknown error occurred while getting claims', error),
+      }
+    }
   }
 }
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -2629,8 +2629,8 @@ export default class GoTrueClient {
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/lib/rfc4648.ts
hf, line 1: Please use this instead: https://github.com/supabase-community/base64url-js

src/lib/helpers.ts
```diff
@@ -194,23 +168,39 @@ export class Deferred<T = any> {
...
 
   if (parts.length !== 3) {
-    throw new Error('JWT is not valid: not a JWT structure')
+    throw new AuthInvalidJwtError('Invalid JWT structure')
   }
 
-  if (!base64UrlRegex.test(parts[1])) {
-    throw new Error('JWT is not valid: payload is not in base64url format')
+  // Regex checks for base64url format
+  const base64UrlRegex = /^([a-z0-9_-]{4})*($|[a-z0-9_-]{3}=?$|[a-z0-9_-]{2}(==)?$)$/i
```
hf, line 187: There's no padding in Base64 URL:

```suggestion
  const base64UrlRegex = /^([a-z0-9_-]{4})*($|[a-z0-9_-]{3}$|[a-z0-9_-]{2}$)$/i
```

You should move this to be toplevel constant `BASE64ULR_REGEX`, only V8 knows how to make regexes automatically toplevel constants. All other engines will recompile the regex on each call.

src/lib/helpers.ts
```diff
@@ -194,23 +168,39 @@ export class Deferred<T = any> {
...
+    }
   }
-
-  const base64Url = parts[1]
-  return JSON.parse(decodeBase64URL(base64Url))
+  const decoder = new TextDecoder()
+  const data = {
+    header: JSON.parse(decoder.decode(base64url.parse(parts[0], { loose: true }))),
+    payload: JSON.parse(decoder.decode(base64url.parse(parts[1], { loose: true }))),
+    signature: base64url.parse(parts[2], { loose: true }),
```
hf, line 197: Please use our library, it's been tested to support UTF-8 unlike other libraries / atob / btoa.

src/GoTrueClient.ts
```diff
@@ -2599,4 +2594,126 @@ export default class GoTrueClient {
...
+    let jwk = jwks.keys.find((key) => key.kid === kid)
+    if (jwk) {
+      return jwk
+    }
+
+    // try fetching from cache
+    jwk = this.jwks.keys.find((key) => key.kid === kid)
+    if (jwk) {
+      return jwk
+    }
```
hf, line 2609: It's important to have a process to clear-out entries from the cache according to the cache headers sent in the response. If this is used in a long-running Node server, and the JWT keys get rotated, the new keys will never be fetched.

src/GoTrueClient.ts
```diff
@@ -2599,4 +2594,126 @@ export default class GoTrueClient {
...
+      // Convert JWK to CryptoKey
+      const publicKey = await crypto.subtle.importKey('jwk', signingKey, algorithm, true, [
+        'verify',
+      ])
+      // Verify the signature
+      const isValid = await crypto.subtle.verify(
+        algorithm,
+        publicKey,
+        signature,
+        new TextEncoder().encode(`${rawHeader}.${rawPayload}`)
```
hf, line 2695: Don't use TextEncoder, it's not available in many runtimes. https://github.com/hapijs/joi/issues/2141

Just use our Base64 URL library

src/GoTrueClient.ts
```diff
@@ -2599,4 +2594,126 @@ export default class GoTrueClient {
...
+    jwt?: string,
+    jwks: { keys: JWK[] } = { keys: [] }
+  ): Promise<
+    | {
+        data: { claims: { [key: string]: any } }
+        error: null
+      }
+    | { data: null; error: AuthError }
+    | { data: null; error: null }
+  > {
```
hf, line 2643: Do a basic WebCrypto check here:
```suggestion
  > {
  
  if (!('crypto' in globalThis && 'subtle' in globalThis.crypto)) {
    // fall back to `GET /user`
  } else {
    // ...
```

src/GoTrueClient.ts
```diff
@@ -2599,4 +2594,126 @@ export default class GoTrueClient {
...
+        error: null,
+      }
+    } catch (error) {
+      if (isAuthError(error)) {
+        return { data: null, error }
+      }
+      return {
+        data: null,
+        error: new AuthUnknownError('Unknown error occurred while getting claims', error),
+      }
```
hf, line 2716: Why not just return the `error` instead of wrapping?

src/GoTrueClient.ts
```diff
@@ -2599,4 +2594,126 @@ export default class GoTrueClient {
...
+
+      if (!isValid) {
+        throw new AuthInvalidJwtError('Invalid JWT signature')
+      }
+
+      // If verification succeeds, decode and return claims
+      return {
+        data: {
+          claims: payload,
+        },
```
hf, line 2706: Either return `payload` as `data` or include more information, such as:

```suggestion
        data: {
          claims: payload,
          header,
          signature,
        },
```

src/GoTrueClient.ts
```diff
@@ -2599,4 +2594,126 @@ export default class GoTrueClient {
...
+  }
+
+  /**
+   * @experimental This method may change in future versions.
+   * @description Gets the claims from a JWT. If the JWT is symmetric JWTs, it will call getUser() to verify against the server. If the JWT is asymmetric, it will be verified against the JWKS using the WebCrypto API.
+   */
+  async getClaims(
+    jwt?: string,
+    jwks: { keys: JWK[] } = { keys: [] }
+  ): Promise<
```
hf, line 2636: Since this is a new API, we can make it aware of new API keys.

Do a quick check if `jwt` starts with `sb_` and output an error that you can't use an API key here.

src/GoTrueClient.ts
```diff
@@ -2599,4 +2594,126 @@ export default class GoTrueClient {
...
+    | {
+        data: { claims: { [key: string]: any } }
```
hf, line 2638: Let's have the type be the [required claims](https://supabase.com/docs/guides/auth/auth-hooks/custom-access-token-hook) + this.

Something like:

```suggestion
        data: { claims: { RequiredJWTClaims & { [key: string]: any } }
```

src/GoTrueClient.ts
```diff
@@ -2599,4 +2594,126 @@ export default class GoTrueClient {
...
+          error: null,
+        }
+      }
+
+      if (!header.kid) {
+        throw new AuthInvalidJwtError('Missing kid claim')
+      }
+
+      const algorithm = getAlgorithm(header.alg)
+      const signingKey = await this.fetchJwk(header.kid, jwks)
```
hf, line 2684: Ideally the fetching is done based on the `iss` claim in the JWT, not the setup of the client.

hf: It should however throw an error if they don't match if you're calling `getClaims()` but not throw an error if you're using `getClaims(<my custom jwt>)`

hf: How this works with custom domains we need to explore further...

hf: Maybe have the custom domain `/auth/v1/.well-known/jwks.json` redirect to the project ref.

hf: And/or the `getClaims()` can have a third argument for options which can be used to pin the issuer

src/GoTrueClient.ts
kangmingtay: it will because when a new JWT needs to be validated, there will be a cache miss, which would trigger `getClaims` to fetch the latest JWKS

src/GoTrueClient.ts
kangmingtay: there's already a check when in the `decodeJWT` function that will return an error if it's an invalid JWT structure

src/GoTrueClient.ts
kangmingtay: hmm why? the intention behind nesting it in `claims` is to be consistent with the response structure for `getUser`  and `getSession`

src/GoTrueClient.ts
kangmingtay: discussed internally not to fetch the JWKS from the issuer and use the url that the client is instantiated with

src/GoTrueClient.ts
hf: Oh good point!

src/GoTrueClient.ts
hf: Just to have access to all parts of the JWT, otherwise developers will need to do their own parsing to maybe access the header (to get the `kid` for example) or signature.

src/GoTrueClient.ts
```diff
@@ -2599,4 +2596,124 @@ export default class GoTrueClient {
...
+          data: {
+            claims: payload,
+          },
+          error: null,
+        }
+      }
+
+      if (!header.kid) {
+        throw new AuthInvalidJwtError('Missing kid claim')
+      }
```
hf, line 2683: Actually if there's no `kid` I think it should call `getUser()`. Self-signed JWTs might not be having `kid` right now but they might be "compatible" with Supabase Auth.

src/lib/helpers.ts
```diff
@@ -344,3 +329,31 @@ export function parseResponseAPIVersion(response: Response) {
     return null
   }
 }
+
+export function validateExp(exp: number) {
+  if (!exp) {
+    throw new Error('Missing exp claim')
+  }
+  const timeNow = Date.now() / 1000
```
hf, line 337:
```suggestion
  const timeNow = Math.floor(Date.now() / 1000)
```

src/lib/types.ts
```diff
@@ -1199,3 +1199,32 @@ export type AuthMFAEnrollPhoneResponse =
...
+  aal: AuthenticatorAssuranceLevels
+  session_id: string
+}
+
+export type JwtPayload = RequiredClaims & {
+  [key: string]: any
+}
+
+export interface JWK {
+  kty: 'RSA' | 'EC' | 'OCT'
```
hf, line 1225:
```suggestion
  kty: 'RSA' | 'EC' | 'oct'
```

hf (approved):
With the suggestions above!

hf (approved):

PR #1037, opened 2025-03-17
feat: add `signInWithWeb3` with solana

hf:
Adds `signInWithWeb3` which allows [sign in with Solana](https://github.com/supabase/auth/pull/1918).

**Example usage**

First [connect to the user's wallet](https://solana.com/developers/courses/intro-to-solana/interact-with-wallets) then:

```tsx
const wallet = useWallet()

<Button onClick={() => {
  await supabase.auth.signInWithWeb3({ chain: 'solana', wallet, statement: "Custom message" })
}}>
  Sign in with Solana
</Button>
```

It also works without Wallet Standard, at least with Phantom, Solflare and Brave by reading the `window.solana` object or you can custom send a wallet object that satisfies the `signIn()` or `signMessage()` APIs.

Additionally you can sign the message yourself and send it via the raw API:

```typescript
await supabase.auth.signInWithWeb3({
  chain: 'solana',
  message: 'supabase.com wants you to sign in with your Solana account:\n...',
  signature: signatureUint8Array,
})
```


src/lib/version.ts
hf, line 1: Remove

src/GoTrueClient.ts
```diff
@@ -598,6 +602,126 @@ export default class GoTrueClient {
...
+    } else {
+      throw new Error(
+        '@supabase/auth-js: Both wallet as an object and url must be specified in non-browser environments.'
+      )
+    }
+
+    const url = new URL(options?.url ?? window.location.href)
+
+    const [output] = await resolvedWallet.signIn({
+      nonce: uuid(),
```
hf, line 669: This should be a number not a UUID.

src/lib/types.ts
```diff
@@ -611,6 +612,88 @@ export type SignInWithIdTokenCredentials = {
   }
 }
 
+export interface SolanaSignInInput {
```
Bewinxed, line 615: I recommend you import the types from this package [@solana/wallet-standard-features](https://www.npmjs.com/package/@solana/wallet-standard-features)

import type { SolanaSignInInput, SolanaSignInOutput } from '@solana/wallet-standard-features';

as it could update in the future, you only need the types so you can add it as a DevDependency.

src/GoTrueClient.ts
```diff
@@ -598,6 +602,126 @@ export default class GoTrueClient {
...
+    const { chain, wallet, statement, options } = credentials
+
+    let resolvedWallet: SolanaWallet
+
+    if (typeof wallet === 'object' && wallet?.isSolana === true && (options?.url || isBrowser())) {
+      resolvedWallet = wallet
+    } else if (isBrowser()) {
+      if (!wallet || wallet === 'phantom') {
+        if (
+          'phantom' in window &&
```
Bewinxed, line 634: We can probably use a map to store some commonly used wallets:

Brave Browser uses `window.solana` (alias for `window.braveSolana`)
Solflare => `window.solflare`
BackPack => `window.backpack`

They all have aliases at window.solana, they claim it's discouraged to use the alias but I don't see that changing for a while.

Those are the most commonly used wallets.

src/GoTrueClient.ts
```diff
@@ -598,6 +602,126 @@ export default class GoTrueClient {
...
+          window.phantom.isPhantom === true &&
+          'solana' in window.phantom &&
+          window.phantom.solana &&
+          typeof window.phantom.solana === 'object' &&
+          'isSolana' in window.phantom.solana &&
+          window.phantom.solana.isSolana === true &&
+          'signIn' in window.phantom.solana &&
+          typeof window.phantom.solana.signIn === 'function'
+        ) {
+          resolvedWallet = window.phantom.solana as SolanaWallet
```
Bewinxed, line 647: Do you think we can extract this to a "getUsedWallet" (or similar) function? to tidy things up?

Can probably then cast the `window` object as 
`typeof window & {
    solana?: SolanaWallet
    } to eliminate the type errors.

hf: This is a simple `declare` I'll fix

src/lib/types.ts
hf: Seems to be OK.

src/GoTrueClient.ts
hf: Let's get started with Phantom & Brave first, then I'll figure out how to make this abstract. In any case devs can just do `signInWithWeb3({ wallet: window.backpack })`

src/GoTrueClient.ts
```diff
@@ -601,6 +604,214 @@ export default class GoTrueClient {
...
+          version: '1',
+          domain: url.host,
+          uri: url.href,
+
+          ...(statement ? { statement } : null),
+        })
+
+        let outputToProcess: any
+
+        if (Array.isArray(output) && output[0] && typeof output[0] === 'object') {
```
cstockton, line 685: Missing bounds check on output len, may be worth fixing to have a better error when output is `[]`

hf: Ah that's actually allowed in JavaScript the bad parts 🤣

PR #1039, opened 2025-03-21
fix: set jwks_cached_at

kangmingtay:
* Update `fetchJwk` to refetch the JWKs if the cache is stale
* Add tests for various scenarios

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -6,6 +6,7 @@ import {
   AUTO_REFRESH_TICK_THRESHOLD,
   GOTRUE_URL,
   STORAGE_KEY,
+  JWKS_TTL,
 } from './lib/constants'
 import {
   AuthError,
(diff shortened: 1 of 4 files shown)


src/GoTrueClient.ts
```diff
@@ -2607,7 +2610,9 @@ export default class GoTrueClient {
 
     // try fetching from cache
     jwk = this.jwks.keys.find((key) => key.kid === kid)
-    if (jwk) {
+
+    // jwk exists and jwks isn't stale
+    if (jwk && Date.now() - JWKS_TTL < this.jwks_cached_at) {
```
hf, line 2615: Any chance we can use better date logic here? `Date` objects are comparable also.

hf (approved):

src/GoTrueClient.ts
kangmingtay: hmm what did you have in mind? we pretty much just do an inequality check in most places - i've rearranged the vars to make it abit clearer

kangmingtay (commented):
rerun dogfooding ci

PR #1062, opened 2025-05-09
feat: warn on async subscriber to onAuthStateChange

AJamesPhillips:

Feature to warn developers that subscribing to onAuthStateChange with an async function might cause problems.

Currently no warning is issued either when subscribing, or when calling other supabase functions from inside the onAuthStateChange subscriber leading to the supabase-js client silently hanging.  Related issue: https://github.com/supabase/supabase-js/issues/1401

## What is the new behavior?

The new behaviour results in the developer being warned that they have subscribed with an async function.

## Additional context

I couldn't get the tests to run locally so this PR may not be functioning correctly.

### DIFF
--- a/src/GoTrueClient.ts
+++ b/src/GoTrueClient.ts
@@ -1847,6 +1847,14 @@ export default class GoTrueClient {
   ): {
     data: { subscription: Subscription }
   } {
+    if (callback.constructor.name === 'AsyncFunction') {
+      console.warn(
+        'Calling onAuthStateChange with an async function may cause a deadlock.'
+        + '  For a solution please see the docs: ' +
+        'https://supabase.com/docs/reference/javascript/auth-onauthstatechange'
+      )
+    }
+
     const id: string = uuid()
     const subscription: Subscription = {
       id,
--- a/test/GoTrueClient.test.ts
+++ b/test/GoTrueClient.test.ts
@@ -814,6 +814,34 @@ describe('The auth client can signin with third-party oAuth providers', () => {
     })
   })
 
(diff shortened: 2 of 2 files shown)


src/GoTrueClient.ts
```diff
@@ -1847,6 +1847,14 @@ export default class GoTrueClient {
   ): {
     data: { subscription: Subscription }
   } {
+    if (callback.constructor.name === 'AsyncFunction') {
+      console.warn(
+        'Calling onAuthStateChange with an async function may cause a deadlock.'
+        + '  For a solution please see the docs: ' +
+        'https://supabase.com/docs/reference/javascript/auth-onauthstatechange'
+      )
+    }
```
hf, line 1856: In and of itself there is no problem subscribing an async function, but the problem arises if you use another Supabase Client Library within, and that is almost impossible to detect reliably :(

If you want, you can try to find a way -- maybe setting a boolean on the object and checking it in any other auth-js method?

AJamesPhillips: Yep, I totally agree.  I couldn't think of any elegant way of dealing with this but wanted to help prevent someone else spending half a day getting confused by a race condition.  Perhaps this warning could only be displayed if the debug flag was set to true?

mandarini:
Hi @AJamesPhillips ! Thanks for the contribution and your patience.

This repository is deprecated and has moved to the new [Supabase JS monorepo](ttps://github.com/supabase/supabase-js). I’m going to close it to keep the old repo tidy, before archiving.

Note: This repository is now archived, but you can still see your work, and if needed, copy it over to the new repo. No work is lost!

PR #1082, opened 2025-07-13
feat: add sign in with ethereum to `signInWithWeb3`

Bewinxed:

This PR adds SIWE (Sign-In-With-Ethereum) support to auth-js, related to [this PR on /auth](https://github.com/supabase/auth/pull/2069)

Multiple providers & SIWS (Solana) supported

## What is the new behavior?

Add SIWE (Ethereum) support.

## Additional context

The types are inspired by the the [viem](https://viem.sh/) library, which has minimal & modern types, but they were simplified and copied over to the local code to remove dependency on it.


src/lib/web3/ethereum.ts
```diff
@@ -0,0 +1,192 @@
...
+    requestId,
+    resources,
+    scheme,
+    uri,
+    version,
+  } = parameters
+
+  // Validate fields
+  {
+    if (chainId !== Math.floor(chainId))
```
hf, line 134: That's an interesting equality check I've not seen in a long time. Any specific reason for it, just curious?

src/lib/web3/ethereum.ts
```diff
@@ -0,0 +1,192 @@
...
+export function toHex(value: string): Hex {
+  const bytes = new TextEncoder().encode(value)
+  const hex = Array.from(bytes, (byte) => byte.toString(16).padStart(2, '0')).join('')
+  return ('0x' + hex) as Hex
+}
+
+/**
+ * Creates EIP-4361 formatted message.
+ */
+export function createSiweMessage(parameters: SiweMessage): string {
```
hf, line 117: I see there's some runtime type checks here `typeof ... === 'string'` here. We generally avoid doing this unless absolutely necessary in the client library as we expect JavaScript developers to be fully aware of the types and not abuse them.

To simplify the code, consider removing them?

hf: Also maybe rename the function to indicate this is fallback, only if the wallet doesn't support creating the statement itself?

hf (approved):
Beautiful!

src/lib/web3/ethereum.ts
Bewinxed: Unlike SIWS, in SIWE, the dApp is supposed to create the message, not the wallet.

I can remove the runtime checks and keep basic validation, or we can remove it and leave it to the user to use their own logic/libraries (like `viem`, `ethers`) to create the messages?

PR #1084, opened 2025-07-17
fix: use JSON-based deep clone instead of structuredClone

cemalkilic:
Not all JS runtimes support `structuredClone` (introduced in https://github.com/supabase/auth-js/pull/1023). (Example: https://github.com/supabase/supabase-js/issues/1504) 

- As the session data is safe to JSON serialize, replace `structuredClone` use with JSON de/serialize.
- Added the helper function to increase the readability (making the intent clear).


hf (approved):

hf:
Fixes: https://github.com/supabase/supabase-js/issues/1504

PR #1100, opened 2025-08-13
Only warn if multiple clients share the same storage-key

sh41:

Bug fix  

A warning: `Multiple GoTrueClient instances detected in the same browser context. It is not an error, but this should be avoided as it may produce undefined behavior when used concurrently under the same storage key.`  appears when using multiple clients, even if those clients do not share a storage-key. 

## What is the new behavior?

The warning will only appear if there are multiple clients with the same storage key. 

## Additional context

My specific use case for this is in tests where I want to have multiple clients running in parallel. One of them sets up fixtures and uses the service role, others represent different users. The nature of the tests is that they all run in the same instance, so the warning was appearing. Now I can instantiate the clients with unique storage keys and the warning will only appear if I actually have multiple clients that share those storage keys. 

```typescript
// example with supabase-js / ssr
import { createClient  } from "@supabase/supabase-js";
import { createBrowserClient  } from "@supabase/supabase-ssr";
const fixturesClient = createClient(supabaseUrl, serviceRoleKey, { auth: { storageKey: "FIXTURES_CLIENT"}});
// create fixtures
const userClient = createBrowserClient(supabaseUrl, anonKey, { auth: { storageKey: "USER"}});
// login user 
await userClient.auth.signInWithPassword({ email, password})
```


mandarini:
Hi @sh41 ! Thanks for the contribution and your patience.

This repository is deprecated and has moved to the new [Supabase JS monorepo](ttps://github.com/supabase/supabase-js). I’m going to close it to keep the old repo tidy, before archiving.

Note: This repository is now archived, but you can still see your work, and if needed, copy it over to the new repo. No work is lost!

PR #1125, opened 2025-09-25
fix: make webauthn param optional in .authenticate() and .register() and move register params to webauthn: object for consistency

Bewinxed:

Bug fix / API improvement

The `webauthn` parameter in the `_authenticate()` method is required, and the `_register()` method has WebAuthn-related parameters (`rpId`, `rpOrigins`, `signal`) at the top level instead of being grouped under a `webauthn` object.

## What is the new behavior?

- Made the `webauthn` parameter optional in `_authenticate()` method with default empty object fallback
- Moved WebAuthn-related parameters in `_register()` method into a `webauthn` object for consistency with `_authenticate()` 
- Both methods now have consistent API signatures with optional `webauthn` parameter

## Additional context

None.


mandarini:
Moved to: https://github.com/supabase/supabase-js/pull/1765

PR #1127, opened 2025-09-25
chore: add webauthn tests

Bewinxed:

Enhancement + Small bug fix
Add tests for the `webauthn` factorType and its (de)serialization/merging helpers.
Fix fallback naming for the credential when no friendlyname is provided and the username is null

The `factorId` parameter was being sent in both the URL path and the request body when calling the MFA challenge endpoint, which could cause issues with the API.

No tests for factorType `webauthn`

## What is the new behavior?

- Destructured `factorId` from params in the `_challenge` method to only include it in the URL path, not the request body
- Added tests for serialization/deserialization.
- Added WebAuthn helper tests to validate merge operations
- Configured docker-compose.yml with WebAuthn environment variables for local testing

## Additional context

We still need to bump the /auth version in the `dockor-compose.yml` once the build that supports webauthn is out.


mandarini:
moved here: https://github.com/supabase/supabase-js/pull/1763

